PR #6, opened 2022-11-23
feat: add reactwind showcase

nmerget:


packages/components/src/components/card/card.scss
```diff
@@ -21,8 +21,7 @@
 		text-decoration: none;
 	}
 
-	&[data-variant="w-full"],
-	&[data-variant="width-full"] {
+	&[data-variant="w-full"] {
```
mfranzke, line 24: I think we should prefer a clear naming instead of the abbreviation mainly for clarity.

nmerget: Will be handled here https://github.com/db-ui/mono/pull/85

packages/components/src/components/card/card.scss
```diff
@@ -15,14 +15,34 @@
...
+
+			& > .db-card-image {
+				height: 100%;
+			}
+		}
+	}
+
+	&[data-padding="default"] {
+		padding: $db-spacing-fixed-md;
+	}
```
mfranzke, line 39: why would somebody need to define something that's even called "default" itself ? This doesn't sound clear to me.

nmerget: You are right. I removed it for now.

We should discuss this with UX next year. I guess for something like the Card we should define a default. 
UX said we should use an extension, but it doesn't make sense for development.

showcases/reactwind-showcase/index.html
```diff
@@ -0,0 +1,20 @@
...
+		<link rel="icon" type="image/svg+xml" href="/favicon.ico" />
+		<meta name="viewport" content="width=device-width, initial-scale=1.0" />
+		<title>Vite + React + TS + Tailwind</title>
+	</head>
+	<body>
+		<div
+			id="root"
+			class="db-bg-neutral-1 bg-no-repeat bg-scroll"
+			style="
+				background-image: url('images/Bahnstrecke_Vendo_GettyImages-1126385830.jpg');
```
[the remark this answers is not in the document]
nmerget: fixed

packages/components/src/components/card/card.lite.tsx
```diff
@@ -34,12 +38,20 @@ export default function DBCard(props: DBCardProps) {
...
+			data-color-variant={props.colorVariant}
+			data-direction={props.direction}>
 			<Show when={state.stylePath}>
 				<link rel="stylesheet" href={state.stylePath} />
 			</Show>
+			<Show when={props.imgSrc}>
+				<img
+					class="db-card-image"
+					src={props.imgSrc}
+					alt={props.imgAlt}
```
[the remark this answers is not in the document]
nmerget: fixed

mfranzke (approved):

PR #9, opened 2022-11-24
refactor: renamed base to foundations and core to components

nmerget:


package.json
```diff
@@ -11,16 +11,17 @@
...
+    "update:base:scripts": "cpr ../base/scripts packages/foundations/scripts -o",
+    "update:base:scss": "cpr ../base/scss packages/foundations/scss -o",
+    "update:base:tokens": "cpr ../base/tokens packages/foundations/tokens -o",
+    "update:base:sd-config": "cpr ../base/style-dictionary.config.json packages/foundations/style-dictionary.config.json -o",
+    "update:base:sd": "cpr ../base/style-dictionary.js packages/foundations/style-dictionary.js -o",
     "update:base": "npm-run-all -p update:base:*",
+    "postupdate:base": "replace-in-file packages/foundations/**  --configFile=scripts/base-foundation.js && node scripts/rename-db-ui-base.js",
+    "update:core:button": "cpr ../core/source/_patterns/01-elements/buttons/button.scss packages/components/src/components/button/button.scss -o",
+    "update:core:card": "cpr ../core/source/_patterns/02-components/cards/card.scss packages/components/src/components/card/card.scss -o",
+    "update:core:cards": "cpr ../core/source/_patterns/02-components/cards/cards.scss packages/components/src/components/cards/cards.scss -o",
```
mfranzke, line 24: why not update the script names themselves ?

nmerget: fixed

packages/components/mitosis.config.js
```diff
@@ -1,5 +1,5 @@
 const PowerAppsPlugin = require('./plugins/power-apps');
-const Debug = require('./plugins/debug');
+/* Const Debug = require('./plugins/debug'); */
```
annsch, line 2: shouldn't we delete this instead of commenting out?

nmerget: fixed

packages/components/mitosis.config.js
```diff
@@ -12,12 +11,9 @@ module.exports = {
 		angular: {
 			typescript: true
 		},
-		vue3: {
```
mfranzke, line 15: what's the reason to remove vue3 ?

nmerget: we use the default settings at the moment, and don't need to overwrite it

scripts/base-foundation.js
```diff
@@ -0,0 +1,4 @@
+module.exports = {
+	from: /@db-ui\/base/g,
+	to: '@db-ui/foundations'
+};
```
mfranzke, line 4: what about a comment in this file that describes what is for or what is doing (most likely some "parameters" for a replacement function).

mfranzke (approved):

PR #20, opened 2022-12-06
feat: cicd

nmerget:


.github/workflows/01-build-packages.yml
```diff
@@ -0,0 +1,35 @@
...
+      - name: ⬇ Checkout repo
+        uses: actions/checkout@v3
+
+      - name: 🔄 Init Cache
+        uses: ./.github/actions/npm-cache
+
+      - name: 🔨 Build Packages
+        run: npm run build
+
+      - name: 🧹 Clean node_modules and assets
```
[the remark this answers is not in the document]
nmerget: fixed

package.json
```diff
@@ -1,16 +1,19 @@
...
+    "showcases/*"
   ],
+  "type": "module",
   "scripts": {
     "prepare": "node -e \"try { require('husky').install() } catch (e) {}\"",
     "lint": "npm-run-all -p lint:*",
     "lint:markdownlint": "markdownlint -c .markdown-lint.yml **/*.md",
     "lint:stylelint": "stylelint **/*.scss",
-    "lint:xo": "xo",
+    "lint:xo": "xo --fix",
```
mfranzke, line 16: this isn't relevant for the pipeline, but would lead to a changed file during commit process on localhost – is this what you're targeting at ?

.github/workflows/release.yml
```diff
@@ -0,0 +1,57 @@
...
+  #    needs: [cypress]
+  #  deploy:
+  #    uses: ./.github/workflows/03-deploy-gh-pages.yml
+  #    needs: [checks-done, get-publish-version]
+  #    with:
+  #      release: ${{ needs.get-publish-version.outputs.release }}
+  #      preRelease: ${{ needs.get-publish-version.outputs.preRelease }}
+
+  #    needs: [checks-done, get-publish-version, build-outputs]
+  publishpackages:
```
mfranzke, line 50: naming: you've used a dash for word separation in the names before

mfranzke (changes requested):

package.json
```diff
@@ -1,16 +1,19 @@
 {
-  "name": "db-one",
-  "version": "1.0.0",
+  "name": "@db-ui/mono",
```
mfranzke, line 2: what does the root package contain that we would actually be publishing (in combination with line 85: `"access": "public"`)

.github/workflows/default.yml
```diff
@@ -0,0 +1,48 @@
...
+#  cypress:
+#    uses: ./.github/workflows/02-cypress.yml
+#    needs: [install-cypress]
+
+#  checks-done:
+#    runs-on: ubuntu-latest
+#    steps:
+#      - name: 🎉 Checks done
+#        run: echo "🎉 All Cypress tests were successful."
+#    needs: [cypress]
```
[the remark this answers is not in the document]
nmerget: Maybe you should close your tabs sometimes or reload the page while doing a review haha, I did it already ;)

package.json
nmerget: I had a lot of issues with the line endings CSLF and LF and xo rejected the commit all the time. I think we can use --fix here if we trust xo

package.json
nmerget: We won't publish from the root monorepo, the reason is that we have the workspaces for all output targets inside the packages/components. For deployment, I created another directory build-outputs to separate the publishing process.

packages/foundations/tokens/screens.json
```diff
@@ -1,25 +1,25 @@
 {
 	"screens": {
 		"sm": {
-			"value": "640",
+			"value": "40",
```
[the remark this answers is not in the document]
nmerget: I had to change this to rem values because for our new foundation targets (compose and ios-swift) we need to use rem to calculate the correct size.

Is it important to use those in web with px values?

showcases/angular-showcase/angular.json
```diff
@@ -36,8 +37,8 @@
 								},
 								{
 									"type": "anyComponentStyle",
-									"maximumWarning": "2kb",
-									"maximumError": "4kb"
+									"maximumWarning": "10kb",
+									"maximumError": "20kb"
```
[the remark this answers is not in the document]
nmerget: Those are the assets like .css which are causing problems

.github/workflows/03-deploy-gh-pages.yml
```diff
@@ -0,0 +1,107 @@
...
+        env:
+          RELEASE: ${{ inputs.release }}
+          PRE_RELEASE: ${{ inputs.preRelease }}
+          BRANCH_NAME: ${{ steps.extract_branch.outputs.branch-name }}
+        run: |
+          if [[ $RELEASE == "true" || $PRE_RELEASE == "true" ]]
+          then
+            echo "##[set-output name=name;]$(echo ${GITHUB_REF#refs/tags/})"
+          else
+            echo "##[set-output name=name;]$(echo $BRANCH_NAME)"
```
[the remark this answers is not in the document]
nmerget: fixed

.github/workflows/release.yml
nmerget: fixed

packages/foundations/tokens/screens.json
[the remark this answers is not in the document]
mfranzke: This is somehow more of a conceptual question, because the main difference is regarding whether those should change in case that only the text-size gets increased by browser settings vs. the zoom functionality is actually being used (e.g. in Firefox: "Zoom text only").

package.json
mfranzke: I actually **love** XO

mfranzke (approved):

PR #67, opened 2022-12-12
feat: add build for v-components

nmerget:


package.json
```diff
@@ -64,8 +68,10 @@
     "stylelint": "14.14.1",
     "stylelint-config-prettier": "9.0.4",
     "stylelint-config-sass-guidelines": "9.0.1",
+    "tslib": "^2.4.1",
     "typescript": "^4.7.3",
     "validate-branch-name": "^1.3.0",
+    "vue": "^3.2.45",
```
mfranzke, line 74: do we need the frameworks as a dependency in root or doesn't it even already work by having them in the relevant workspace ?

showcases/angular-showcase/package.json
```diff
@@ -37,6 +37,8 @@
     "karma-coverage": "~2.2.0",
     "karma-jasmine": "~5.1.0",
     "karma-jasmine-html-reporter": "~1.7.0",
+    "node-sass": "^7.0.3",
```
mfranzke, line 40: do we really need the deprecated `node-sass` package here? Without further insights, at least I could say that the [`sass-loader`](https://www.npmjs.com/package/sass-loader) mentions to even also support its successor `sass`.

mfranzke (approved):

package.json
nmerget: fixed

showcases/angular-showcase/package.json
nmerget: fixed

annsch (dismissed):

annsch (approved):

PR #109, opened 2022-12-14
Feat examples

nmerget:


examples/react-example/src/index.scss
```diff
@@ -0,0 +1,10 @@
+@use "@db-ui/foundations/build/scss/rollup.assets-paths" as *;
+@use "@db-ui/foundations/build/scss/icon/icons" as *;
+@use "@db-ui/foundations/build/scss/db-ui-foundations" as *;
+@use "@db-ui/foundations/build/scss/variables.global" as *;
+
+@import "@db-ui/foundations/build/scss/color-classes";
```
mfranzke, line 6: Maybe a quick comment for each included file both here and within the README adds some more value for the consuming users on its contents and why those get included.

.xo-config.json
```diff
@@ -14,6 +14,7 @@
 		"n/file-extension-in-import": 0,
 		"@typescript-eslint/no-unsafe-return": 0,
 		"@typescript-eslint/no-unsafe-assignment": 0,
-		"unicorn/prefer-top-level-await": 0
+		"unicorn/prefer-top-level-await": 0,
+		"unicorn/filename-case": 0
```
mfranzke, line 18: why do we need to define this rule (totally deactivating any naming style)?

examples/vue-example/src/main.ts
```diff
@@ -0,0 +1,6 @@
+import { createApp } from 'vue';
+import '@db-ui/v-components/dist/style.css';
```
mfranzke, line 2: I can't find this `import` on the related `examples/react-example/src/main.tsx` file for React – is this done on purpose ?

nmerget: Yes, I think it is some vue-specific thing...

.xo-config.json
nmerget: fixed

examples/react-example/src/index.scss
nmerget: fixed

annsch (approved):

PR #116, opened 2022-12-16
feat: add gh-page to preview showcases

nmerget:


nmerget:
> Please provide a `.license` file to `showcases/next-showcase/public/vercel.svg`

Fixed

showcases/patternhub/styles/globals.scss
```diff
@@ -0,0 +1,23 @@
+@use "@db-ui/foundations/build/scss/webpack.assets-paths" as *;
```
mfranzke, line 1: is patternhub using webpack?

.eslintignore
```diff
@@ -4,3 +4,4 @@
 **/helpers
 /packages/components/**
 !/packages/components/scripts
+showcases/patternhub/next.config.js
```
mfranzke, line 7:
```suggestion

```
I don't understand the reason why in this case? In my opinion we shouldn't ignore our individual code (and configuration) implementations. Plus there's only one remark that `xo` is giving as feedback to that file which should be easily changeable.

mfranzke:
It seems like that a lot of files in this PR are related to code formatting / beautifying. It's frustrating to go through these changes that aren't code related, but only code style. Could you please separate those ones, by doing another commit or PR to main first with the code style changes?

showcases/patternhub/styles/globals.scss
nmerget: It's based on nextjs and they use webpack: https://nextjs.org/docs/api-reference/next.config.js/custom-webpack-config

.eslintignore
nmerget: fixed

nmerget:
> It seems like that ~a lot of~ some files in this PR are related to code formatting / beautifying. It's frustrating to go through these changes that aren't code related, but only code style. Could you please separate those ones, by doing another commit or PR to main first with the code style changes?

Reverted it, let's look at it here: https://github.com/db-ui/mono/issues/244

showcases/patternhub/components/default-page.tsx
```diff
@@ -0,0 +1,68 @@
...
+		.map((route) => getRouteWithBasePath(route));
+};
+
+const DefaultPage = ({ children }: any) => {
+	const router = useRouter();
+	return (
+		<StaticContent>
+			<DbPage>
+				<DbHeader slot="header">
+					<DbBrand src="https://raw.githubusercontent.com/db-ui/mono/main/packages/foundations/assets/images/db_logo.svg">
```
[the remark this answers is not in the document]
nmerget: fixed

showcases/patternhub/components/default-page.tsx
```diff
@@ -0,0 +1,68 @@
...
+					{process.env.NEXT_PUBLIC_GITHUB_VERSION_SWITCHER ===
+						'true' && (
+						<GithubVersionSwitcher
+							owner={process.env.NEXT_PUBLIC_GITHUB_OWNER}
+							repo={process.env.NEXT_PUBLIC_GITHUB_REPO}
+						/>
+					)}
+				</DbHeader>
+				<main>
+					<div>{children}</div>
```
mfranzke, line 59: what for do we need that `<div>` ?

showcases/patternhub/package.json
```diff
@@ -0,0 +1,37 @@
+{
+  "name": "db-patternhub",
+  "version": "0.1.0",
+  "private": true,
+  "scripts": {
+    "copy:components": "cpr ../../output/next/src ./components/src -o",
+    "prebuild": "npm-run-all copy:*",
```
mfranzke, line 7: I like the idea with pre-scripts, but as soon as the amount of scripts increases, this might get confusing. And especially this might be simpler written in the `build` script directly: `npm-run-all copy:* build:*`

mfranzke (approved):

showcases/patternhub/components/default-page.tsx
nmerget: I removed the `<main>` tag. The mdx files wrapping the content around main anyways.
So the `<div>` hat to remain as container item to work with `<db-page>`: https://github.com/db-ui/elements/blob/main/packages/db-ui-elements-stencil/src/components/db-page/db-page.scss#L33

showcases/patternhub/package.json
nmerget: I think it doesn't matter, I changed it to npm-run-all

showcases/patternhub/components/default-page.tsx
mfranzke: what's the purpose of that overflow:hidden https://github.com/db-ui/elements/blob/main/packages/db-ui-elements-stencil/src/components/db-page/db-page.scss#L19? Maybe we could get rid of that container item.

mfranzke (approved):

mfranzke (approved):

mfranzke (approved):

PR #118, opened 2022-12-16
feat: input component

annsch:


packages/components/src/components/input/input.lite.tsx
```diff
@@ -0,0 +1,55 @@
...
+				<DBIcon icon={props.iconBefore} className=" icon-before" />
+			</Show>
+			<input
+				name={props.name}
+				type={props.type || 'text'}
+				placeholder={props.placeholder}
+			/>
+			<label for={props.name} aria-hidden="true" id="id-label">
+				{props.label}
+			</label>
```
mfranzke, line 40: The `input` and `label` elements are missing some mandatory attributes, like:
- unique `id` on `input`, that ensures the connection of the `label` with a `for` attribute
- `aria-labelledby` on `input` that references the unique `id` of the `label`

compare to e.g. https://github.com/db-ui/core/blob/dbux-3/source/_patterns/01-elements/input/input.hbs#L18

packages/components/src/components/input/input.lite.tsx
```diff
@@ -0,0 +1,55 @@
+import { onMount, Show, useMetadata, useStore } from "@builder.io/mitosis";
+import { DBInputState, DBInputProps, iconVariants } from "./model";
+import { DBIcon } from '../icon';
+import "./input.scss";
+
+useMetadata({
+  isAttachedToShadowDom: true,
```
mfranzke, line 7: ShadowDOM would most likely lead to problems and wouldn't be necessary most likely, compare to our previous evaluation: https://github.com/db-ui/elements/blob/main/doc/adr/adr-03-viewEncapsulation.md#entscheidung-und-begr%C3%BCndung

packages/foundations/scss/_init.scss
```diff
@@ -23,6 +23,27 @@
 	}
 }
 
+body,
+button,
+input,
+optgroup,
+select,
+textarea {
+	font-family: $db-font-family-sans;
```
mfranzke, line 32: Doesn't this declaration duplicate https://github.com/db-ui/mono/blob/main/packages/foundations/scss/_tonality.scss#L7 ?

annsch: @mfranzke you're right, but we have to overwrite user agent stylesheet, so I changed that to _inherit_. We could discuss where to put those overwrites: should it done in every component or overall in the init.scss. What do you mean?

packages/components/src/components/input/input.lite.tsx
annsch: thank you for this hint, I changed that in the current version

packages/components/src/components/input/input.lite.tsx
annsch: That happens on draft: PRs ;) Updated those missing attributes (disabled and required as well ...)

packages/components/src/components/input/input.lite.tsx
```diff
@@ -0,0 +1,67 @@
...
+  onMount(() => {
+    if (props.stylePath) {
+      state.stylePath = props.stylePath;
+    }
+  });
+
+  return (
+		<div
+			class={
+				'db-input' +
```
mfranzke, line 26:
```suggestion
				'db-input ' +
```
My understanding of this would be that there wouldn't be a space in between `db-input` and another class added by className if we don't add one here at the end (or a more complicated solution within the next line).

packages/components/src/components/input/input.lite.tsx
```diff
@@ -0,0 +1,67 @@
...
+				(props.className || '') +
+				(props.iconBefore ? ' icon-before' : '') +
+				(props.iconAfter ? ' icon-after' : '')
```
mfranzke, line 29: we could use the `:has` selector within CSS to achieve a different styling instead.

packages/components/src/components/input/input.lite.tsx
```diff
@@ -0,0 +1,67 @@
...
+				placeholder={props.placeholder}
+				aria-labelledby={props.id + '-label'}
+				disabled={props.disabled}
+				required={props.required}
+			/>
+			<label for={props.id} aria-hidden="true" id={props.id + '-label'}>
+				<span>{props.label}</span>
+				<Show when={props.required}>
+					<span> *</span>
+				</Show>
```
mfranzke, line 51:
```suggestion

```
why not use CSS `::after { content: " *";` for this?

packages/foundations/scss/_init.scss
mfranzke: With `inherit` and selecting the regular HTML tags, it's fine to keep in here instead of every component, I think.

showcases/angular-current-showcase/src/styles.scss
```diff
@@ -3,3 +3,10 @@
 @use "@db-ui/foundations/build/scss/webpack.assets-paths" as *;
 @use "@db-ui/foundations/build/scss/icon/icons" as *;
 @use "@db-ui/components/src/db-ui-components" as *;
+@use "@db-ui/foundations/build/scss/db-ui-foundations" as *;
+
+@use "@db-ui/foundations/build/scss/tonality" as *;
+@use "@db-ui/foundations/build/scss/variables.global" as *;
+
+@import "@db-ui/foundations/build/scss/color-classes";
+
```
mfranzke, line 12: isn't there a file in the components / foundations part that includes this already that we could `@use` here to prevent redundancy?

annsch: fixed according to react showcase

mfranzke (approved):

PR #211, opened 2023-01-18
feat: add event handler for button and card

nmerget:


packages/components/src/components/card/card.lite.tsx
```diff
@@ -39,9 +48,12 @@ export default function DBCard(props: DBCardProps) {
...
 		<div
 			class={'db-card' + (props.className ? ' ' + props.className : '')}
-			data-variant={props.variant}
+			data-variant={
+				props.variant || ((props.click || props.onClick) && 'ia')
+			}
 			data-color-variant={props.colorVariant}
-			data-direction={props.direction}>
+			data-direction={props.direction}
+			onClick={(event) => state.handleClick(event)}>
```
mfranzke, line 56: We should either use a `button` or `a` HTML tag within the `card` component to enable keyboard and screenreader compatibility.

mfranzke: Opened a follow up ticket: https://github.com/db-ui/mono/issues/212

mfranzke (approved):

PR #226, opened 2023-01-19
feat: add db-brand, db-header, db-page

nmerget:


packages/components/src/shared/model.ts
```diff
@@ -15,6 +15,13 @@ export type GlobalTextProps = {
 	pattern?: string;
 };
 
+export type ImageProps = {
+	imgAlt?: string;
+	imgSrc?: string;
+	imgHeight?: number | string;
+	imgWidth?: number | string;
```
mfranzke, line 22:
```suggestion
	imgHeight?: number;
	imgWidth?: number;
```

> The intrinsic width of the image in pixels. Must be an integer without a unit.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-width

mfranzke (dismissed):

packages/components/src/shared/model.ts
mfranzke: how is this resolved ?

mfranzke (approved):

showcases/react-showcase/src/app.tsx
```diff
@@ -12,97 +15,116 @@ const App = () => {
...
-				</section>
+						<DBInput
+							label="Startdatum"
+							placeholder="irgendein Text"
+							type="datetime-local"
+							id="input-func-date"
+						/>
+					</section>
+				</div>
 			</div>
```
mfranzke, line 126:
```suggestion
			</main>
```

every page should contain a `<main>` element, mainly for a11y purpose to act as a landmark.

mfranzke (changes requested):

showcases/react-showcase/src/app.tsx
nmerget: Had to revert it...
`<DBPage>` has a `<main>` tag already: https://github.com/db-ui/mono/blob/feat-db-brand/packages/components/src/components/page/page.lite.tsx#L40

We gonna refactor the showcases in the future. At the part you mentioned would be a native router outlet (inside the main tag) that loads another component (content-site) based on the route. So I think everything is fine here

mfranzke (approved):

PR #227, opened 2023-01-19
refactor: removed styles from components to reduce dependencies

nmerget:


mfranzke:
@nmerget could you please add a little more information within the description, as this is not self-explaining to me.

mfranzke (approved):

PR #228, opened 2023-01-19
feat: add db-ui-42.scss for one file integrations

nmerget:


packages/components/src/styles/db-ui-42-webpack.scss
```diff
@@ -0,0 +1,8 @@
+/* This is a predefined beginner friendly bundle for webpack apps */
+
+@use "@db-ui/foundations/build/scss/webpack.assets-paths" as *;
+@use "@db-ui/foundations/build/scss/icon/icons" as *;
+@use "@db-ui/components/src/styles/db-ui-components" as *;
+
+@import "@db-ui/foundations/build/scss/variables.global";
+@import "@db-ui/foundations/build/scss/color-classes";
```
mfranzke, line 8: It doesn't work anymore to `@use` (previously we used `@import`) the file `packages/components/src/styles/db-ui-42.scss` in here?

nmerget: fixed

showcases/angular-current-showcase/angular.json
```diff
@@ -33,7 +33,7 @@
 								{
 									"type": "initial",
 									"maximumWarning": "500kb",
-									"maximumError": "1mb"
+									"maximumError": "2mb"
```
mfranzke, line 36: do we really need more than one MB for our assets at the moment? Is there probably something wrong (we most likely still have the duplicated CSS issue).

nmerget: I thought that as well, and we could reduce the size with something like postcss? 
Let's add a new task for this. At the moment we don't have a big issue, because the shipped file size is around 40kb, only the full file has 1.5mb.

mfranzke: We should definitely integrate something like that ([CSSNANO](https://cssnano.co/) looks promising, I've tested it the other day, it could even merge some declarations that we're currently spreading across some files for better separation), but this should most likely only squeeze out the last bytes – if we do have a general problem within our imports of our modular SCSS files, our users who might do an individual SCSS consumption and build would either still have this problem or additionally even also setup postcss on their end which might be acceptable, but still strange if we could instead analyze and fix our file imports instead.

nmerget: image

It is a angular thing, I don't know if we really have a problem with this

nmerget: Guess I found the issue:

@import "@db-ui/foundations/build/scss/color-classes";

This file is large, maybe we can lazyload it?

mfranzke: I've opened another task for having a look at the CSS contents as well.

mfranzke: is `@db-ui/foundations/build/scss/color-classes` included within that `styles.css` file as well? Outputting it separatly would probably both solve this problem as well as give our users the possibility to choose whether they want to use the classes or not by referencing that file if needed.

mfranzke (approved):

showcases/angular-current-showcase/angular.json
nmerget: fixed

PR #230, opened 2023-01-20
feat: add eslint plugin to inform user that the style is v2

nmerget:


packages/eslint/index.js
```diff
@@ -0,0 +1,9 @@
+const { onOldComponentUsed } = require('./old-components');
+
+module.exports = {
+	rules: {
+		'old-component-used': {
```
annsch, line 5: should we rename this to v2? 'old' could be misleading?

annsch: or deprecated?

mfranzke: > or deprecated?

they aren't deprecated (already and especially not in the other package). So I like the `v2`.

mfranzke:
Even though that I like the solution to make it transparent to the consumer of our packages, which components are still only migrated and will get redesigned the sooner the later, the question to me is whether this is the best way to do it via eslint warnings.

Especially in this situation, the consuming developers aren't able to do / change / correct anything, so it would lead to that these are forced to ignore eslint warnings, which might be a bad thing. Even especially as we even also could help out as we're early in the progress of implementation, they would be left on their own.
Instead we could at least mention this within the documentation or README.

packages/eslint/index.js
nmerget: fixed

nmerget:
> Even though that I like the solution to make it transparent to the consumer of our packages, which components are still only migrated and will get redesigned the sooner the later, the question to me is whether this is the best way to do it via eslint warnings.
> 
> Especially in this situation, the consuming developers aren't able to do / change / correct anything, so it would lead to that these are forced to ignore eslint warnings, which might be a bad thing. Even especially as we even also could help out as we're early in the progress of implementation, they would be left on their own. Instead we could at least mention this within the documentation or README.

I updated the README. So I would show an overview of all components and some kind of progress and eslint will be optional. Maybe we can add other eslint rules later on to have something like "We are ready with the component, please update"?
I guess we can start with this and do a migration for the components?

.eslintrc.js
```diff
@@ -1,10 +0,0 @@
-module.exports = {
-	env: {
-		browser: true,
-		es2021: true,
-		node: true
-	},
-	rules: {
-		'no-console': 'error'
```
mfranzke, line 8: why remove this file? Storing those rules in here is more universal, than "just" serving `xo`.

scripts/start.js
```diff
@@ -49,6 +49,6 @@ inquirer
 				startCommand += ` start-showcase:${answer}`;
 			}
 
-		console.log(`Start with command: '${startCommand}'`);
+		console.warn(`Start with command: '${startCommand}'`);
```
mfranzke, line 52: I understand, that https://github.com/db-ui/mono/pull/230/files#diff-94de25957c3047cf6889109c85175c89d62a55646fe7864103ac0fe9265050c6R9 is supposed to be a warning, but this sounds to me like a regular logging instead of a warning.

mfranzke, line 52: I understand, that https://github.com/db-ui/mono/pull/230/files#diff-94de25957c3047cf6889109c85175c89d62a55646fe7864103ac0fe9265050c6R9 is supposed to be a warning, but this sounds to me like a regular logging instead of a warning.

```suggestion
		console.log(`Start with command: '${startCommand}'`);
```

.eslintrc.js
nmerget: At least for me the `eslintrc.js` wasn't used by `xo`. So I thought it is useless because we do not run eslint in this repo

scripts/start.js
nmerget: fixed

mfranzke (approved):

PR #253, opened 2023-01-25
fix: input default value

annsch:
- solved Problem with default value in react
- provide default value vor label (shouldn't be empty according to guidelines 3.0)
- placeholder value mustn't be empty to enable focus transition
- test case for form submit to test input behaviour in different outputs


nmerget:
Did you test it for vue as well?

showcases/react-showcase/src/app.tsx
```diff
@@ -61,7 +67,10 @@ const App = () => {
 						type="datetime-local"
 						id="input-expr-date"
 					/>
-				</section>
+					<DBButton variant="secondary" type="submit">
```
nmerget: `type` does not exist yet :S

annsch: fixed

nmerget (approved):

nmerget (approved):

PR #257, opened 2023-01-25
feat: playwright component tests

dkolba:


.github/workflows/e2e.yml
```diff
@@ -0,0 +1,58 @@
...
+        run: npm run build
+
+      - name: 📲 Install esbuild (binary workaround)
+        working-directory: ./output/react
+        run: npm i -D esbuild-linux-64
+
+      # - name: 🎭 Install Playwright
+      #   working-directory: ./output/react
+      #   run: npx playwright install --with-deps
+
```
mfranzke, line 47: is this a leftover that could get deleted? Or an outstanding rework, that we might need later on? In case of the latter, could you please add a `todo` comment.

output/react/package.json
```diff
@@ -14,13 +14,15 @@
     "mv:dist": "cpr dist ../../build-outputs/react/dist -o",
     "mv:package.json": "cpr package.json ../../build-outputs/react/package.json -o",
     "mv:readme": "cpr README.md ../../build-outputs/react/README.md -o",
-    "postbuild": "npm-run-all -p mv:*"
+    "postbuild": "npm-run-all -p mv:*",
+    "test-ct": "playwright test -c playwright.config.ts"
```
mfranzke, line 18:
```suggestion
    "test:ct": "playwright test -c playwright.config.ts"
```

what does `ct` stand for? Probably it's better to use the long form / name in here instead of the abbreviation for clarity.
Plus it seems to me like we could align the syntax from hyphen to colon.

output/react/tests/button.spec.tsx
```diff
@@ -0,0 +1,58 @@
...
+	test('button should not have any automatically detectable accessibility issues', async ({
+		page,
+		mount
+	}) => {
+		await mount(<DBButton text="Test" />);
+		const accessibilityScanResults = await new AxeBuilder({
+			page
+		})
+			.include('.db-button')
+			.analyze(); // 4
```
mfranzke, line 54: what for are those number comments ? And what happened to 3 ?

.github/workflows/01-build-packages.yml
```diff
@@ -18,12 +18,21 @@ jobs:
       - name: 🔨 Build Packages
         run: npm run build
 
+      - name: list folder
+        run: ls ./packages/components/src/styles
+
```
mfranzke, line 23: Is this a debugging leftover? Or do you actually want to have them listed during build?

output/react/package.json
dkolba: Fixed

output/react/tests/button.spec.tsx
dkolba: Fixed

.github/workflows/01-build-packages.yml
dkolba: Fixed

packages/components/src/components/button/button.lite.tsx
```diff
@@ -63,9 +63,9 @@ export default function DBButton(props: DBButtonProps) {
 
 			<DBIcon icon={props.icon} withText={!props.onlyIcon}>
 				{/* we need spacings around props.text for compilation */}
-				<Show when={props.text}> {props.text} </Show>
 				{props.children}
 			</DBIcon>
+			<Show when={props.text}> {props.text} </Show>
```
nmerget, line 68: We should add `aria-label` to `<button>` and to this `<Show when={props.text && !props.onlyIcon}`

dkolba: Fixed

.github/workflows/default.yml
```diff
@@ -26,21 +26,21 @@ jobs:
...
+    needs: [build-packages]
+
   build-showcases:
     uses: ./.github/workflows/01-build-showcases.yml
     needs: [build-packages]
 
-  #  cypress:
-  #    uses: ./.github/workflows/02-cypress.yml
-  #    needs: [install-cypress]
-
```
mfranzke, line 36: do you want to keep this in here for future reference?

nmerget (dismissed):

.github/workflows/default.yml
dkolba: Deleted all references to Cypress, we will use Playwright in the future for E2E tests

nmerget (dismissed):

mfranzke (approved):

PR #261, opened 2023-01-25
chore: added a comment

mfranzke:


nmerget:
I guess @annsch did this to test if it works with the native input and our input inside the form element

annsch (approved):

PR #263, opened 2023-01-25
fix: pipeline

mfranzke:
reverted some parts out of https://github.com/db-ui/mono/pull/256

### DIFF
--- a/packages/components/src/components/divider/model.ts
+++ b/packages/components/src/components/divider/model.ts
@@ -1,6 +1,6 @@
 import { GlobalProps, GlobalState } from '../../shared/model';
 
-export interface DBDividerDefaultProps {
+export type DBDividerDefaultProps = {
 	margin?: 'none' | '_';
 	variant?: 'horizontal' | 'vertical';
 }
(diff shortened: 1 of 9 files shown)


packages/components/src/components/divider/model.ts
```diff
@@ -1,9 +1,9 @@
-import { type GlobalProps, type GlobalState } from '../../shared/model';
+import { GlobalProps, GlobalState } from '../../shared/model';
 
-export type DBDividerDefaultProps = {
+export interface DBDividerDefaultProps {
```
nmerget, line 3: This could be a type, only the imports are bad

dkolba (commented):
LGTM

nmerget (approved):

mfranzke:
> > LGTM
> 
> @dkolba there's a special button to express your love, that even also makes the world go round (for this PR).

again, @nmerget was faster …

mfranzke:
are you looking at the pipeline with the same amount of excitement as I do ?

PR #264, opened 2023-01-25
feat: add angular example and fix issues for examples

nmerget:


examples/angular-example/.editorconfig
```diff
@@ -0,0 +1,16 @@
...
+indent_size = 2
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.ts]
+quote_type = single
+
+[*.md]
+max_line_length = off
+trim_trailing_whitespace = false
```
mfranzke, line 16:
```suggestion
[*]
indent_style = space
indent_size = 2

[*.md]
max_line_length = off
trim_trailing_whitespace = false
```

examples/angular-example/.vscode/extensions.json
```diff
@@ -0,0 +1,4 @@
+{
+  // For more information, visit: https://go.microsoft.com/fwlink/?linkid=827846
```
mfranzke, line 2: Comments aren't allowed this way in JSON. Is it possible to use either one of these two solutions instead?
- [`JSON5`](https://json5.org/)
- using a hack to define this like this:
```suggestion
  "_for_more_information_visit": "https://go.microsoft.com/fwlink/?linkid=827846"
```

examples/angular-example/.vscode/launch.json
```diff
@@ -0,0 +1,20 @@
+{
+  // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
```
mfranzke, line 2: Comments aren't allowed this way in JSON. Is it possible to use either one of these two solutions instead?
- [`JSON5`](https://json5.org/)
- using a hack to define this like this (does the functionality still work in VS Code?):
```suggestion
  "_for_more_information_visit": "https://go.microsoft.com/fwlink/?linkid=830387"
```

examples/angular-example/.vscode/tasks.json
```diff
@@ -0,0 +1,42 @@
+{
+  // For more information, visit: https://go.microsoft.com/fwlink/?LinkId=733558
```
mfranzke, line 2: Comments aren't allowed this way in JSON. Is it possible to use either one of these two solutions instead?
- [`JSON5`](https://json5.org/)
- using a hack to define this like this (does the functionality still work in VS Code?):
```suggestion
  "_for_more_information_visit": "https://go.microsoft.com/fwlink/?LinkId=733558"
```

examples/angular-example/.editorconfig
mfranzke: As these are self-enclosing example files, there shouldn't be any reliance on files outside of these `examples` path. So this comment is most likely obsolete, right?

examples/angular-example/tsconfig.app.json
```diff
@@ -0,0 +1,10 @@
+/* To learn more about this file see: https://angular.io/config/tsconfig. */
+{
```
mfranzke, line 2: Comments aren't allowed this way in JSON. Is it possible to use either one of these two solutions instead?
- [`JSON5`](https://json5.org/)
- using a hack to define this like this (does the functionality still work in VS Code?):

```suggestion
{
  "_to_learn_more_about_this_file_see": "https://angular.io/config/tsconfig",
```

examples/angular-example/tsconfig.json
```diff
@@ -0,0 +1,29 @@
+/* To learn more about this file see: https://angular.io/config/tsconfig. */
+{
```
mfranzke, line 2: Comments most likely aren't allowed this way in JSON. Is it possible to use either one of these two solutions instead?
- [`JSON5`](https://json5.org/)
- Creating another file at the same spot with the same name, but the fileending `.md`
- using a hack to define this like this (does the functionality still work?):

```suggestion
{
  "_to_learn_more_about_this_file_see": "https://angular.io/config/tsconfig",
```

mfranzke (approved):

examples/angular-example/.editorconfig
nmerget: Yes your are right. Those examples should be used as starting point for a standalone app

mfranzke: Than it wasn't necessary to commit my suggestion. I've reverted that.

mfranzke (approved):

PR #282, opened 2023-01-26
feat: link component

annsch:


packages/foundations/scss/_init.scss
```diff
@@ -42,6 +42,15 @@ a {
 	}
 }
 
+div,
+button,
+input,
+select,
+textarea {
+	font-family: inherit !important;
```
nmerget, line 50: Is there another solution intead of `!important`? :(

packages/components/src/components/link/model.ts
```diff
@@ -0,0 +1,32 @@
...
+	href?: string;
+	anchorLink?: string;
+	title?: string;
+	rel?: string;
+	role?: string;
+	disabled?: boolean;
+	selected?: boolean;
+	label?: string;
+	hreflang?: string;
+	target?: string;
```
nmerget, line 19: Can we add all valid target options here: `_self`, `_blank`, ...

packages/components/src/components/link/model.ts
```diff
@@ -0,0 +1,32 @@
+import {
+	ClickEventProps,
+	ClickEventState,
+	GlobalProps,
+	GlobalState
+} from '../../shared/model';
+
+export interface DBLinkDefaultProps {
```
nmerget, line 8: Do we need `referrerpolicy` -> https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a?retiredLocale=de#attr-referrerpolicy

packages/components/src/components/link/link.lite.tsx
```diff
@@ -0,0 +1,57 @@
+import { onMount, Show, useMetadata, useStore } from '@builder.io/mitosis';
+import { DBIcon } from '../icon';
+import { DBLinkState, DBLinkProps } from './model';
+
+useMetadata({
+	isAttachedToShadowDom: true,
```
mfranzke, line 6: have we already decided on that we always want to use the ShadowDOM? I don't necessarily see any benefit on the elements.

packages/components/src/components/link/link.scss
```diff
@@ -0,0 +1,90 @@
...
+
+	// "disabled" links
+	&[aria-disabled="true"] {
+		opacity: var(--db-disabled-text-opacity, 0.5);
+
+		pointer-events: none;
+
+		&::before {
+			color: initial;
+		}
```
mfranzke, line 34: what for do we need this declaration ?

packages/components/src/components/link/link.scss
```diff
@@ -0,0 +1,90 @@
...
+	// Changing color on hover should be animated by a transition
+	transition: outline 0.06s, color $db-transition-color;
+
+	&:hover {
+		color: inherit;
+	}
+
+	// "disabled" links
+	&[aria-disabled="true"] {
+		opacity: var(--db-disabled-text-opacity, 0.5);
```
mfranzke, line 28:
```suggestion
		opacity: var(--db-disabled-text-opacity, $db-disabled-text-opacity);
```

There should most likely be a SCSS variable `$db-disabled-text-opacity` that we could use instead.

packages/components/src/components/link/link.scss
mfranzke, line 34: what for do we need this declaration ?

```suggestion
```

packages/components/src/components/link/link.scss
annsch: @mfranzke why do we need both? This is not a design token.

packages/foundations/scss/_init.scss
annsch: 💯 just another copy & paste 🐙

packages/components/src/components/link/model.ts
annsch: You're right, I missed that

packages/components/src/components/link/model.ts
annsch: done

packages/components/src/components/link/link.lite.tsx
annsch: changed that, I'm not sure if we could test the vanilla showcase already @nmerget

mfranzke (approved):

PR #294, opened 2023-01-27
fix: icon sizes for tonalities

annsch:


nmerget (approved):

packages/foundations/scss/icon/_icons.helpers.scss
```diff
@@ -90,15 +90,9 @@
...
-	$position: "before",
-	$font-size: $icon-size
-) {
+@mixin icon-meta($size: $icon-size, $style: $icon-style, $position: "before") {
 	&::#{$position} {
-		--icon-font-family: "#{"icons-" + $size + "-" + $style}",
-			"missing-icons";
-		--icon-font-size: #{to-rem($pxValue: $font-size)};
+		--icon-font-family: var(--db-base-icon-font-family);
+		--icon-font-size: var(--db-base-icon-font-size);
```
mfranzke, line 96: I don't get what for we would still need this mixin, as none of the parameters is being retrieved and would set and metadata related to that parameter.

PR #337, opened 2023-02-03
Feat docker compose e2e

nmerget:


output/react/package.json
```diff
@@ -15,7 +15,8 @@
     "mv:package.json": "cpr package.json ../../build-outputs/react/package.json -o",
     "mv:readme": "cpr README.md ../../build-outputs/react/README.md -o",
     "postbuild": "npm-run-all -p mv:*",
-    "test:components": "playwright test -c playwright.config.ts"
+    "test:components": "playwright test -c playwright.config.ts",
+    "regenerate:components": "playwright test -c playwright.config.ts --update-snapshots"
```
dkolba, line 19: "regenerat:components" sounds ambiguous since the goal is to regenerate the screenshots

e2e/docker-compose.regenerate.yml
```diff
@@ -0,0 +1,8 @@
+services:
+  playwright:
+    stdin_open: true
+    tty: true
```
dkolba, line 4: Line 3+4 are also in the docker-compose.yml, do we need them here as well?

packages/components/test/playwright/config.ts
```diff
@@ -19,7 +19,7 @@ const config: PlaywrightTestConfig = {
 	/* Opt out of parallel tests on CI. */
 	workers: process.env.CI ? 1 : undefined,
 	/* Reporter to use. See https://playwright.dev/docs/test-reporters */
-	reporter: process.env.CI ? 'github' : 'html',
```
dkolba, line 22: I guess there is an env var for skipping the automatic launch of the html report.

dkolba (changes requested):

packages/components/test/playwright/config.ts
nmerget: I think the default should be line or list if you running a command on a console.

If someone wants the HTML report, it can be done by adding `--reporters=html` in the cli

output/react/package.json
nmerget: fixed

e2e/docker-compose.regenerate.yml
nmerget: fixed

dkolba (dismissed):

dkolba (approved):

PR #342, opened 2023-02-06
feat: add update pr for push to main

nmerget:


.github/workflows/update-prs.yml
```diff
@@ -0,0 +1,20 @@
...
+    name: Update PRs
+    runs-on: ubuntu-latest
+    steps:
+      - name: ⌚ Update PRs
+        id: cleanup
+        uses: actions/github-script@v6
+        with:
+          script: |
+            const script = require('./.github/scripts/update-prs.js')
+            console.log(await script({github, context}))
```
mfranzke, line 20: @nmerget is this still a leftover, or do you want to log here? In case of the latter, let's add some more information / text at the beginning.

nmerget: fixed

.github/scripts/update-prs.js
```diff
@@ -0,0 +1,32 @@
...
+	const { repo, owner } = context.repo;
+	const pulls = await github.rest.pulls.list({
+		owner,
+		repo,
+		state: 'open',
+		base: 'main',
+		per_page: 100
+	});
+
+	const nonDraftPulls = pulls.filter((pr) => !pr.draft);
```
mfranzke, line 13: I like !

mfranzke (approved):

PR #346, opened 2023-02-06
feat: add new linters

nmerget:


.jscpd.json
```diff
@@ -0,0 +1,27 @@
...
+		"**/node_modules/**",
+		"**/__snapshots__/**",
+		"**/.next/**",
+		"**/.github/**",
+		"**/next-showcase/components/**",
+		"**/patternhub/components/**",
+		"**/index.html",
+		"**/angular.json",
+		"**/tsconfig.json",
+		"**/main.ts"
```
mfranzke: it's clear to me that we want to ignore generated files, but why ignore sources?

nmerget: made it more clear, it was for angular

mfranzke (approved):

PR #363, opened 2023-02-07
refactor: replace placeholder in reactwind with missing components

nmerget:


packages/components/src/components/brand/brand.scss
```diff
@@ -30,3 +29,10 @@
 		}
 	}
 }
+
+.inside-header {
+	height: $db-sizing-sm;
+	@media screen and (min-width: $db-screens-lg) {
+		height: $db-sizing-md;
+	}
+}
```
mfranzke, line 38: why put this into the global "scope"?

packages/components/src/components/brand/model.ts
```diff
@@ -19,6 +19,11 @@ export interface DBBrandDefaultProps {
 	 * The rel attribute can be set to make a relationship to the current URL (https://developer.mozilla.org/en-US/docs/web/api/htmlanchorelement/rel).
 	 */
 	anchorRelation?: string;
+
+	/**
+	 * This is a workaround for shadow dom and header
+	 */
+	insideHeader?: boolean;
```
mfranzke, line 26: This whole adding another property just to enable the cascade thing could probably get easier for the developers, if we would switch to CSS variables for the logos height that could peak the ShadowDOM.

nmerget: fixed

mfranzke (dismissed):

mfranzke (approved):

PR #365, opened 2023-02-07
test: 💍 Add DBBrand component test

dkolba:


packages/components/src/components/brand/brand.spec.tsx
```diff
@@ -30,15 +45,17 @@ test.describe('DBBrand component on mobile', () => {
...
-		}
-	);
+	test('DBBrand should not have any automatically detectable accessibility issues', async ({
+		page,
+		mount
+	}) => {
+		await mount(comp);
+		const accessibilityScanResults = await new AxeBuilder({ page })
+			// TODO: Check whether a default background color in DBBrand makes sense
+			.disableRules('color-contrast')
```
mfranzke, line 55: Is this, why you need to deactivate the check because playwright elsewhere couldn't test for that aspect?

```suggestion
			// TODO: Check whether a default background color in DBBrand makes sense – previous to that we need to deactivate this test, as it's not testable standalone
			.disableRules('color-contrast')
```

dkolba: DBBrand is transparent, so we get whatever color is in the foreground and whatever color is in the background bleeding into the foreground. In this case, it's the default background of the test setup (grey) and the color of the text in DBBrand (#282d37 a greyish black), hence axe complains about color contrast.
Either we ignore it (in this case its only specific to the test setup, or we give DBBrand different text color or background color.

PR #366, opened 2023-02-07
feat: add config to work with vModel

nmerget:


showcases/vue-showcase/src/App.vue
```diff
@@ -59,7 +66,14 @@ function onClick() {
 					placeholder="irgendein Text"
 					iconAfter="heart"
 					id="input-reg"
+					:value="modelAndChange"
```
annsch, line 69: @nmerget where should we place the documentation relevant topics e.g. using v-model with attributes (and not the default modelValue prop) – maybe using a README.md for input?

nmerget: Guess we have to clear this in general. UX said we should move everything into the mapo. But what's about development specific stuff? idk...

nmerget: Let's fix this issue here: https://github.com/db-ui/mono/issues/398

Maybe we can automate everything :)

annsch (approved):

PR #372, opened 2023-02-08
feat: add infotext component

nmerget:


packages/components/src/components/infotext/model.ts
```diff
@@ -0,0 +1,25 @@
+import { GlobalProps, GlobalState } from '../../shared/model';
+
+export interface DBInfotextDefaultProps {
+	icon?: string;
+	size?: 'medium' | 'small';
+
+	title?: string;
+	variant?: 'adaptive' | 'critical' | 'information' | 'warning' | 'success'; // TODO: Move this to global props
```
mfranzke, line 8: could you want to do this "move" now or do you want to do it later ?

packages/components/src/components/infotext/infotext.scss
```diff
@@ -0,0 +1,43 @@
...
+			@if ($variant == "critical") {
+				color: $db-colors-critical-enabled;
+			} @else if($variant == "information") {
+				color: $db-colors-information-enabled;
+			} @else if($variant == "warning") {
+				color: $db-colors-warning-enabled;
+			} @else if($variant == "success") {
+				color: $db-colors-success-enabled;
+			}
+		}
```
mfranzke, line 41: you could as well define key/value pairs on line 5 and that for use a simple each than here:
https://sass-lang.com/documentation/values/maps#do-something-for-every-pair

packages/components/src/components/infotext/infotext.lite.tsx
```diff
@@ -0,0 +1,62 @@
...
+	});
+
+	onMount(() => {
+		if (props.stylePath) {
+			state.stylePath = props.stylePath;
+		}
+	});
+
+	return (
+		<div
```
mfranzke, line 45: Is this really supposed to be a block level element, or isn't it inline?

```suggestion
		<span
```

packages/components/src/components/infotext/infotext.lite.tsx
```diff
@@ -0,0 +1,62 @@
...
+			data-size={props.size}>
+			<Show when={state.stylePath}>
+				<link rel="stylesheet" href={state.stylePath} />
+			</Show>
+			<Show when={props.size !== 'small'}>
+				<DBIcon
+					icon={state.getIcon(props.icon, props.variant)}></DBIcon>
+			</Show>
+			{props.children}
+		</div>
```
mfranzke, line 60:
```suggestion
		</span>
```

packages/components/src/components/infotext/infotext.lite.tsx
```diff
@@ -0,0 +1,62 @@
...
+			if (icon) return icon;
+			switch (variant) {
+				case 'critical':
+					return 'error';
+				case 'warning':
+					return 'error-triangle';
+				case 'success':
+					return 'check-circle';
+				default:
+					return 'info';
```
mfranzke, line 33: let's align with the colleagues whether this mapping is even also a global one, that we could persist at a more global place ?

mfranzke (dismissed):

packages/components/src/components/infotext/infotext.lite.tsx
annsch: @mfranzke I had the same question when implementing this for RI UI. In case of RI UI (for example) we need this to be a block element. We could discuss if this could be a data attribute as well?

packages/foundations/scss/_body-styles.scss
```diff
@@ -0,0 +1,8 @@
+@mixin body-style($size) {
+	--db-base-icon-font-size: var(--db-base-icon-font-size-#{$size});
+	--db-base-icon-font-family: var(--db-base-icon-font-family-#{$size});
+	/* prettier-ignore */
+	--db-base-icon-font-family-filled: var(--db-base-icon-font-family-filled-#{$size});
+	line-height: var(--db-type-body-line-height-#{$size});
+	font-size: var(--db-type-body-font-size-#{$size});
```
mfranzke, line 7: why do you both set variables as well as declarations (instead of e.g. only variables, and the `line-height` and `font-size` only once at a global level / selector)? This part seems to be that global and separate from infotext, that it might be a good candidate for a separate PR (with a description).

mfranzke (changes requested):

packages/foundations/scss/_body-styles.scss
nmerget: https://github.com/db-ui/mono/pull/397

mfranzke (dismissed):

mfranzke (approved):

PR #373, opened 2023-02-08
fix: issue with brand

nmerget:


packages/components/src/components/brand/brand.spec.tsx
```diff
@@ -3,18 +3,19 @@ import AxeBuilder from '@axe-core/playwright';
...
-		imgAlt="The Deutsche Bahn logo"
-		imgHeight={50}
-		imgWidth={35}
 		anchorChildren={true}
 		anchorRef="/"
 		anchorTitle="I am hover text"
-		anchorRelation="alternate">
+		anchorRelation="alternate"
+		hideDefaultAsset={true}>
+		<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 40 28" id="logo">
```
dkolba, line 13: Haha, so simple. Love it

packages/components/src/components/brand/brand.lite.tsx
```diff
@@ -36,19 +34,21 @@ export default function DBBrand(props: DBBrandProps) {
...
 				title={props.anchorTitle}
 				rel={props.anchorRelation}>
-				<img
-					src={props.imgSrc ?? DEFAULT_VALUES.src}
-					alt={props.imgAlt}
-					height={props.imgHeight}
-					width={props.imgWidth}
-					className="db-logo"
-				/>
+				<Show when={!props.hideDefaultAsset}>
```
mfranzke, line 37: you can't differentiate by `props.children` here?

packages/components/src/components/brand/brand.spec.tsx
```diff
@@ -3,18 +3,19 @@ import AxeBuilder from '@axe-core/playwright';
...
-		imgAlt="The Deutsche Bahn logo"
-		imgHeight={50}
-		imgWidth={35}
 		anchorChildren={true}
 		anchorRef="/"
 		anchorTitle="I am hover text"
-		anchorRelation="alternate">
+		anchorRelation="alternate"
+		hideDefaultAsset={true}>
+		<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 40 28" id="logo" aria-hidden="true">
```
mfranzke, line 13:
```suggestion
		<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 40 28" id="logo" aria-hidden="true" class="db-logo">
```

packages/components/src/components/brand/brand.scss
```diff
@@ -7,26 +7,28 @@
...
-			}
-		}
+	svg {
+		@extend %logo-height;
 	}
 }
+
+.db-logo {
+	@extend %logo-height;
+}
```
mfranzke, line 34: why not setting `db-logo` as class on the SVG as well to prevent the need for a SCSS placeholder in this case?

mfranzke (changes requested):

packages/components/src/components/brand/brand.lite.tsx
nmerget: Nope, because you can use children as your normal app name and the default image

packages/components/src/components/brand/model.ts
```diff
@@ -19,6 +19,11 @@ export interface DBBrandDefaultProps {
 	 * The rel attribute can be set to make a relationship to the current URL (https://developer.mozilla.org/en-US/docs/web/api/htmlanchorelement/rel).
 	 */
 	anchorRelation?: string;
+
+	/**
+	 * This is for testing the component with a svg instead of an asset.
```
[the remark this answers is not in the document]
nmerget: fixed

packages/components/src/components/brand/brand.spec.tsx
nmerget: fixed

packages/components/src/components/brand/brand.scss
nmerget: fixed

dkolba (approved):

PR #389, opened 2023-02-09
chore: update colors from zeplin

nmerget:
Some colors changed by UX definitions and for accessibility.

Added "elements" color to have extra colors for icons in some components like `infotext`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11725,18 +11725,6 @@
         "@esbuild/win32-x64": "0.16.17"
       }
     },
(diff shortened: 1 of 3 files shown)


mfranzke:
@nmerget what's the reason for this update? Could you please provide a description?

mfranzke (changes requested):
@nmerget what's the reason for this update? Could you please provide a description?

mfranzke (approved):

PR #397, opened 2023-02-10
fix: issues with tonality (line-height and fonts)

nmerget:


packages/components/src/components/button/button.scss
```diff
@@ -117,6 +117,7 @@
 
 	&[data-size="small"] {
 		@extend %size-Small;
+		@extend %db-body-sm;
```
mfranzke, line 120: Isn't this supposed to get inserted on line 111 ?

mfranzke (approved):

mfranzke (dismissed):

mfranzke (approved):

PR #399, opened 2023-02-10
refactor: react showcase to use a default component with config

nmerget:
Add a default component to react showcase, to add new components only by configuration and to have a sample look and feel like the provided screens in zeplin.

Fixed issues for other components found by new added screens.


packages/components/src/components/button/button.scss
```diff
@@ -112,7 +112,7 @@
 
 	&%size-Small {
 		@extend %db-body-sm;
-		min-height: $db-sizing-sm;
+		height: $db-sizing-sm;
```
mfranzke, line 115: what for do you change this property? The commit message doesn't make this any clearer:
_refactor: react showcase to use a default component with config_

nmerget: moved to https://github.com/db-ui/mono/pull/408

mfranzke (approved):

PR #408, opened 2023-02-13
fix: minor issues in button, input, link & section

nmerget:


packages/components/src/components/link/model.ts
```diff
@@ -34,10 +34,10 @@ export interface DBLinkDefaultProps {
 		| 'strict-origin-when-cross-origin'
 		| 'unsafe-url';
 	selected?: boolean;
-	size?: string;
+	size?: 'small';
```
mfranzke, line 37: this declaration makes "small" the default, doesn't it? Because the guidelines mention medium to be the default.

packages/components/src/components/link/model.ts
```diff
@@ -34,10 +34,10 @@ export interface DBLinkDefaultProps {
...
 	target?: '_self' | '_blank' | '_parent' | '_top';
 	title?: string;
-	variant?: string;
+	variant?: 'primary';
```
mfranzke, line 40: this declaration makes `primary` the default, doesn't it? Because the guidelines mention `adaptive` to be the default.

packages/components/src/components/link/model.ts
nmerget: The attribute is not required, therefore the default (medium) is always set. I added it, but at the moment it is not used to separate some stylings.

packages/components/src/components/link/model.ts
nmerget: https://github.com/db-ui/mono/pull/408#discussion_r1104287579

mfranzke (approved):

PR #410, opened 2023-02-13
feat: add styling for <pre>

nmerget:


packages/foundations/scss/_init.scss
```diff
@@ -107,3 +109,10 @@ nav,
 fieldset {
 	border: none;
 }
+
+// Code
```
mfranzke, line 113: Code should especially get marked up as [`<code>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/code), whereas [`<pre>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/pre) might be any similar preformatted text like e.g. terminal output etc.

nmerget: Yes, you are right. I changed it to make it clear that it should only affect <code> but also <pre> if it contains <code>. Now it should work to have some sort of difference for a code block, especially for our documentation.

mfranzke (approved):

PR #419, opened 2023-02-14
fix: issue with card test

nmerget:


mfranzke (approved):

mfranzke (commented):
There are still further places in which 'ia' is used instead of 'interactive':
image

PR #426, opened 2023-02-14
feat: new component - alert

nmerget:


packages/components/_templates/mitosis/new/tsx.ejs.t
```diff
@@ -22,7 +22,7 @@ export default function DB<%= h.changeCase.pascal(name) %>(props: DB<%= h.change
   });
 
   return (
-    <div className={'db-<%= name %>' + (props.className ? ' ' + props.className : '')}>
+    <div class={'db-<%= name %>' + (props.className ? ' ' + props.className : '')}>
```
mfranzke, line 25: @nmerget doesn't this fix a problem in general that we would need in main independently from this new component as a fix for ongoing component development right away?

nmerget: https://github.com/db-ui/mono/pull/437 <- yes, moved it

.jscpd.json
```diff
@@ -24,7 +24,9 @@
 		"**/angular.json",
 		"**/tsconfig.json",
 		"**/angular-*/**/main.ts",
-		"**/README.md"
+		"**/README.md",
+		"packages/components/src/components/alert/alert.lite.tsx",
+		"showcases/react-showcase/src/components/alert/index.tsx"
```
mfranzke, line 29: Does this relate to this component even only? Or does it need to be a (more) glob(al) declaration in here?
And if it's repetitive code, couldn't we even also store it more globally to not repeat it?

I couldn't easily identify it by a quick look.

nmerget: It's just this one. I use DBLink twice with nearly the same configurations, which is detected as copy & paste.
In React we could use the spread operator (...props) to pass the props to the Link but it's not possible in mitosis at the moment: https://github.com/BuilderIO/mitosis/blob/main/docs/gotchas.md#cant-destructure-props-as-rest

packages/components/src/components/alert/alert.lite.tsx
```diff
@@ -0,0 +1,111 @@
...
+		<div
+			class={'db-alert' + (props.className ? ' ' + props.className : '')}
+			data-variant={props.variant}
+			data-type={props.type}>
+			<Show when={state.stylePath}>
+				<link rel="stylesheet" href={state.stylePath} />
+			</Show>
+			<div class="db-alert-icon-container">
+				<DBIcon icon={state.getIcon(props.icon, props.variant)} />
+			</div>
```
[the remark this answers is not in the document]
nmerget: fixed

packages/components/src/components/alert/alert.scss
```diff
@@ -0,0 +1,96 @@
...
+		@include handle-links(false);
+	}
+
+	padding: $db-spacing-fixed-md;
+	display: flex;
+	gap: $db-spacing-fixed-md;
+	width: 100%;
+
+	.db-icon {
+		&::before {
```
annsch, line 47: This style would have no effect in web components: according to https://github.com/db-ui/mono/pull/456 we should try to making our components more resilient (thx @mfranzke) and avoid styling pseudo elements esp. for child components.

nmerget: How do we do it?

annsch: just remove the &::before. font-style and -size should affect the pseudo-element anyway @nmerget

nmerget: But if the filled icons works, I need to overwrite the icon family? :O

annsch: as discussed we should refactor this separately in DBIcon: https://github.com/db-ui/mono/issues/497

annsch (approved):

PR #438, opened 2023-02-15
fix(input): filled icons for inputstate

### LINKED ISSUE #293: fix: icons for input field
- [ ] icons for data-variants should contain filled icons
- [ ] invalid input does not include the error icon when input field is aria-invalid

annsch:
closes #293

### DIFF
--- a/showcases/react-showcase/src/components/input/index.tsx
+++ b/showcases/react-showcase/src/components/input/index.tsx
@@ -86,9 +86,10 @@ const variants: DefaultComponentVariants[] = [
 				name: 'Invalid',
 				example: (
 					<DBInput
-						label="MinLength = 5"
+						label={defaultLabelText}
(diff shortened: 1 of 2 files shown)


mfranzke (approved):

showcases/react-showcase/src/components/input/index.tsx
```diff
@@ -82,6 +82,16 @@ const variants: DefaultComponentVariants[] = [
 					/>
 				)
 			},
+			{
+				name: 'Invalid',
+				example: (
+					<DBInput
+						label="MinLength = 5"
```
mfranzke, line 89: what about adding this as description instead ?

annsch: fixed

mfranzke (approved):

PR #441, opened 2023-02-15
refactor: header to work with 3.0 navigation

nmerget:
## Proposed changes

- refactored `db-header`
- added `db-main-navigation` as connection between `db-header` and `db-navigation-item`

## Types of changes


packages/components/src/components/header/header.scss
```diff
@@ -1,48 +1,150 @@
...
-	.meta-navigation {
-		display: none;
+.db-header-drawer {
+	.db-header-drawer-backdrop {
+		background-color: $db-colors-neutral-on-bg-enabled;
+		opacity: 0.5;
+		width: 100%;
+		height: 100%;
+		position: fixed;
+		z-index: 9995;
```
mfranzke, line 26: you should most likely be able to insert stuff like this with CSS only by `::before { content: ""; … }`, especially to clean up the HTML code.

nmerget: add `<dialog>` to handle this issue

packages/components/_templates/mitosis/new/readme.ejs.t
```diff
@@ -1,7 +1,7 @@
 ---
 to: src/components/<%= name %>/README.md
 ---
-How to use DB<%= h.changeCase.pascal(name) %>
+# How to use DB<%= h.changeCase.pascal(name) %>
```
annsch, line 5: isn't this a separate MR? ;)

packages/components/overrides/angular/src/components/header/action-bar.directive.ts
```diff
@@ -0,0 +1,6 @@
+import { Directive } from '@angular/core';
+
+@Directive({
+	selector: '[dbActionBar]'
+})
+export class ActionBarDirective {}
```
annsch, line 6: What about some comments why we need those overwrites? For me it's not clear, why we need them.

packages/components/src/components/brand/brand.scss
```diff
@@ -18,9 +18,10 @@
 	}
 }
 
+.db-brand-title {
+	display: contents;
```
annsch, line 22: really smart to use this ;)

packages/components/_templates/mitosis/new/readme.ejs.t
nmerget: @mfranzke did it already: https://github.com/db-ui/mono/commit/59b3551f58a5757a9da0b58fc16a09ffddfa4c91

packages/components/overrides/angular/src/components/header/action-bar.directive.ts
nmerget: add comments for those directives

packages/components/src/components/brand/brand.scss
nmerget: add a comment, it is just for internal testing :(

README.md
mfranzke, line 1: This files entry is missing an update on the `header`:
https://github.com/db-ui/mono/pull/441/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R77

packages/components/src/components/brand/brand.lite.tsx
```diff
@@ -86,11 +86,11 @@ export default function DBBrand(props: DBBrandProps) {
 					/>
 				</Show>
 				<Show when={props.anchorChildren}>
-					<>{props.children}</>
+					<strong class="db-brand-title">{props.children}</strong>
```
mfranzke, line 89:
```suggestion
					<>{props.children}</>
```

Let's do this via CSS even only.

packages/components/src/components/brand/brand.lite.tsx
```diff
@@ -86,11 +86,11 @@ export default function DBBrand(props: DBBrandProps) {
...
 				</Show>
 			</a>
 			<Show when={!props.anchorChildren}>
-				<>{props.children}</>
+				<strong class="db-brand-title">{props.children}</strong>
```
mfranzke, line 93:
```suggestion
				<>{props.children}</>
```

Let's do this with CSS even only.

packages/components/src/components/brand/brand.scss
```diff
@@ -18,10 +18,12 @@
 	}
 }
 
+.db-brand-title {
+	/* This is a workaround for internal tests to show svg inside the strong tag */
+	display: contents;
+}
+
```
mfranzke, line 25:
```suggestion
.db-brand-title {
	/* This is a workaround for internal tests to show svg inside the strong tag */
	display: contents;
}

```

Not necessary anymore, as we won't need the strong tag.

packages/components/src/components/brand/brand.scss
```diff
@@ -18,10 +18,12 @@
 	}
```
mfranzke, line 18:
```suggestion
	font-weight: 700;
	}
```

packages/components/src/components/drawer/README.md
mfranzke, line 1: @nmerget From my understanding, this is still the old documentation format, isn't it ?

packages/components/src/components/header/header.lite.tsx
```diff
@@ -1,55 +1,132 @@
...
+			<div class="db-header-meta-navigation db-header-hide-on-mobile">
 				<Slot name="meta-navigation" />
 			</div>
-			{props.children}
+			<div class="db-header-navigation-bar">
+				<div class="db-header-brand-container">
+					<Slot name="brand" />
+					<DBDivider
+						className="db-header-divider db-header-hide-on-mobile"
+						variant="vertical"></DBDivider>
```
mfranzke, line 101: It should be possible to implement this with a simple `::after { content: "" }` rule on the `.db-header-brand-container`, doesn't it ?

packages/components/src/components/header/header.lite.tsx
```diff
@@ -1,55 +1,132 @@
...
+						variant="vertical"
+						margin="none"></DBDivider>
+					<DBButton
+						id="button-burger-menu"
+						className="db-header-hide-on-desktop"
+						icon="menu"
+						noText
+						variant="text"
+						onClick={() => state.toggle()}>
+						Burger Menu
```
mfranzke, line 123: @nmerget Let's make it possible to control this from the outside. Maybe the included label is a little bit too generic.

packages/components/src/components/drawer/drawer.scss
```diff
@@ -81,12 +81,18 @@ $spacings: (
...
+		);
 		padding-inline: map.get($spacing, "inline");
 	}
 
 	.db-drawer-content {
-		padding-inline: map.get($spacing, "inline");
+		padding-inline: var(
+			--db-drawer-content-padding-inline,
+			map.get($spacing, "inline")
+		);
```
mfranzke, line 95: I like this approach to provide a `Styling API` to this component by providing these `CSS Custom Properties`, but:
- Have we already decided how we document those?
- Are those requested / aligned with the creative colleagues, or "just" identified as relevant by us / DEVs (don't get me wrong, it's still a good initiative, but it should be aligned for consistency)?

packages/components/src/components/drawer/README.md
nmerget: removed it

packages/components/src/components/header/header.lite.tsx
```diff
@@ -1,55 +1,132 @@
...
+					</div>
+					<div class="db-header-call-to-action">
+						<Slot name="call-to-action" />
+					</div>
+				</div>
+				<div class="db-header-action-container">
+					<DBDivider
+						className="db-header-divider"
+						variant="vertical"
+						margin="none"></DBDivider>
```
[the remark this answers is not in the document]
nmerget: done

packages/components/src/components/header/header.lite.tsx
nmerget: done

packages/components/src/components/header/header.lite.tsx
nmerget: done

packages/components/src/components/drawer/drawer.scss
nmerget: 1) Nope
2) Nope, this could be helpful. I guess the best would be some cli tool that generates an `.md` file with all css-properties based on the description above. But I couldn't find something like this.

But we should handle this in a new issue for all components

nmerget: https://github.com/db-ui/mono/issues/1349

packages/components/src/components/brand/brand.scss
```diff
@@ -10,18 +10,18 @@
...
 	@extend %brand-display;
 
+	font-weight: 700;
+
 	a {
 		@extend %brand-display;
 		color: inherit;
 		// Deactivate underline on site name being made clickable by added to the link as well
 		text-decoration: none;
+		font-weight: 700;
```
mfranzke, line 20:
```suggestion
```

This property is inherited already.

packages/components/src/components/header/docs/Angular.md
```diff
@@ -18,9 +24,59 @@ import { DBHeaderModule } from '@db-ui/ngx-components';
...
+	toggleDrawer = (open: boolean) => {
+		this.drawerOpen = open;
+	};
+}
+```
+
+```html app.component.html
+<!-- app.component.html -->
+<db-header [drawerOpen]="drawerOpen" (onToggle)="toggleDrawer($event)">
+	<db-brand brand>My Awesome App</db-brand>
```
mfranzke, line 59: @nmerget what's that `brand` attribute included for?

packages/components/src/components/header/header.lite.tsx
```diff
@@ -1,55 +1,128 @@
...
+					<div class="db-header-meta-navigation">
+						<Slot name="meta-navigation" />
+					</div>
+				</div>
+				<div class="db-header-action-bar">
+					<Slot name="action-bar" />
+				</div>
+			</DBDrawer>
+
+			<div class="db-header-meta-navigation db-header-hide-on-mobile">
```
mfranzke, line 95: We shouldn't include styling related information like e.g. `db-header-hide-on-mobile` in structuring markup.

packages/components/src/components/header/docs/Angular.md
nmerget: Yes that's the angular way

packages/components/src/components/header/header.lite.tsx
nmerget: I guess we could move it upwards into foundations with `[data-hide-on-mobile]` and `[data-hide-on-desktop]`?

I need this here because I need to move all the related slots into the drawer...

mfranzke: I've opened a follow up ticket for this: https://github.com/db-ui/mono/issues/1393

packages/components/src/components/select/select.scss
```diff
@@ -19,7 +19,7 @@
 
 	&:has(select:focus-within) {
 		&::after {
-			transform: rotate(180deg);
```
mfranzke, line 22: what's the difference in between both declarations? The direction the arrow animates to turn around ?

packages/components/src/components/page/page.scss
```diff
@@ -20,6 +20,17 @@
...
+	opacity: 0;
+
+	&[data-fade-in="true"] {
+		transition: opacity $db-transition-emotional-straight;
+	}
+
+	&:not([data-fonts-loaded]),
+	&[data-fonts-loaded="true"] {
+		opacity: 1;
+	}
```
mfranzke, line 32: I think we need to realign about this again – I don't understand the reason why to emulate this via JavaScript, as it should be possible to set it via CSS even only through `font-display`

packages/components/src/components/page/model.ts
```diff
@@ -4,10 +4,17 @@ export interface DBPageDefaultProps {
 	type?: 'fixedHeaderFooter';
 	slotHeader?: any;
 	slotFooter?: any;
+
+	/**
+	 * Set this to have a transition with opacity to avoid layout-shifts https://simonhearne.com/2021/layout-shifts-webfonts/
+	 */
+	fadeIn?: boolean;
```
mfranzke, line 11: I think we need to realign about this again – I don't understand the reason why to emulate this via JavaScript, as it should be possible to set it via CSS even only through `font-display`. Plus that this solution depends on JavaScript and doesn't have a graceful degradation. And hide all parts of the page, not just text. So I think it's an interesting idea, but either don't understand all requirements for it and/or would prefer to solve it via CSS even only.

The article itself mentions the following:
> The dirty little secret of this blog post is that you can resolve layout shifts due to fonts with a single line of CSS:

> `font-display: optional`

packages/components/src/components/header/docs/React.md
```diff
@@ -12,3 +14,51 @@ const App = () => <DBHeader slotBrand={<DBBrand>Header</DBBrand>} />;
...
+					Notification
+				</DBButton>
+				<DBButton icon="help" variant="text" noText>
+					Help
+				</DBButton>
+			</>
+		}
+	>
+		<DBMainNavigation>
+			Look at the docs for DBMainNavigation
```
mfranzke, line 58: if we mention something like this, indicating a reference, let's link it directly.

packages/components/src/components/header/header.scss
```diff
@@ -1,44 +1,189 @@
...
-	padding-inline: $db-spacing-fixed-md;
+	min-block-size: $default-mobile-header-height;
 
 	.db-link {
 		display: inline-block;
 	}
 
-	.desktop-navigation,
-	.meta-navigation {
+	@media screen and (min-width: $db-screens-m) {
```
mfranzke, line 21: you've introduced the `screen()` SCSS function – it's probably related to another context, as it also includes other selectors at the moment, but what about providing a more generic one (additionally)?

packages/components/src/components/header/header.scss
```diff
@@ -1,44 +1,189 @@
...
+		--db-drawer-max-width: #{$db-screens-s};
+	}
+
+	&[data-on-forcing-mobile="true"] {
+		visibility: hidden;
+	}
```
mfranzke, line 27: I probably don't fully understand the mechanics behind this, but a quick comment on hiding this visibly only: Hiding this via `visibility: hidden;` would mean that it still gets read out to screenreader users, which is normally not what we want – any content, especially more complex constructs like e.g. navigations / menus that are visually hidden from the user should also get hidden for screenreaders. The easiest would be to use `display: none` for this, or instead the `hidden` attribute; but again, I'm not 100% sure whether I got it right.

packages/components/src/components/header/header.scss
```diff
@@ -1,44 +1,189 @@
...
+}
+
+.db-header-navigation-bar {
+	display: flex;
+	position: relative;
+	padding-block: $db-spacing-fixed-xs;
+	padding-inline: $db-spacing-fixed-md;
```
mfranzke, line 34:
```suggestion
	padding: $db-spacing-fixed-xs $db-spacing-fixed-md;
```

packages/components/src/components/header/docs/Vue.md
```diff
@@ -16,3 +18,51 @@ import { DBHeader, DBBrand } from "@db-ui/v-components";
...
+				Help
+			</DBButton>
+		</template>
+		<template v-slot:meta-navigation>
+			<DBLink href="#">Imprint</DBLink>
+			<DBLink href="#">Help</DBLink>
+		</template>
+
+		<DBMainNavigation>
+			Look at the docs for DBMainNavigation
```
mfranzke, line 64: https://github.com/db-ui/mono/pull/441/files#r1289915373

packages/components/src/components/input/input.scss
```diff
@@ -119,7 +119,7 @@
 
 		&:has(input:is(:focus, :not(:placeholder-shown))) {
 			&::after {
-				transform: rotate(180deg);
+				transform: rotate(-180deg);
```
mfranzke, line 122: https://github.com/db-ui/mono/pull/441#discussion_r1289899081

packages/components/src/components/select/select.scss
nmerget: It should rotate outward (with a minus) and with a positive value it rotates clock-wise

packages/components/src/components/header/header.scss
nmerget: fixed

packages/components/src/components/page/model.ts
mfranzke: I misinterpreted this implementation, it's only an optional property.

packages/components/src/components/header/docs/Vue.md
nmerget: fixed

packages/components/src/components/header/docs/React.md
nmerget: fixed

packages/components/src/components/header/header.scss
nmerget: Visibility is correct here. I don't want to hide the header from screenreaders but I want to hide the header for a short time from the screen to add "force-mobile" if the user passes the property `force-mobile=true` to the header component.

This is for Apps that want mobile navigation, even for desktop devices.

packages/components/src/components/main-navigation/docs/HTML.md
```diff
@@ -0,0 +1,55 @@
...
+						<a href="#">Sub-Navi-Item 2</a>
+					</li>
+				</menu>
+			</li>
+
+			<li class="db-navigation-item" data-icon="account">
+				<a href="#">Navi-Item 2</a>
+			</li>
+
+			<li class="db-navigation-item" data-disabled="true">
```
mfranzke, line 49:
```suggestion
			<li class="db-navigation-item" aria-disabled="true">
```

This seems to have changed in the meantime, doesn't it?

packages/components/src/components/main-navigation/main-navigation.scss
```diff
@@ -0,0 +1,88 @@
+@use "@db-ui/foundations/build/scss/variables" as *;
+@use "@db-ui/foundations/build/scss/variables.global" as *;
+@use "@db-ui/foundations/build/scss/helpers/component" as *;
+@use "../../styles/db-puls" as *;
+@use "../../styles/form-components" as *;
+
+.db-main-navigation {
+	-webkit-tap-highlight-color: transparent; /* for removing the highlight */
```
mfranzke, line 8: what's the reason for this declaration?

nmerget: It is for mobile devices only:

When you click on a button for example there is a short overlay container (light blue) indicating that the item was clicked.
But in the current mobile navigation, we open up another drawer layer. With the tap-highlight enabled the button below the new drawer layer will be highlighted which looks very wired. Moreover, the new drawer layer is an indicator as well. So we don't need this tap highlight in the navigation.

But it makes sense for other components like the regular button etc., and we should check if we want to use a custom color for this, like "informational". But this is fine-tuning.

mfranzke (approved):

PR #456, opened 2023-02-16
refactor(input): making input more resilient

annsch:
- removed styles for pseudo-elements to prepare input web component fix

### DIFF
--- a/packages/components/src/components/input/input.scss
+++ b/packages/components/src/components/input/input.scss
@@ -11,10 +11,6 @@
 @use "@db-ui/foundations/build/scss/color/color-variants" as *;
 @use "@db-ui/foundations/build/scss/helpers/component";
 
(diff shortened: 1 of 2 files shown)


packages/components/src/components/input/input.scss
```diff
@@ -147,22 +146,20 @@
 			$db-spacing-fixed-xs var(--db-input-padding-inline-start);
 
 		// TODO: Evaluate whether those could get moved to ../_form-elements.scss
-		&:focus,
+		&:focus-visible,
```
mfranzke, line 149: why does this need to be a visible focus ?

annsch: @mfranzke changed this. For input elements this should be :focus

nmerget (approved):

PR #488, opened 2023-02-20
feat: add new colors from zeplin

nmerget:
- Update colors from zeplin
- Fixed background colors for interactive elements with RGB values (button & input)
- Update variants for button based on new colors


packages/components/src/components/button/button.scss
```diff
@@ -18,6 +19,7 @@
 
 	border: 1px solid transparent;
 	border-radius: 4px;
+	color: var(--db-current-color, $db-colors-neutral-on-bg-enabled);
```
annsch, line 22: Isn't this something we could also provide als global placeholder? (border, border-radius and color in combination)

nmerget: fixed

annsch (dismissed):
as discussed the transparent border for elements (e.g. input) will change in the future by separate token

packages/components/src/styles/_component_defaults.scss
```diff
@@ -0,0 +1,27 @@
...
+}
+
+%default-interactive-component {
+	@extend %default-adaptive-border;
+	@extend %default-border-radius;
+	@extend %default-color;
+}
+
+%default-background-transition {
+	// Changing background-color on hover should be animated by a transition
```
annsch, line 24:
```suggestion
	// TODO Changing background-color on hover should be animated by a transition
```

annsch (approved):

packages/foundations/scss/helpers/_component.scss
```diff
@@ -0,0 +1,27 @@
...
+@use "../color-placeholder" as *;
+@use "../helpers/functions" as *;
+
+%default-adaptive-border {
+	border: solid 1px var(--db-current-color, $db-colors-neutral-on-bg-enabled);
+}
+
+%default-border-radius {
+	border-radius: to-rem(4);
+}
```
mfranzke, line 11: instead of a placeholder with only one property we could define a Design Token that could as well get used on other technical contexts.

PR #489, opened 2023-02-20
refactor: set pipelines and local node versions equally

mfranzke:
Not using the same Node version leads to problems in the pipeline that weren't occuring on localhost, like e.g. the following in DB UI Elements repository: https://github.com/db-ui/elements/actions/runs/4220820918/jobs/7327485768

We're using the `init cache` action as the basis as this perfectly separates workflows / workflow steps that we even also run locally and the ones that are only run in the pipeline; great idea, thanks to @nmerget.

Fixes https://github.com/db-ui/mono/issues/474


dkolba (dismissed):

nmerget:
Why don't you just add this to the "Init Cache" Action?

mfranzke:
> https://github.com/db-ui/elements/actions/runs/4220820918/jobs/7327485768

As I understand it errors like https://github.com/db-ui/elements/actions/runs/4220820918/jobs/7327485768 result out of the incorrect Node version on runtime / executing our code and that out of dependencies, and not installation (in which we most likely have the same `npm` version).

nmerget:
> Member

Yes but "Init Cache" is called on every workflow where we use node. So we could move it into this. Moreover you have an if statement which is always false, because the step you use is right below the "node-version" step

mfranzke:
> > Member
> 
> Yes but "Init Cache" is called on every workflow where we use node. So we could move it into this. Moreover you have an if statement which is always false, because the step you use is right below the "node-version" step

I shouldn't finalize PRs, when I'm out of condition to do so. Thanks for that catch.

nmerget (approved):

PR #499, opened 2023-02-21
feat: add show code for showcase react

nmerget:
- Add new query param for mapo iframes
- Add show code button


packages/components/src/components/code-docs/code-docs.scss
```diff
@@ -0,0 +1,57 @@
...
+
+	.code {
+		@extend %db-bg-secondary;
+		position: absolute;
+		bottom: 0;
+		right: 0;
+		left: 0;
+		top: 0;
+		border-radius: 8px;
+		display: flex;
```
mfranzke, line 34:
```suggestion
		display: flex;
		justify-content: space-around;
```

having a better vertical rhythm especially on [long contents](https://db-ui.github.io/mono/review/feat-react-show-code/react-showcase/index.html#/alert?tonality=regular&color=neutral-0&page=variant) might be beneficial.

packages/components/src/components/code-docs/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+TODO
+
+## React
+
+TODO
+
+## Vue
+
+TODO
```
mfranzke, line 13: this ones missing the `README`

packages/components/src/components/code-docs/code-docs.lite.tsx
```diff
@@ -0,0 +1,94 @@
...
+							)}
+						</For>
+					</Show>
+				</div>
+			</Show>
+			<DBButton
+				class="code-button"
+				size="small"
+				variant="primary"
+				onClick={() => state.toggleCode()}>
```
mfranzke, line 89: it might be much simpler to use `<details><summary>Show Code</summary> … Code … </details>` here and w/o the need for JavaScript. Besides that the current implementation isn't accessible, as there's no relation in between the clicked `button` and the "opened" code panel / code being shown.

packages/components/src/components/link/link.scss
```diff
@@ -10,6 +10,8 @@
 
 .db-link {
 	@extend %default-color;
+	cursor: pointer;
```
mfranzke, line 13:
```suggestion

```

I would expect that the `a` HTML tag has this CSS declaration out of the box already.

packages/components/src/components/link/link.scss
```diff
@@ -10,6 +10,8 @@
...
+	width: fit-content;
```
mfranzke, line 14: what for is this declaration?

packages/components/src/components/link/link.scss
nmerget: The `a` has it only if we use `href` and there might be cases where we don't need a real link, only the styling

showcases/patternhub/data/routes.ts
```diff
@@ -1,6 +1,21 @@
 import type { DbMainnavigationDataType } from '@db-ui/elements/dist/types/components/db-mainnavigation/db-mainnavigation-type';
 
 const componentChildren = [
+	{
+		label: 'TODO: Add to specific category; DBCodeDocs',
```
mfranzke, line 5: Do you want to add this category right away or open up a follow up task?

showcases/react-showcase/src/app.tsx
```diff
@@ -1,43 +1,24 @@
...
-					setTonality(value);
-				}
-
-				if (key === COLOR_CONST && color !== value) {
-					setColor(value);
-				}
-			}
-		}
-	}, [searchParameters]);
+	const [tonality, setTonality, color, setColor, pageName, , fullscreen] =
```
mfranzke, line 12: Is the empty entry within `pageName, , fullscreen` on purpuse?

showcases/react-showcase/src/components/index.tsx
```diff
@@ -1,51 +1,62 @@
...
+					key={`${example.name}-${exampleIndex}`}
+					style={example.style}
+					className={example.className}>
+					<p className="example-name">
+						<small>{example.name}</small>
+					</p>
+					{example.example}
+				</li>
+			))}
+		</ul>
```
mfranzke, line 26: Using a `<dl><dt></dt><dd></dd> … <dt></dt><dd></dd> … </dl>` here might be much simpler, compare to https://jsbin.com/movupejupa/edit?html,css,output

showcases/react-showcase/src/components/index.tsx
```diff
@@ -1,51 +1,62 @@
...
-								<p>
-									<small>{example.name}</small>
-								</p>
-								{example.example}
-							</li>
-						))}
-					</ul>
+					<DBLink
+						content="external"
+						onClick={() => setPageName(variant.name.toLowerCase())}>
```
mfranzke, line 56: please use `href` here, as this is not a link elsewhere.

showcases/react-showcase/src/components/index.tsx
```diff
@@ -1,51 +1,62 @@
...
+						<strong>{variant.name}</strong>
```
mfranzke, line 57:
```suggestion
						{variant.name}
```

This could better achieve via CSS.

showcases/react-showcase/src/app.tsx
nmerget: Yes there would be `setPageName` and we don't need it here

packages/components/src/components/link/link.scss
mfranzke: than it's not a link anymore, but should be a `button`.

packages/components/src/components/link/link.scss
nmerget: db-link has `display:flex` so the click area is 100% of the parent container, which is not what we want as link:

image

packages/components/src/components/code-docs/README.md
nmerget: fixed for react

packages/components/src/components/link/link.scss
nmerget: okay changed it for this case

I guess we have to think about the framework-specific routers and db-link :S

showcases/react-showcase/src/components/index.tsx
nmerget: done

packages/components/src/components/code-docs/code-docs.lite.tsx
nmerget: fixed

showcases/patternhub/data/routes.ts
nmerget: removed it, we should not show it to the users, it's just internal

packages/components/src/components/code-docs/code-docs.lite.tsx
```diff
@@ -0,0 +1,94 @@
...
+				'db-code-docs' + (props.className ? ' ' + props.className : '')
+			}>
+			<Show when={state.stylePath}>
+				<link rel="stylesheet" href={state.stylePath} />
+			</Show>
+			{props.children}
+			<details class="code-details" onToggle={() => state.toggleCode()}>
+				<summary
+					class="db-button code-button"
+					role="button"
```
mfranzke, line 54:
```suggestion

```

This is already the default role of this element: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/summary#technical_summary

packages/components/src/components/code-docs/code-docs.lite.tsx
```diff
@@ -0,0 +1,94 @@
...
+						</DBInfotext>
+					</Show>
+					<Show
+						when={
+							props.codeSnippets && props.codeSnippets.length > 0
+						}>
+						<For each={props.codeSnippets}>
+							{(snippet: string) => (
+								<pre key={state.getSnippetId()}>
+									<code class="language-typescript">
```
mfranzke, line 76: Why do you use an additional `<pre>` element here, whereas I would expect that the `<code>` element should be already enough?

showcases/react-showcase/src/components/index.tsx
```diff
@@ -1,51 +1,67 @@
...
+			.filter((example) => example.code)
+			.map((example) => `/* ${example.name} */\n${example.code}`)}>
+		<dl className="variants-list">
+			{examples.map((example, exampleIndex) => (
+				<div
+					key={`${example.name}-${exampleIndex}`}
+					style={example.style}
+					className={example.className}>
+					<dt className="example-name">
+						<small>{example.name}</small>
```
mfranzke, line 22:
```suggestion
						{example.name}
```

Please use CSS for this styling purpose.

showcases/showcase-styles.css
```diff
@@ -12,15 +12,24 @@
...
 .variants-list {
-	list-style-type: none;
-	margin: 0;
-	padding: 0;
 	display: flex;
 	flex-wrap: wrap;
 	gap: var(--db-spacing-fixed-md);
 }
 
+dd {
```
mfranzke, line 25:
```suggestion
.variants-list dd {
```

This would be quite a global selector.

packages/components/src/components/code-docs/code-docs.lite.tsx
nmerget: I can use a `<div>` as well. But I need a container to apply the `display:grid` for the code and copy button.

mfranzke: you could apply this to `div.code` as well e.g. with a `grid-template-columns: 2fr 1fr;` declaration.

nmerget: I can't do this, because I need a key for react.
I'm not able to use empty tags with key `<key={state.getSnippetId()}><code>...</code><DBButton></DBButton><>`.

Therefore I have to wrap code and button anyways. And moving the button into `<code>` wouldn't be what I want.

mfranzke: you need to try harder …

nmerget: Add a issue for this https://github.com/db-ui/mono/issues/592

mfranzke (approved):

PR #515, opened 2023-02-23
feat(component): Checkbox

mfranzke:
resolves https://github.com/db-ui/mono/issues/362

predecessor: https://github.com/db-ui/mono/pull/825


packages/components/src/components/checkbox/model.ts
```diff
@@ -0,0 +1,41 @@
...
+	ChangeEventState,
+	ChangeEventProps,
+	GlobalProps,
+	GlobalState,
+	FormProps,
+	FormState
+} from '../../shared/model';
+
+export interface DBCheckboxDefaultProps {
+	checked?: boolean;
```
nmerget, line 15: `checked` `describedbyid` `invalid` -> they are the same in radio, move it to the shared models please

showcases/patternhub/data/routes.ts
```diff
@@ -1,6 +1,21 @@
 import type { DbMainnavigationDataType } from '@db-ui/elements/dist/types/components/db-mainnavigation/db-mainnavigation-type';
 
 const componentChildren = [
+	{
```
nmerget: You need to move this to the correct location

showcases/react-showcase/src/components/checkbox/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+			example: getCheckbox({ children: exampleName, name: 'Size' }),
+			code: '<DBCheckbox name="Size">Label</DBCheckbox>'
+		},
+		{
+			example: getCheckbox({
+				children: exampleName,
+				name: 'Size',
+				size: 'small'
+			}),
+			code: '<DBCheckbox size="small" name="Size">Label</DBCheckbox>'
```
nmerget, line 101: Something is not working correctly with a checked small checkbox. Maybe the current checkmark icon is too big?

packages/components/src/components/checkbox/checkbox.scss
```diff
@@ -0,0 +1,98 @@
...
+
+	&:checked {
+		// We need to reset the general icons margin, as we want to use it in a tiny space
+		--icon-margin-after: 0;
+
+		@extend %db-secondary-component-ia;
+
+		// Check icon
+		@include icon(glyph(done), 24, "outline");
+		/* stylelint-disable order/order */
```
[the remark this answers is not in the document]
mfranzke: it's because `::before` is added here directly after the icon mixin, but shouldn't be according to stylelint order, whereas this grouping makes perfectly sense logically.

packages/components/src/components/checkbox/model.ts
mfranzke: `describedbyid` and `invalid` will be part of https://github.com/db-ui/mono/pull/871

mfranzke: extracted `checked` within this PR.

packages/components/src/components/checkbox/checkbox.scss
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	&[data-size="small"] {
+		@extend %size-Small;
+	}
+
+	&:required {
+		& + label {
+			&::after {
+				content: " *";
```
nmerget, line 92: Do we really want to force this?

packages/components/src/components/checkbox/docs/React.md
```diff
@@ -0,0 +1,25 @@
...
+// App.tsx
+import { useState } from "react";
+import { DBCheckbox } from "@db-ui/react-components";
+
+const App = () => (
+	<DBCheckbox
+		name="checkbox"
+		value="Checkbox checked"
+		onChange={(event) => {
+			setCheckbox(event.target.checked);
```
nmerget, line 17: You forget the state inside the App component

packages/components/src/components/checkbox/docs/Vue.md
```diff
@@ -0,0 +1,26 @@
...
+For general installation and configuration look at the [v-components](https://www.npmjs.com/package/@db-ui/v-components) package.
+
+### Use component
+
+```vue App.vue
+<!-- App.vue -->
+<script setup lang="ts">
+import { DBCheckbox } from "@db-ui/v-components";
+import { ref } from "vue";
+const checkbox = ref("");
```
nmerget, line 12: This is the same example as `db-radio`...
But you are able to select multiple checkboxes, so we cannot use a string here. 
An array would be better

packages/components/src/components/checkbox/model.ts
```diff
@@ -0,0 +1,41 @@
...
+	GlobalProps,
+	GlobalState,
+	FormProps,
+	FormState,
+	FormCheckProps,
+	FormCheckState
+} from '../../shared/model';
+
+export interface DBCheckboxDefaultProps {
+	size?: 'small';
```
nmerget, line 17: You should add descriptions for this props

packages/components/src/components/checkbox/checkbox.scss
mfranzke: This is how it's specified.

packages/components/src/components/checkbox/checkbox.scss
```diff
@@ -0,0 +1,96 @@
...
+
+	align-content: center;
+
+	appearance: none;
+
+	aspect-ratio: 1;
+
+	border: calc(var(--checkbox-size, #{$db-sizing-sm}) / 16 + 0.5px) solid
+		currentColor;
+	border-radius: 3px;
```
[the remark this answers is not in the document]
mfranzke: Adapted the one from `input` elements, which is 4px.

packages/components/src/components/checkbox/checkbox.scss
mfranzke: @leape, in Sketch the `required` state seems to have been removed, has it? And/or are we still supposed to include the asterisk after the labels textual content for these mandatory fields?

packages/components/src/components/checkbox/docs/Vue.md
mfranzke: I shouldn't have copy&pasted without a thought …

packages/components/src/components/checkbox/checkbox.scss
nmerget: We solve this here: https://github.com/db-ui/mono/issues/981

nmerget (approved):

PR #517, opened 2023-02-23
feat(component): Radio

mfranzke:
Resolves https://github.com/db-ui/mono/issues/361

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ Here is an overview of all components you can use:
 | Pagination                                            | 02 - Action       |     ❌      |   ❌    |    ❌    |        ❌        |
 | Popover                                               | 06 - Utility      |     ❌      |   ❌    |    ❌    |        ❌        |
 | Progress                                              | 05 - Feedback     |     ❌      |   ❌    |    ❌    |        ❌        |
-| Radio                                                 | 03 - Data-Input   |     ❌      |   ❌    |    ❌    |        ✅        |
+| [Radio](packages/components/src/components/radio)     | 03 - Data-Input   |     ✅      |   v3    |    ❌    |        ✅        |
 | Rating                                                | 03 - Data-Input   |     ❌      |   ❌    |    ❌    |        ❌        |
 | Section                                               | 01 - Layout       |     ❌      |   ❌    |    ❌    |        ✅        |
 | Select                                                | 03 - Data-Input   |     ❌      |   ❌    |    ❌    |        ✅        |


packages/components/src/components/radio/radio.scss
```diff
@@ -0,0 +1,114 @@
+@use "@db-ui/foundations/build/scss/variables" as *;
+@use "@db-ui/foundations/build/scss/variables.global" as *;
+@use "@db-ui/foundations/build/scss/helpers/functions" as *;
+@use "@db-ui/foundations/build/scss/color-placeholder" as *;
+@use "@db-ui/foundations/build/scss/helpers/shadow-dom" as *;
+
+.elm-radio {
+	// Using another CSS variable for the size to differentiate for the small variant later
+	--radio-size: #{$db-sizing-sm};
```
nmerget, line 9: In the designs we use font-size for the radio, why do you use this variable?

With https://github.com/db-ui/mono/blob/main/packages/foundations/scss/_body-styles.scss#L7 you can use `--db-base-font-size` everytime you need the current font-size.

If we want to use another font-size e.g. in data-variant=small we would @extend the %db-body-sm placeholder.

packages/components/src/components/radio/radio.scss
```diff
@@ -0,0 +1,114 @@
...
+	&:required {
+		& + label {
+			&::after {
+				content: "*";
+			}
+		}
+	}
+
+	// Sizes
+	&%size-Small {
```
nmerget, line 91: `@extend %db-body-sm;` <- use this to overwrite font-size & line-height

packages/components/src/components/radio/radio.scss
```diff
@@ -0,0 +1,114 @@
...
+@use "@db-ui/foundations/build/scss/variables.global" as *;
+@use "@db-ui/foundations/build/scss/helpers/functions" as *;
+@use "@db-ui/foundations/build/scss/color-placeholder" as *;
+@use "@db-ui/foundations/build/scss/helpers/shadow-dom" as *;
+
+.elm-radio {
+	// Using another CSS variable for the size to differentiate for the small variant later
+	--radio-size: #{$db-sizing-sm};
+	
+	@extend %db-bg-neutral-0-ia;
```
nmerget, line 11: The radio button is adaptive as well, so we cannot set the color like this.

Maybe we can use https://github.com/db-ui/mono/blob/main/packages/components/src/components/button/button.scss#L18?

packages/components/src/components/radio/radio.scss
```diff
@@ -0,0 +1,114 @@
...
+
+	appearance: none;
+
+	aspect-ratio: 1;
+
+	border: max(calc(var(--radio-size, #{$db-sizing-sm}) / 16 + 0.5px), 2px)
```
nmerget, line 17: Okay, why do we calculate it like this? For the designs, its just 2px for medium and 1px for small isn't it?

packages/components/src/components/radio/radio.scss
```diff
@@ -0,0 +1,114 @@
...
+		solid currentColor;
+	border-radius: to-em($pxValue: 16);
```
nmerget, line 19: We should ask UX here: They should provide us the tokens or 3 default borders, somehow it differs in each component which could be a hint that we should use tokens.

packages/components/src/components/radio/radio.scss
```diff
@@ -0,0 +1,114 @@
...
+
+	// * the basic, unchecked style
+	display: inline-block;
+
+	height: var(--radio-size);
+
+	transition: border-width 0.5s ease;
```
nmerget, line 26: We have transitions inside foundations, we should bundle them at one place, as variables/placeolder/mixin so we can easily adapt every transition to have the same look and feel for every component

packages/components/src/components/radio/radio.scss
```diff
@@ -0,0 +1,114 @@
...
+				--db-current-color: #{$db-colors-neutral-on-bg-pressed};
+				color: $db-colors-neutral-on-bg-pressed;
+			}
+		}
+	}
+	// * the invalid style using the :invalid pseudo class (and [aria-invalid="true"] equivalent, see #136 and #141)
+	// TODO: We need to set the correct styling / color here after alignment
+	&:invalid,
+	&[aria-invalid="true"] {
+		border-color: #c13e34;
```
nmerget, line 56: Even though there is no design for it at the moment, we should use the same way we did it with inputs:
https://github.com/db-ui/mono/blob/main/packages/components/src/components/input/input.scss#L306

And therefore we should combine them into a placeholder to reuse it across all form elements

packages/components/src/components/radio/radio.scss
```diff
@@ -0,0 +1,114 @@
...
+		}
+	}
+
+	&[data-size="small"] {
+		@extend %size-Small;
+	}
+
+	// Adopted by https://www.heise.de/developer/artikel/a11y-Reduced-Motion-4356171.html
+	@media (prefers-reduced-motion: reduce) {
+		transition-duration: 0.01s !important;
```
nmerget, line 112: hmm `!important` do we need it here?

packages/components/src/components/radio/model.ts
```diff
@@ -0,0 +1,43 @@
...
+}
+
+export type DBRadioProps = DBRadioDefaultProps &
+	GlobalProps &
+	GlobalTextProps &
+	ChangeEventProps &
+	FocusEventProps &
+	ValidEventProps;
+
+export type DBRadioDefaultState = {
```
nmerget, line 32: Nearly all of those props are the same in input.

Please add a new `FormDefaultState` to `shared/model.ts` and include it in radio and input to reuse the same.

packages/components/src/components/radio/radio.scss
mfranzke: yes, because we want to ensure that transition is definitely being shortened to an unnoticeable timeframe.

packages/components/src/components/radio/radio.lite.tsx
```diff
@@ -95,13 +95,14 @@ export default function DBRadio(props: DBRadioProps) {
...
 				class={
-					'elm-radio' + (props.className ? ' ' + props.className : '')
+					'db-radio' + (props.className ? ' ' + props.className : '')
 				}
 				id={state.mId}
 				name={props.name}
 				checked={state._checked}
 				disabled={props.disabled}
 				value={state._value}
+				aria-labelledby={state.mId + '-label'}
```
mfranzke: This is not necessary on radio types, but only textual ones.

mfranzke:
@nmerget somehow the styles by the `.example-list` tend to let the radio elements increase in width:
https://db-ui.github.io/mono/review/361-featcomponent-radio/components/radio/examples

image

nmerget:
> @nmerget somehow the styles by the `.example-list` tend to let the radio elements increase in width: https://db-ui.github.io/mono/review/361-featcomponent-radio/components/radio/examples
> 
> image

This was just a quick overview for all combinations. Maybe a https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl?retiredLocale=de would be better here to have a default behaviour

mfranzke:
> > @nmerget somehow the styles by the `.example-list` tend to let the radio elements increase in width: https://db-ui.github.io/mono/review/361-featcomponent-radio/components/radio/examples
> > image
> 
> This was just a quick overview for all combinations. Maybe a https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl?retiredLocale=de would be better here to have a default behaviour

fixed by https://github.com/db-ui/mono/pull/610

mfranzke:
on hold until we fix the `ref` topic.

packages/components/src/components/radio/radio.scss
```diff
@@ -0,0 +1,97 @@
...
+@use "@db-ui/foundations/build/scss/helpers/shadow-dom" as *;
+@use "@db-ui/foundations/build/scss/helpers/component";
+@use "@db-ui/foundations/build/scss/color/color-variants" as *;
+
+$font-size-height: calc(var(--db-base-font-size) * var(--db-base-line-height));
+
+.db-radio {
+	@extend %default-color;
+	@extend %bg-transparent-interactive;
+
```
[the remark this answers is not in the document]
mfranzke: I've added it to the list within https://github.com/db-ui/mono/issues/498

annsch (approved):

annsch (approved):

PR #535, opened 2023-02-24
feat: providing divider SCSS function

mfranzke:
To enable borders with a transparent color related to the current `color` used on that element we need to provide a workaround previous to Relative Color Syntax being available in the relevant browsers.


packages/foundations/scss/helpers/_functions.scss
```diff
@@ -35,3 +35,38 @@
...
+		background-color: currentColor;
+		opacity: 0.16;
+
+		position: absolute;
+		@if $position == "top" {
+			height: 1px;
+			top: 0;
+			right: 0;
+			left: 0;
+		} @else if $direction == "bottom" {
```
nmerget, line 55: Why do you use `$direction` here?

packages/foundations/scss/helpers/_functions.scss
```diff
@@ -35,3 +35,38 @@
...
+			height: 1px;
+			bottom: 0;
+			right: 0;
+			left: 0;
+		} @else if $direction == "left" {
+			width: 1px;
+			bottom: 0;
+			top: 0;
+			left: 0;
+		} @else if $direction == "left" {
```
nmerget, line 65: Should this be `right`? :P

nmerget:
Can you add this to the DBDivider Component as well? So we have the same behavior an opacity etc.

packages/foundations/scss/helpers/_functions.scss
```diff
@@ -35,3 +35,38 @@
...
+// https://codepen.io/bramus/pen/VwMZOQK
+// https://bugs.chromium.org/p/chromium/issues/detail?id=1274133
+// https://github.com/WebKit/WebKit/commit/1890bfb551da24735804cb1a3d5f14b210f7c7b8
+@mixin divider($position: "top", $slot: "before") {
+	&::#{$slot} {
+		content: "";
+		background-color: currentColor;
+		opacity: 0.16;
+
+		position: absolute;
```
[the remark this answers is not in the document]
mfranzke: I've added that one for resilience of the workaround into the mixin as well.

packages/foundations/scss/helpers/_functions.scss
mfranzke: You're my Maximilian in code reviews.

mfranzke:
> Can you add this to the DBDivider Component as well? So we have the same behavior an opacity etc.

Divider behaves totally differently (regarding how the SCSS mixin is setup to fit those needs for borders).

nmerget (approved):

PR #538, opened 2023-02-25
refactor(style): both simplifying and extending inherit declarations

mfranzke:
To inherit font related declarations in total, we could only use the shorthand `font` to inherit from. Additionally we most likely even also want to inherit `letter-` and `word-spacing`.

Adapted from https://adrianroselli.com/2021/03/under-engineered-select-menus.html


mfranzke:
most likley we only want `font-size` and `line-height` inherited, but not e.g. `font-weight` from the outside.

PR #549, opened 2023-02-27
refactor: focus outline radius

mfranzke:
The border radius is supposed to be 0.25rem


nmerget:
I guess we should add it for the button as well. Maybe we can create a placeholder for this to equalize it? More over should we use rem or px for the radius?

mfranzke:
> I guess we should add it for the button as well. Maybe we can create a placeholder for this to equalize it? More over should we use rem or px for the radius?

It gets added to the button already by the following selector:
https://github.com/db-ui/mono/pull/549/files#diff-3e45cc333da0df78a6653a7bd973570c0c815dd068b8b6c1a7394dee0655a479R84

The additional declaration for input elements is only necessary because of the overwrite of the elements construct and that for the selector in this component.

nmerget (dismissed):

nmerget (approved):

PR #569, opened 2023-03-02
feat: add refs for custom event workarounds

nmerget:
Add forwardRef (react) and refs for angular and vue to get access to the components


packages/components/src/components/button/button.lite.tsx
```diff
@@ -40,6 +40,8 @@ useMetadata({
 });
 
 export default function DBButton(props: DBButtonProps) {
+	// This is used as forwardRef
+	let component: any;
```
annsch, line 44: Shouldn't we avoid using any and provide a more specific type? When we really don't know which type this will be, we should use _unknown_ instead of _any_ in typescript

packages/components/src/components/alert/alert.lite.tsx
```diff
@@ -24,6 +24,8 @@ const DEFAULT_VALUES = {
 };
 
 export default function DBAlert(props: DBAlertProps) {
+	// This is used as forwardRef
+	let component: any;
```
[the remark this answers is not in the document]
nmerget: I would provide it for every component because we would lose users if there is no way to add a custom event to the component. Let's try it like this and check if we can remove the it in the future

packages/components/src/components/button/button.lite.tsx
nmerget: For react I think it's better to provide any, because we can parse down a jsx component which is a function

annsch (dismissed):

annsch (approved):

PR #570, opened 2023-03-02
feat: add default component for vue and angular

nmerget:
- [x ] Add default component for vue
- [ ] Add default component for angular

### DIFF
--- a/showcases/shared/default-component-data.ts
+++ b/showcases/shared/default-component-data.ts
@@ -1,0 +1,17 @@
+export type DefaultComponentExample = {
+	name?: string;
+	example?: any;
+	style?: any;
+	className?: string;
+};
+
+export type DefaultComponentVariants = {
+	name: string;
+	examples: DefaultComponentExample[];
+};
+
+export type DefaultComponentProps = {
+	title: string;
+	description?: any;
+	variants: DefaultComponentVariants[];
+};
--- a/showcases/vue-showcase/src/components/DefaultComponent.vue
+++ b/showcases/vue-showcase/src/components/DefaultComponent.vue
@@ -1,0 +1,61 @@
+<script setup lang="ts">
+import { DBDivider } from "../../../../output/vue/vue3/src";
+import type {
+	DefaultComponentExample,
+	DefaultComponentProps,
+	DefaultComponentVariants
+} from "../../../shared/default-component-data";
+
+interface DefaultExample extends DefaultComponentExample {
+	name?: string;
+	example?: any;
+	style?: any;
+	className?: string;
+}
+interface DefaultVariants extends DefaultComponentVariants {
+	name: string;
+	examples: DefaultExample[];
+}
+/* Workaround see: https://vuejs.org/guide/typescript/composition-api.html#typing-component-props */
+interface DefaultProps extends DefaultComponentProps {
+	title: string;
+	description?: any;
+	variants: DefaultVariants[];
+}
+
+const props = defineProps<DefaultProps>();
+
+const getSlotName = (
+	title: string,
+	variantIndex: number,
+	exampleIndex: number
+) => {
+	return `${title.toLowerCase()}-${variantIndex}-${exampleIndex}`;
+};
+</script>
+
+<template>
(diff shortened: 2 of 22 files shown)


.xo-config.json
```diff
@@ -13,6 +13,7 @@
 	"rules": {
 		"@db-ui/v2-component-used": "warn",
 		"no-console": ["error", { "allow": ["warn", "error"] }],
+		"import/no-cycle": 0,
```
mfranzke, line 16: The [eslint rules description](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-cycle.md) sounds like that it makes sense in general – what's the reason to deactivate it?

.github/workflows/01-lint.yml
```diff
@@ -15,7 +15,11 @@ jobs:
         uses: ./.github/actions/npm-cache
 
       - name: ⚡ Run Lint
-        run: npm run lint
+        run: |
+          npm run lint:jscpd
+          npm run lint:markdownlint
+          npm run lint:stylelint
+          npm run lint:xo
```
mfranzke, line 22: What's the reason for this change (especially as those aren't run in parallel anymore, are they)?

.xo-config.json
nmerget: The issue is that I moved all .ts files for configuring the showcases into the root showcase folder.
The linting in cicd failed always because it takes to long, see:
https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-cycle.md#when-not-to-use-it

package.json
```diff
@@ -46,7 +46,7 @@
     "lint:jscpd": "jscpd --exitCode 1 .",
     "lint:markdownlint": "markdownlint -c .markdown-lint.yml **/*.md",
     "lint:stylelint": "stylelint **/*.scss",
-    "lint:xo": "xo",
+    "lint:xo": "cross-env TIMING=1 xo",
```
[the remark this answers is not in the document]
nmerget: At the moment `xo` does not provide any information, besides of errors and warning.

To resolve https://github.com/db-ui/mono/pull/570#discussion_r1129096677 I added the Timing to get additional information in cicd.

.github/workflows/01-lint.yml
nmerget: I get a js heap issue for linting, I thought this might be the solution... But there is still the heap issue

.xo-config.json
nmerget: fixed in https://github.com/db-ui/mono/pull/638

package.json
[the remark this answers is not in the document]
nmerget: fixed in https://github.com/db-ui/mono/pull/638

.github/workflows/01-lint.yml
nmerget: fixed in https://github.com/db-ui/mono/pull/638

showcases/shared/default-component-data.ts
```diff
@@ -1,9 +1,7 @@
-import { type CSSProperties, type ReactElement } from 'react';
-
 export type DefaultComponentExample = {
-	name: string;
-	example: ReactElement;
-	style?: CSSProperties;
+	name?: string;
+	example?: any;
+	style?: any;
```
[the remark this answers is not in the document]
mfranzke: 👮‍♀️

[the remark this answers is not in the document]
nmerget: fixed

showcases/vue-showcase/src/components/DefaultComponent.vue
```diff
@@ -0,0 +1,132 @@
...
+} from "../../../../packages/components/src/shared/constants";
+import { useRoute } from "vue-router";
+import { Ref, ref } from "vue";
+
+interface DefaultExample extends DefaultComponentExample {
+	name?: string;
+	example?: any;
+	style?: any;
+	className?: string;
+	code?: any;
```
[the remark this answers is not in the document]
nmerget: fixed

PR #585, opened 2023-03-06
feat: add iframe component parser to include it in mapo

nmerget:
UX wished for a way to include an iframe with a headline, text, and 3 tags inside mapo.

With this code we can add additional components in the future.


mfranzke:
@nmerget there's a code scanning error reported and some problems within the pipeline.

showcases/patternhub/components/component-parser/data.ts
```diff
@@ -0,0 +1,12 @@
+export type ComponentParserType = {
+	componentsString: string;
+};
+
+export type ComponentType = {
+	index?: string | number;
+	type?: string;
+	content?: string | ComponentType[];
+	props?: any;
```
mfranzke, line 9: @dkolba, what's your opinion ?

mfranzke (approved):

PR #607, opened 2023-03-10
fix: issues with power apps (icon, onClick/Change)

nmerget:


packages/components/_templates/power-apps/new/comp-xml.ejs.t
```diff
@@ -11,18 +11,61 @@ force: true
...
+				  <% } -%>
+
+				  <% if(typeof prop.type !== 'undefined' && prop.type !== "Icon" && prop.type !== "DefaultVariant"){ -%>
+                  of-type="<%= prop.type %>"
+                  <% } else if(prop.type === "Icon"){ -%>
+                  of-type="Enum">
+                  <value name="None" display-name-key="None">_</value>
+                  <value name="Account" display-name-key="Account">account</value>
+                  <value name="Edit" display-name-key="Edit">edit</value>
+                  <!-- TODO: Add other icons here -->
```
mfranzke, line 35: we could make this a freetext field as well, as there might be icons in the hundrets and we would need to maintain this list in here.

nmerget: Hmm, but for a default user who doesn't know about icons it would be better to have a dropdown. 
Also we should show the icons in the dropdown somehow...

nmerget: image

This would be perfect for a user I think

mfranzke (dismissed):

mfranzke (approved):

PR #609, opened 2023-03-10
feat: playwright component all variants

dkolba:


e2e/Dockerfile
```diff
@@ -9,7 +9,6 @@ COPY packages/ packages/
 # Install dependencies
 RUN npm install
 # Install browsers
```
mfranzke, line 11:
```suggestion
```

If we remove the command, we most likely should remove the comment as well.

nmerget (approved):

PR #610, opened 2023-03-10
fix: issue with patternhub example list

nmerget:


showcases/patternhub/styles/globals.scss
```diff
@@ -23,16 +23,16 @@ iframe {
 }
 
 .example-list {
-	display: flex;
-	flex-direction: column;
-	width: fit-content;
-	padding: 16px;
```
mfranzke, line 29: Why do you remove those two as well ?

PR #611, opened 2023-03-10
chore: update component parser to use all components

nmerget:
- Add remaining components to component parser
- Add tailwind to get access to classes with db spacings


showcases/patternhub/pages/iframe/index.tsx
```diff
@@ -15,7 +15,7 @@ const IframeComponent = () => {
 
 	return (
 		<div
-			className={`iframe-component-container db-ui-${tonality} db-bg-${color}`}>
+			className={`flex flex-col p-fix-md db-ui-${tonality} db-bg-${color}`}>
```
mfranzke, line 18: why is it necessary to switch to tailwind?

showcases/patternhub/package.json
```diff
@@ -30,10 +31,13 @@
     "@types/node": "18.15.0",
     "@types/react": "18.0.28",
     "@types/react-dom": "18.0.11",
+    "autoprefixer": "^10.4.14",
     "eslint": "8.35.0",
     "eslint-config-next": "13.2.4",
     "open-cli": "^7.1.0",
+    "postcss": "^8.4.21",
     "sass": "^1.58.3",
+    "tailwindcss": "^3.2.7",
```
mfranzke, line 40: I don't see the need to integrate tailwind.

nmerget: We can use it to expose our variables as classes. UX can use e. g. `p-fix-md` which is the same as `padding: $db-spacing-fixed-md`. They can basically style everything with that in the iFrame

showcases/patternhub/pages/iframe/index.tsx
nmerget: It's not necessary for this part, it's for the iframe. But why not use it here ;)

mfranzke: As discussed, we won't need it and therefor shouldn't integrate it.

nmerget: I disagree with this, I removed it for now. Let's see what UX will say and what they want to use. If they want to set custom spacings, the tailwind approach would be easy...

showcases/patternhub/package.json
```diff
@@ -14,7 +14,8 @@
     "lint": "next lint",
     "next:dev": "next dev",
     "next:start": "next start",
-    "start": "npm-run-all copy next:start"
+    "start": "npm-run-all copy next:start",
+    "tailwind:generate": "tailwind -i ./styles/tailwind.css -o ./styles/tailwind.generated.css"
```
mfranzke, line 18: do we still need this ?

mfranzke (approved):

PR #624, opened 2023-03-13
fix: issue with forwardRef

nmerget:


packages/components/src/components/input/input.lite.tsx
```diff
@@ -22,7 +16,6 @@ useMetadata({
 export default function DBInput(props: DBInputProps) {
 	// This is used as forwardRef
 	let component: any;
-	const formRef = useRef<HTMLInputElement>(null);
```
mfranzke, line 25: Do we need this change within `packages/components/src/components/tab/tab.lite.tsx` as well ?

nmerget: Good question, I would keep it for now, because `tab` was just a test component and we have to rework it anyways.

mfranzke (approved):

PR #626, opened 2023-03-13
fix: e2e tests

mfranzke:
fixed heavily supported by @dkolba.

- we don't need `RUN npx playwright install`
- we shouldn't depend on a `build` previous to running the `e2e` process, but couldn't run `RUN npm run build` with confidence within _docker_ at the moment.


nmerget:
Can you update the `README.md` inside `e2e` directory? We need to run `npm run build` before right?

mfranzke:
@nmerget done that.

PR #638, opened 2023-03-14
fix: issue with xo and add addtional comments to exceptions

nmerget:
- Changed rules for `xo` or add a comment for exceptions
- Increase js-heap for `xo `
- Add timetable output for `xo` for better debugging


mfranzke:
why do you switch from single to double quotes ?

.prettierrc.json
```diff
@@ -1,5 +1,7 @@
 {
-	"trailingComma": "none",
+	"trailingComma": "es5",
+	"singleQuote": false,
```
mfranzke, line 3: why do you switch from single to double quotes ?

nmerget: I thought we have to change it, based on this comment:

```
To stick with Prettier's defaults, add this to your Prettier config:

module.exports = {
	trailingComma: 'es5',
	singleQuote: false,
	bracketSpacing: true,
};
If contradicting options are set for both Prettier and XO an error will be thrown.
```

But I reverted it

mfranzke (approved):

PR #639, opened 2023-03-14
fix: input label opacity for nested tonality

annsch:
-[x] fixes input label styles in nested tonalities


packages/components/src/components/input/input.scss
```diff
@@ -11,6 +11,19 @@
...
+
+.db-ui-functional {
+	--db-input-disabled-label-opacity: 0;
+}
+
+.db-ui-regular,
+.db-ui-expressive {
+	--db-input-disabled-label-opacity: 0.25;
+}
+
```
mfranzke, line 26: Is this only about inputs, and not about other form element types as well?

annsch: @mfranzke The input component has different behaviour depending on current tonality. Hopefully this is only relevant for input component.

packages/components/src/components/input/input.scss
```diff
@@ -11,6 +11,19 @@
...
+:root {
+	--db-input-disabled-label-opacity: 0.25;
+}
+
+.db-ui-functional {
+	--db-input-disabled-label-opacity: 0;
+}
+
+.db-ui-regular,
+.db-ui-expressive {
```
mfranzke, line 23: Is it technically possible to merge this with the selector at line 14 ?

packages/components/src/components/input/input.scss
```diff
@@ -11,6 +11,19 @@
 @use "@db-ui/foundations/build/scss/color/color-variants" as *;
 @use "@db-ui/foundations/build/scss/helpers/component";
 
+:root {
```
mfranzke, line 14:
```suggestion
:root,
.db-ui-regular,
.db-ui-expressive {
```

packages/components/src/components/input/input.scss
```diff
@@ -11,6 +11,19 @@
...
+}
+
+.db-ui-functional {
+	--db-input-disabled-label-opacity: 0;
+}
+
+.db-ui-regular,
+.db-ui-expressive {
+	--db-input-disabled-label-opacity: 0.25;
+}
```
mfranzke, line 25:
```suggestion
```

packages/components/src/components/input/input.scss
```diff
@@ -401,19 +418,6 @@
 	}
 }
 
-// disable transform animation for functional input
-.db-ui-functional .db-input > label {
-	&:focus,
-	&:not(:placeholder-shown) {
```
mfranzke, line 407: this didn't work before, because it was nested within label and not an `input`, right? And we don't need it anymore?

mfranzke (approved):

packages/components/src/components/input/input.scss
annsch: @mfranzke this did work before (actually I am wondering why :D) but I had to remove this explicit dependency of .db-ui-functional. This overwrites transform animation for nested tonalities, e.g.:
``````
<section class="db-ui-functional">
> <section class="db-ui-expressive"> 
> <input class="db-input"> <-- has no animation for label 
> </section>
</section>
``````

annsch: I'd like to discuss the problem I tried to solve with this with you separately. Currently I do not have a solution :( 
My goal is to remove transitions only for inputs in .db-ui-functional

mfranzke: I've created a ticket for this, let's keep the wheels turning for this PR: https://github.com/db-ui/mono/issues/670

PR #651, opened 2023-03-15
refactor: aligned focus outlines

### LINKED ISSUE #650: Aligned focus outlines
The focus outline needs to be 6px according to the guidelines.

- [x] It's not supposed to get assigned in rem, but in px, as it shouldn't increase in size when only text-size gets increased.
- [x] It's supposed to get set on `focus-visible`, and not `focus` already.
- [x] The `card` component has a special value of 11px.

mfranzke:
~Resolves #650~

The scope of this changed dramatically: We shouldn't set a `border-radius` for `:focus-visible` state, but still let the `border-radius` defined for the elements as is, except for links that don't have a standard visible border at all.


nmerget:
I changed it to 4px in the past, because the outline is not working the same way as in sketch/zeplin. 

If we add 6px the button, for example, gets a weird-looking effect, because we have the transparent border-radius... In zeplin they use a rect with a rounded border with 6px so this is a completely different approach. Let's make a meeting with us to discuss this.

packages/components/src/components/card/card.scss
```diff
@@ -14,6 +14,8 @@
 	);
 
 	border-radius: 8px;
+	// Different border-radius on focus for this component
+	--db-focus-outline-radius: 11px; // stylelint-disable-line order/order
```
mfranzke, line 18: remove this.

packages/foundations/scss/helpers/_component.scss
```diff
@@ -29,8 +28,8 @@
 	@extend %default-color;
 	@extend %default-border-radius;
 
-	&:focus {
-		border-radius: var(--db-focus-outline-radius, #{to-rem(6)});
+	&:focus-visible {
+		border-radius: var(--db-focus-outline-radius, 6px);
```
mfranzke, line 32: we most likely don't want to overwrite this on interactive elements.

packages/components/src/components/link/link.scss
```diff
@@ -56,6 +56,10 @@
 		}
 	}
 
+	&:focus-visible {
+		border-radius: 4px;
```
nmerget, line 60: Did you ask UX to generate those border-radius? We need 4px and 8px (for cards).

Moreover, should we use the placeholder here? Was it your intention to use `px` instead of `rem`?

mfranzke: > Did you ask UX to generate those border-radius? We need 4px and 8px (for cards).
not so far. It would probably still take some time to integrate those.

mfranzke: > Was it your intention to use `px` instead of `rem`?

I haven't had a chance for another research and a final decision, but we most likely shouldn't use `rem` for anything else than `font-size` properties, as only those should get resized depending on font resizes, but not any further properties.

nmerget: Okay so you change it here `to-rem` ?

mfranzke: > Did you ask UX to generate those border-radius? We need 4px and 8px (for cards).

not so far. It would probably still take some time to integrate those.
Update: I've created a ticket: DBSWC-977

nmerget (approved):

PR #667, opened 2023-03-17
chore: slight optimizations after radio merge

mfranzke:
- added missing description to form components page
- formatted radios `README.md` file
- removed unused imports
- spacing in between label and radio needs to be tonality related (based on a CSS token instead of a fixed variable value, by switching to the non `-regular` including SCSS variable)
- the dot shouldn't change `background-color` on `:hover`


packages/components/src/components/radio/README.md
```diff
@@ -20,10 +20,7 @@ TODO
 To get DBRadio work with `v-model` you have to use v-model argument syntax:
 
 ```ts
-<DBRadio
-	label="Textlabel"
-	v-model:value="vModelTest"
-></DBRadio>
+<DBRadio label="Textlabel" v-model:value="vModelTest"></DBRadio>
```
nmerget, line 23: I guess this should be `v-model:checked` shouldn't it?

packages/components/src/components/radio/radio.scss
```diff
@@ -55,7 +55,7 @@ $font-size-height: calc(var(--db-base-font-size) * var(--db-base-line-height));
 	& + label {
 		@extend %shadow-dom-extend;
 
-		margin-inline-start: $db-spacing-fixed-regular-xs;
+		margin-inline-start: var(--db-spacing-fixed-xs);
```
nmerget, line 58: Just use `$db-spacing-fixed-xs` from variables.global, it's the same and should be used all the time

packages/components/src/components/radio/README.md
mfranzke: this was just a reformatting, I didn't question the declaration itself.

nmerget (approved):

PR #672, opened 2023-03-17
feat: add drawer

nmerget:
Resolves https://github.com/db-ui/mono/issues/695


packages/components/src/components/drawer/drawer.lite.tsx
```diff
@@ -0,0 +1,106 @@
...
+				ref={dialogContainerRef}
+				class={
+					'db-drawer-container' +
+					(props.className ? ' ' + props.className : '')
+				}
+				data-size={props.size}
+				data-width={props.width}
+				data-direction={props.direction}
+				data-rounded={props.rounded}>
+				<div class="db-drawer-header">
```
mfranzke, line 88:
```suggestion
				<header>
```

nmerget: hmm but then this change should also be made in scss right?

nmerget: And we have to put it into something like:

The <header> element defines a [banner](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/banner_role) landmark when its context is the [<body>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/body) element. The HTML header element is not considered a banner landmark when it is descendant of an [<article>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/article), [<aside>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/aside), [<main>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/main), [<nav>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/nav), or [<section>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/section) element.

mfranzke: yeah, I was lazy about the code suggestion parts.

nmerget: add the header inside of an `article` I think that would fit the most.
I kept the className because there can be multiple drawers in an app, so we provide a easy way for an overwrite

packages/components/scripts/post-build/components.js
```diff
@@ -19,6 +19,24 @@
  * }]}
  */
 const getComponents = () => [
+	{
+		name: 'drawer',
+		defaultStylePath: 'components/drawer/drawer-web-component.css',
```
mfranzke, line 24: why is `defaultStylePath` necessary on this (and `code-docs` components), but not the others? Because these are "internal" components? And why do you reference the `-web-component` here instead of the regular as in `code-docs.css`?

packages/components/src/components/drawer/README.md
mfranzke, line 1: The `drawer-header` slot isn't mentioned in this file.

packages/components/src/components/drawer/drawer.lite.tsx
```diff
@@ -0,0 +1,106 @@
...
+					(props.className ? ' ' + props.className : '')
+				}
+				data-size={props.size}
+				data-width={props.width}
+				data-direction={props.direction}
+				data-rounded={props.rounded}>
+				<header class="db-drawer-header">
+					<div>
+						<Slot name="drawer-header" />
+					</div>
```
mfranzke, line 91:
```suggestion
					<Slot name="drawer-header" />
```

What for do we need this `<div>`?

packages/components/src/components/drawer/drawer.scss
```diff
@@ -0,0 +1,171 @@
...
+
+		&[data-direction="down"] {
+			@extend %direction-down;
+		}
+
+		.db-drawer-header {
+			@include divider("bottom");
+			display: flex;
+			gap: $db-spacing-fixed-lg;
+			justify-content: space-between;
```
mfranzke, line 114:
```suggestion
```

Currently this element has only one direct child (`div`) so I don't understand why we need these declarations.

packages/components/src/components/drawer/drawer.lite.tsx
```diff
@@ -0,0 +1,106 @@
...
+			) {
+				if (props.onClose) {
+					props.onClose();
+				}
+			}
+		},
+		handleDialogOpen: () => {
+			if (dialogRef) {
+				if (props.open && !dialogRef.open) {
+					dialogContainerRef?.classList.remove('hide');
```
mfranzke, line 38:
```suggestion
					dialogContainerRef?.classList.hidden = false;
```

packages/components/src/components/drawer/drawer.lite.tsx
```diff
@@ -0,0 +1,106 @@
...
+					if (props.noBackdrop) {
+						dialogRef.show();
+					} else {
+						dialogRef.showModal();
+					}
+				}
+				if (!props.open && dialogRef.open) {
+					dialogContainerRef?.classList.add('hide');
```
mfranzke, line 46:
```suggestion
					dialogContainerRef.hidden = true;
```

packages/components/src/components/drawer/drawer.lite.tsx
```diff
@@ -0,0 +1,106 @@
...
+					setTimeout(() => {
+						dialogContainerRef?.classList.remove('hide');
```
mfranzke, line 48:
```suggestion
						dialogContainerRef?.classList.hidden = false;
```

packages/components/src/components/drawer/drawer.scss
```diff
@@ -0,0 +1,171 @@
...
+
+			&[data-direction="up"] {
+				animation: show-bottom-to-top $animation-show-timing;
+			}
+
+			&[data-direction="down"] {
+				animation: show-top-to-bottom $animation-show-timing;
+			}
+
+			&.hide {
```
mfranzke, line 151:
```suggestion
			&[hidden] {
```

packages/components/src/components/drawer/drawer-web-component.scss
```diff
@@ -0,0 +1,11 @@
+@use "@db-ui/foundations/build/scss/helpers/functions" as *;
+@forward "../../styles/dialog-init";
+@forward "drawer";
+
+.db-drawer {
+	.db-drawer-container {
+		.db-drawer-header {
```
mfranzke, line 7:
```suggestion
	header {
```

packages/components/src/components/drawer/drawer-web-component.scss
```diff
@@ -0,0 +1,11 @@
...
+			height: to-rem(32);
+		}
```
mfranzke, line 9:
```suggestion
```

packages/components/src/components/drawer/drawer.lite.tsx
```diff
@@ -0,0 +1,106 @@
...
+				ref={dialogContainerRef}
+				class={
+					'db-drawer-container' +
+					(props.className ? ' ' + props.className : '')
+				}
+				data-size={props.size}
+				data-width={props.width}
+				data-direction={props.direction}
+				data-rounded={props.rounded}>
+				<header class="db-drawer-header">
```
mfranzke, line 88:
```suggestion
				<header>
```

packages/components/src/components/drawer/drawer.lite.tsx
```diff
@@ -0,0 +1,106 @@
...
+				state.handleClose(event);
+			}}
+			data-backdrop={!props.noBackdrop}>
+			<Show when={state.stylePath}>
+				<link rel="stylesheet" href={state.stylePath} />
+			</Show>
+			<article
+				ref={dialogContainerRef}
+				class={
+					'db-drawer-container' +
```
mfranzke, line 81:
```suggestion
					'container' +
```

packages/components/src/components/drawer/README.md
nmerget: https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_templates_and_slots#adding_flexibility_with_slots

It is a default named slot so you just add `slot="drawer-header"` to your tag.

packages/components/src/components/drawer/drawer.lite.tsx
nmerget: I need to use it for the animation: https://github.com/db-ui/mono/blob/feat-drawer/packages/components/src/components/drawer/drawer.scss#L151

I don't think I can use `hidden` as a selector is CSS do I? Maybe I can rename the class if you like

packages/components/src/components/drawer/drawer.lite.tsx
```diff
@@ -0,0 +1,106 @@
...
+				data-width={props.width}
+				data-direction={props.direction}
+				data-rounded={props.rounded}>
+				<header class="db-drawer-header">
+					<div>
+						<Slot name="drawer-header" />
+					</div>
+					<Show when={props.withCloseButton}>
+						<DBButton
+							id="button-close-drawer"
```
mfranzke, line 94: `id` doesn't seem to be a prop in the `db-button` component.

packages/components/src/components/drawer/drawer.lite.tsx
```diff
@@ -0,0 +1,106 @@
...
+							icon="close"
+							variant="transparent"
+							onClick={() => state.handleClose('close')}>
+							Close Drawer
+						</DBButton>
+					</Show>
+				</header>
+				<div class="db-drawer-content">{props.children}</div>
```
mfranzke, line 102:
```suggestion
				<div class="content">{props.children}</div>
```

packages/components/src/components/drawer/drawer.scss
```diff
@@ -0,0 +1,171 @@
...
+			.db-drawer-content {
+				padding-inline: $db-spacing-fixed-md;
+			}
+		}
+
+		&[data-rounded="true"] {
+			box-shadow: $db-elevation-4;
+		}
+
+		&[data-direction="left"] {
```
mfranzke, line 98: I'm struggeling with non-semantic positional values – but even also `start` and `end` don't seem to perfectly fit here. What about `forth` and `back`?

mfranzke, line 98:
```suggestion
		&[data-direction="start"] {
```

I'm struggeling with non-semantic positional values – but even also `start` and `end` don't seem to perfectly fit here. What about `forth` and `back`?

packages/components/scripts/post-build/components.js
nmerget: Was a leftover, we don't need it anymore

packages/components/src/components/drawer/drawer.scss
```diff
@@ -0,0 +1,171 @@
...
+		&[data-direction="up"] {
+			@extend %direction-up;
+		}
+
+		&[data-direction="down"] {
+			@extend %direction-down;
+		}
+
+		.db-drawer-header {
+			@include divider("bottom");
```
mfranzke, line 111: Should there be a divider even if the header is empty?

packages/components/src/components/drawer/drawer.lite.tsx
nmerget: fixed

packages/components/src/components/drawer/drawer.scss
nmerget: fixed

packages/components/src/components/drawer/drawer.lite.tsx
nmerget: fixed

packages/components/src/components/drawer/drawer.lite.tsx
nmerget: fixed

packages/components/src/components/drawer/drawer.scss
nmerget: fixed

packages/components/src/components/drawer/drawer.scss
```diff
@@ -0,0 +1,171 @@
...
+			&:not(:has(> :nth-child(2))) {
+				display: none;
+			}
+		}
+
+		.db-drawer-content {
+			display: flex;
+			flex-direction: column;
+			height: 100%;
+			overflow: hidden;
```
mfranzke, line 130:
```suggestion
```

what for do we need to cut longer elements especially due to vertical amount of content in the drawer?

packages/components/src/components/drawer/drawer-web-component.scss
nmerget: removed it for now, we should look at this when we fix all components for vanilla/web-components after april

packages/components/src/components/drawer/drawer.scss
nmerget: I checked other design-systems: https://mui.com/material-ui/react-drawer/#swipeable

I think left and right, up and down makes sense and it's simple. We could name it `from-right-to-left` to make it bulletproof but it could be a little bit of an overkill

packages/components/src/components/drawer/drawer.scss
nmerget: If the header is empty it gets `display:none` so I think it doesn't matter right?

packages/components/src/components/drawer/drawer.scss
```diff
@@ -0,0 +1,171 @@
...
+			padding-inline: $db-spacing-fixed-lg;
+
+			&:not(:has(> :nth-child(2))) {
+				display: none;
+			}
+		}
+
+		.db-drawer-content {
+			display: flex;
+			flex-direction: column;
```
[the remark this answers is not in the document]
nmerget: You are right, I used it from the `refactor-header` branch where I need this styling but I should handle it there for the mobile navigation. Good catch

packages/components/src/components/drawer/drawer.scss
nmerget: https://github.com/db-ui/mono/pull/672#discussion_r1144341252

packages/components/src/components/drawer/drawer.lite.tsx
nmerget: fixed

packages/components/src/components/drawer/drawer.lite.tsx
nmerget: Added it for button, but we should add id for every component: https://github.com/db-ui/mono/issues/697

packages/components/src/components/drawer/drawer.scss
mfranzke: let's stay better than mui.

packages/components/src/components/drawer/drawer.lite.tsx
nmerget: Let' keep it for now, and discuss it for all components with the team

packages/components/src/components/drawer/drawer.lite.tsx
nmerget: Let' keep it for now, and discuss it for all components with the team

packages/components/src/components/drawer/drawer.lite.tsx
nmerget: Let' keep it for now, and discuss it for all components with the team

packages/components/src/components/drawer/drawer.scss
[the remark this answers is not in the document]
mfranzke: always at your service.

packages/components/src/components/drawer/drawer.lite.tsx
mfranzke: What for do we need it? And having `id` both on the wrapping tag as well as the inner (for the ones that wrap native HTML elements) would lead to two tags in the DOM with the same `id` (at least for the Angular and Web Components to my understanding).

packages/components/src/components/drawer/README.md
mfranzke: even in the other frameworks? Tried it for React and if I didn't make anything else wrong it didn't work.

packages/components/src/components/drawer/drawer.scss
mfranzke: Now I got it, there's the close-button still in there. But than on the other hand, would the header ever be empty as it at least contains the button?

nmerget: Yes we can hide the button initially with the prop `noCloseButton`. We would not show the header then; the user must handle the close event. Maybe as a primary button at the end of the children.

packages/components/src/components/drawer/README.md
nmerget: Nope React is special here: Mitosis compiles it to a prop  `slotDrawerHeader` and you pass your jsx into this. Everything is a function so React behaves different

packages/components/src/components/drawer/drawer.lite.tsx
nmerget: fixed

annsch (approved):

packages/components/src/components/drawer/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Angular
+
+TODO
+
+## React
+
+```tsx
+import { useState } from "react";
+import { DBButton, DBDrawer } from "../../../../../output/react/src";
```
dkolba, line 15: Examples should use the proper npm package names

dkolba (dismissed):

dkolba (approved):

PR #673, opened 2023-03-17
fix: issue with sass import warning

nmerget:


packages/components/package.json
```diff
@@ -14,7 +14,7 @@
     "build": "npm-run-all build:*",
     "build:mitosis": "mitosis build",
     "build:mv-styles": "cpr src build -o -f \"(.ts|.tsx|.md)$\"",
-    "build:sass": "sass build --no-source-map --load-path=node_modules/ --load-path=../../node_modules/ --future-deprecation=import",
+    "build:sass": "sass build --no-source-map --load-path=node_modules/ --load-path=../../node_modules/",
```
mfranzke, line 17:
```suggestion
    "build:sass": "sass build --no-source-map --load-path=node_modules/ --load-path=../../node_modules/ --future-deprecation=import",
```

why remove it? It might be a good reminder in case somebody still uses it.

packages/foundations/package.json
```diff
@@ -11,7 +11,7 @@
   "scripts": {
     "build": "npm-run-all build:*",
     "build:01_style-dictionary": "node style-dictionary.js && npm run copy:scss",
-    "build:02_css": "sass --no-source-map --load-path=node_modules/ --load-path=../../node_modules/ build/scss/:build/css/ --future-deprecation=import",
+    "build:02_css": "sass --no-source-map --load-path=node_modules/ --load-path=../../node_modules/ build/scss/:build/css/",
```
mfranzke, line 14:
```suggestion
    "build:02_css": "sass --no-source-map --load-path=node_modules/ --load-path=../../node_modules/ build/scss/:build/css/ --future-deprecation=import",
```

why remove it? It might be a good reminder in case somebody still uses it.

nmerget: https://github.com/db-ui/mono/pull/673#discussion_r1140391061

packages/components/package.json
mfranzke: but than it wouldn't hurt to leave that check in there, would it ?

PR #711, opened 2023-03-23
refactor: show-code for documentation

nmerget:


packages/components/src/components/alert/alert.scss
```diff
@@ -92,6 +92,10 @@ $variants: (
 
 	&[data-type="inline"] {
 		@extend %default-card;
+
+		&[data-elevation="none"] {
+			box-shadow: none;
+		}
```
mfranzke, line 98: does this mean that the `alert` type always needs to have an elevation and that this can't get overwritten?

nmerget: Only `inline` alerts look like the card (rounded borders and elevation). And therefore only the `inline` alert should have the option for `data-elevation="none"`.
The default alert has no elevation at all because it's not a "sibling" of card

PR #716, opened 2023-03-23
feat: split docs file and import them in patternhub

nmerget:


showcases/patternhub/pages/components/index.mdx
```diff
@@ -1,7 +1,6 @@
 import DefaultPage from "../../components/default-page";
+import Readme from "../../../../packages/components/README.md";
```
mfranzke, line 2: we need to ensure that we don't make any external requests e.g. to shields.io, as we don't have enough information and no contract about / regarding their data security.

mfranzke (approved):

showcases/patternhub/pages/components/index.mdx
nmerget: I guess this import will be compiled during the build process. Do you mean the content inside the components README.md?

mfranzke: doing a replacement by base64 or removal due to the build process would be perfectly fine from my point of view.

mfranzke: Created https://github.com/db-ui/mono/issues/723 for clarification.

PR #717, opened 2023-03-23
refactor: using jsx for vue

annsch:


mfranzke:
@annsch could you please leave a quick comment to document why we don't follow up with this idea?

annsch:
Using jsx for Vue would just be a workaround for our rendering process and would not reflect common use of native vue components.

PR #724, opened 2023-03-24
feat: combine showcases

nmerget:


packages/components/_templates/mitosis/new/shared-component-variants.ejs.t
```diff
@@ -1,17 +1,16 @@
 ---
-to: ../../showcases/shared/<%= name %>.ts
+to: ../../showcases/shared/<%= name %>.json
 ---
-import { DefaultComponentVariants } from './default-component-data';
-
-const defaultComponentVariants: DefaultComponentVariants[] = [
+[
 	{
 		name: 'TODO: Edit me in ./showcases/shared/<%= name %>.ts',
```
mfranzke, line 6:
```suggestion
		name: 'TODO: Edit me in ./showcases/shared/<%= name %>.json',
```

In Relation to the change on line 2, this would need to be `.json` as well, wouldn't it?

mfranzke (approved):

PR #746, opened 2023-03-28
docs: add bug report template

mfranzke:
Based on GitHubs proposal.

- [ ] Let's adapt further stuff from https://github.com/BuilderIO/mitosis/issues/new?assignees=&labels=bug%2Ctriage&template=bug_report.yaml


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,38 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: bug
+assignees: ''
+
```
annsch, line 7: should we add labels for different topics (html, react, angular, vue ...)?

mfranzke: this is mainly about the preselected label, but you could choose any further label additionally, which we couldn't anticipate for the users creating that issue.

annsch: @mfranzke my question links to the following
Bildschirmfoto 2023-07-12 um 10 53 14
 example: 

What about integrating this as well?

PR #750, opened 2023-03-28
docs: ✏️ Provide documentation on how to use db-input

annsch:


packages/components/src/components/input/docs/React.md
```diff
@@ -1,3 +1,27 @@
...
+
+<DBInput
+	variant={variant}
+	disabled={disabled}
+	label={label}
+	value={value}
+	description={description}
+	iconAfter={iconAfter}
+	iconBefore={iconBefore}
+	required={required}
```
mfranzke, line 25: I think it might be best to align those in between the frameworks and even also keep them to a "senseful" minimum (either only mandatory or most common attributes), especially to keep it simple here (but obviously show framework specifics, like e.g. events).

packages/components/src/components/input/docs/Vue.md
```diff
@@ -1,18 +1,41 @@
 ## Vue
 
+Load SCSS globally somewhere in your app:
+
+```
+@use "@db-ui/foundations/build/scss/variables.global" as *;
+@use "@db-ui/components/build/styles/db-ui-42-rollup" as *;
+@use "@db-ui/foundations/build/scss/color-classes" as *;
```
mfranzke, line 8:
```suggestion
@use "@db-ui/components/build/styles/db-ui-42-rollup" as *;
```

I think even for Vue we'll only include the main file in general.

mfranzke (approved):

PR #751, opened 2023-03-28
docs: ✏️ Provide documentation on how to use db-card

annsch:


packages/components/src/components/card/docs/HTML.md
```diff
@@ -1,3 +1,9 @@
 ## HTML
 
-TODO
+```html
```
mfranzke, line 3: On top of this HTML section, someone would even also need to include the global styling; so we could either lead to a global section within the Marketingportal, that explains those in general or mention it like we do for the other technologies above:
https://github.com/db-ui/mono/pull/751/files#diff-c6b642e04350b145d6bfddf6c07cdd96fc76e1db20259843cef12085b0ae98ecR3-R10

mfranzke (approved):

nmerget (approved):

PR #785, opened 2023-03-31
refactor: optimize iframe resizer

mfranzke:


mfranzke:
@dkolba to me it seems like that we're having a problem with the base URL here. Let's align.

dkolba (approved):

PR #786, opened 2023-03-31
feat: add datalist for input

nmerget:


packages/components/src/components/input/input.lite.tsx
```diff
@@ -93,6 +98,9 @@ export default function DBInput(props: DBInputProps) {
 
 	onMount(() => {
 		state._id = props.id ? props.id : 'input-' + uuid();
+		state._dataListId = props.dataListId
+			? props.dataListId
+			: `datalist-${state._id}`;
```
mfranzke, line 103:
```suggestion
		state._dataListId = props.dataListId || `datalist-${state._id}`;
```

From my understanding this would be the equivalent.

packages/components/src/components/input/input.lite.tsx
```diff
@@ -122,13 +130,15 @@ export default function DBInput(props: DBInputProps) {
 				aria-labelledby={state._id + '-label'}
 				disabled={props.disabled}
 				required={props.required}
+				defaultValue={props.defaultValue}
```
mfranzke, line 133: I'm not aware that an attribute named `defaultValue` exists. What's the expected result (especially, as we already do have `value`)?

packages/components/src/components/input/input.lite.tsx
```diff
@@ -148,6 +158,24 @@ export default function DBInput(props: DBInputProps) {
 			<Show when={state.iconVisible(props.iconAfter)}>
 				<DBIcon icon={props.iconAfter} class="icon-after" />
 			</Show>
+			<DBIcon icon="expand-more" class="datalist-icon" />
```
mfranzke, line 161:
```suggestion
```
Let's just set this icon via (S)CSS.

packages/components/src/components/input/input.lite.tsx
nmerget: This is a react property. The other framework should ignore it

packages/components/src/components/input/input.lite.tsx
nmerget: I guess this is the same topic, we had to discuss. How do we want to handle icons in components?

mfranzke: In general as a preparation to our alignment:
- Icons that we'd like to set based on the guidelines: Via SCSS mixin.
- Icons that the user could set: Either via SCSS mixin, or HTML attribute `data-icon`, depending on their tech stack.

The `db-icon` element was only a workaround for Web Components.

mfranzke: I've added these aspects and decisions to https://github.com/db-ui/mono/blob/ee3967f31ac60b66b5e349dca34757c4042f86ed/docs/adr-04-icons.md

packages/components/src/components/input/input.scss
```diff
@@ -159,6 +186,10 @@
 		padding: $db-spacing-fixed-xs var(--db-input-padding-inline-end)
 			$db-spacing-fixed-xs var(--db-input-padding-inline-start);
 
+		&::-webkit-calendar-picker-indicator {
+			display: none !important;
+		}
```
annsch, line 191: @nmerget why did you remove this? Should we use the calendar icon until we added the correct icon?

nmerget: fixed

annsch (approved):

PR #799, opened 2023-04-04
fix: icon scss integration & simplified icon reference

mfranzke:
The icon helper was missing a relevant SCSS placeholder, as well as meaningful fallbacks that could make including icons simpler.

And we needed to simplify those styling related declarations only to have them included within SCSS and won't need it within markup at all.


packages/foundations/scss/icon/_icons.helpers.scss
```diff
@@ -30,10 +31,12 @@
 			// * use !important to prevent issues with browser extensions that change fonts
 			font-family: var(
 				--icon-font-family,
-				"#{"icons-" + $size + "-" + $style}, missing-icons"
+				#{"icons-" + $size + "-" + $style},
+				missing-icons,
+				var(--db-base-icon-font-family)
```
nmerget, line 36: Should this become before missing-icons?

mfranzke: I've removed those leftovers.

packages/foundations/scss/icon/_icons.helpers.scss
```diff
@@ -29,11 +30,10 @@
 			/*** icon - partial ***/
 			// * use !important to prevent issues with browser extensions that change fonts
 			font-family: var(
```
nmerget, line 32: 90% of this is included in `%icon` inside `_icons.placeholder`, why do we write it again?

Can we just create a `%icon-partial` which is then included in `%icon` or the other way around?

mfranzke: yes, that's part of https://github.com/db-ui/mono/pull/509 already. But currently the SCSS Mixin way doesn't work at all, as the placeholder wasn't included previously.

__snapshots__/alert/alert.spec.tsx-snapshots/DBAlert-component-should-match-screenshot-for-device-desktop-1-chromium-linux.png
nmerget, line 1: I guess there is some error here

mfranzke: I assume, that this link is incorrectly configured, as internal links get that arrow according to the styleguide.

nmerget (approved):

PR #800, opened 2023-04-04
refactor(link): simplified icon reference

mfranzke:
Those styling related declarations are included within SCSS and won't need it within markup at all.

Must get merged after: https://github.com/db-ui/mono/pull/799

### DIFF
--- a/packages/components/src/components/link/link.lite.tsx
+++ b/packages/components/src/components/link/link.lite.tsx
@@ -1,5 +1,4 @@
 import { onMount, Show, useMetadata, useStore } from '@builder.io/mitosis';
-import { DBIcon } from '../icon';
 import { DBLinkState, DBLinkProps } from './model';
 
 useMetadata({
@@ -50,13 +49,6 @@ export default function DBLink(props: DBLinkProps) {
 				<link rel="stylesheet" href={state.stylePath} />
 			</Show>
 			{props.children}
-			<Show when={props.variant !== 'inline'}>
-				<DBIcon
-					icon={
-						props.content == 'external' ? 'link-external' : 'link'
-					}
-					icntxt={true}></DBIcon>
-			</Show>
 		</a>
 	);
 }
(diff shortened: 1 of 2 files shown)


packages/components/src/components/link/link.scss
```diff
@@ -57,8 +51,7 @@
 			outline-offset: var(--db-focus-outline-offset, 1px);
 		}
 
-		&::after,
-		.db-icon {
+		&::after {
```
annsch, line 54: is this sufficient for angular as well?

packages/components/src/components/link/link.lite.tsx
```diff
@@ -50,13 +49,6 @@ export default function DBLink(props: DBLinkProps) {
 				<link rel="stylesheet" href={state.stylePath} />
 			</Show>
 			{props.children}
-			<Show when={props.variant !== 'inline'}>
```
nmerget, line 53: You need to add a `data-icon={props.content == 'external' ? 'link-external' : 'link'}` to the ´<a>´ otherwise nothing will happen ^.^

packages/components/src/components/link/link.scss
mfranzke: `DBIcon` (and that for `.db-icon`) have been removed in total from source code, as we're only setting the icons via (S)CSS.

packages/components/src/components/link/link.lite.tsx
mfranzke: I would expect that this is already handled via https://github.com/db-ui/mono/pull/800/files/6d72bb678813184b383a6918ad78dfbae8e12b2d#diff-68ccb330bdb4f3c47599a47661753ac7953421f7827f8dbc009c35762001ad38R36-R42 (and the necessary changes out of https://github.com/db-ui/mono/pull/799).

mfranzke:
integrated into https://github.com/db-ui/mono/pull/799

PR #825, opened 2023-04-11
feat: add generation from props

nmerget:


.xo-config.cjs
```diff
@@ -38,7 +38,9 @@ module.exports = {
 				'@typescript-eslint/no-unsafe-call': 0, // We don't need this tsc will handle it anyway
 				'@typescript-eslint/naming-convention': 0, // Too much trouble for different frameworks in a monorepo
 				'@typescript-eslint/restrict-template-expressions': 0, // Legit it ,
-				'@typescript-eslint/no-unsafe-argument': 0 // valid for app
+				'@typescript-eslint/no-unsafe-argument': 0, // valid for app
+				'unicorn/prefer-logical-operator-over-ternary': 0, // legit for jsx apps
```
mfranzke, line 42: even though that it might be legit, isn't the shorter notation possible as well?

.xo-config.cjs
```diff
@@ -38,7 +38,9 @@ module.exports = {
...
+				'no-alert': 0 // we need it for showcase click events
```
mfranzke, line 43: couldn't we add an inline `eslint` ignore comment for this specific place instead?

.xo-config.cjs
nmerget: fixed

.xo-config.cjs
nmerget: fixed

mfranzke (approved):

PR #837, opened 2023-04-12
chore: update props  and docs for button

nmerget:
- Changed variants
  - `primary`
  - `outline` to `outlined`
  - `semi-Transparent` to `solid`
  - `transparent` to `text`
- add remaining props for the button from db-ui-elements
- add a description to props
- add migration guide for button
- renamed `icntxt` and `onlyIcon` to `noText`


packages/components/README.md
```diff
@@ -51,24 +54,6 @@ Or within your JavaScript files, with the related bundler as a prefix (in this c
...
-```
-
-**CSS**
-
-```js
-// main.js
-import "@db-ui/foundations/build/css/icon/icons.css";
-import "@db-ui/foundations/build/css/db-ui-foundations.css";
-```
-
```
mfranzke, line 71: As you've removed these, I'm not sure where you wanted to put them instead.

packages/components/src/components/button/button.spec.tsx
```diff
@@ -51,11 +51,7 @@ test.describe('DBButton component A11y', () => {
 		page,
 		mount
 	}) => {
-		await mount(
-			<DBButton icon="account" onlyIcon={true}>
-				lorem ipsum
-			</DBButton>
-		);
+		await mount(<DBButton icon="account">lorem ipsum</DBButton>);
```
mfranzke, line 54:
```suggestion
		await mount(<DBButton icon="account" icntxt="true">lorem ipsum</DBButton>);
```

As this is the "icon only" testcase, wouldn't it need that attribute instead of removing it altogether?

nmerget: The default with a icon button is without text. So `icntxt` is false and shouldn't be used

packages/components/README.md
nmerget: Add new PR for this: https://github.com/db-ui/mono/pull/841

packages/components/src/components/button/button.spec.tsx
mfranzke: A button without a visible text is more problematic regarding accessibility, as an element should always contain a visual textual description e.g. for voice control. That for I would expect that having an icon as an additional part of button with a visible text would be the default, and as this isn't clearly defined within the specification, I would at least interpret the order of the elements within Zeplin as well as within Marketingportal that Icon with text would be the default: https://marketingportal.extranet.deutschebahn.com/marketingportal/Design-Anwendungen/db-ux-design-system-v3/components/action/button#:~:text=Handlungsaufforderung%20zu%20beeintr%C3%A4chtigen.-,Content,-(Default)%20Text

nmerget: changed the default behaviour

packages/components/src/components/button/model.ts
```diff
@@ -8,15 +8,65 @@ import {
...
+	size?: 'small';
+
+	/**
+	 * Width of the button. Auto width based on text size, full width based on parent width.
+	 */
 	width?: 'full' | 'auto';
-	variant?: 'outline' | 'primary' | 'transparent' | 'semi-transparent';
+
+	/**
+	 * Width of the button. Auto width based on text size, full width based on parent width.
```
mfranzke, line 67: 👮‍♀️

packages/components/src/components/button/model.ts
```diff
@@ -8,15 +8,65 @@ import {
...
+	size?: 'small';
+
+	/**
+	 * Width of the button. Auto width based on text size, full width based on parent elements width.
+	 */
 	width?: 'full' | 'auto';
-	variant?: 'outline' | 'primary' | 'transparent' | 'semi-transparent';
+
+	/**
+	 * Width of the button. Auto width based on text size, full width based on parent elements width.
```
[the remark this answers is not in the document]
nmerget: ?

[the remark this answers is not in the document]
mfranzke: the `variant` property has the same comment / description like the `width` property.

mfranzke (changes requested):
image
The buttons contents aren't horizontally centered within [this branch](https://db-ui.github.io/mono/review/docs-button/components/button) anymore – on [main](https://db-ui.github.io/mono/review/main/components/button) it's still centered.

packages/components/src/components/button/button.lite.tsx
```diff
@@ -68,29 +68,30 @@ export default function DBButton(props: DBButtonProps) {
...
 				(props.className ? ' ' + props.className : '') +
-				(state.iconVisible(props.icon) && !props.icntxt
+				(state.iconVisible(props.icon) && props.noText
 					? ' is-icon-text-replace'
 					: '')
 			}
+			type={props.type}
+			disabled={props.disabled}
+			aria-label={props.label}
+			data-icon={state.iconVisible(props.icon) && props.icon}
```
mfranzke, line 81: image

The buttons contents aren't horizontally centered within [this branch](https://db-ui.github.io/mono/review/docs-button/components/button) anymore – on [main](https://db-ui.github.io/mono/review/main/components/button) it's still centered.

This seems to be related to the aspect, that data-icon="false" is currently set to buttons, even though that this is not a valid value and that attribute shouldn't get set at all in this case.

packages/components/src/components/button/model.ts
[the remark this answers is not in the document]
nmerget: oh good catch

packages/components/src/components/button/button.lite.tsx
nmerget: fixed

mfranzke (dismissed):

PR #860, opened 2023-04-18
feat: add new component select

nmerget:


packages/components/src/components/select/docs/Angular.md
```diff
@@ -0,0 +1,24 @@
...
+  ...
+})
+
+```
+
+### Use component
+
+```html app.component.html
+<!-- app.component.html -->
+<db-select>Select</db-select>
```
mfranzke, line 23: a string as a children of a `<select>` element wouldn't be the regular use case, but rather `<option>` elements.

packages/components/src/components/select/select.scss
```diff
@@ -0,0 +1,171 @@
...
+	&:has(select:disabled) {
+		// Decided against cursor: not-allowed, compare to e.g. https://phabricator.wikimedia.org/T121960
+		opacity: 0.25;
+		pointer-events: none;
+	}
+
+	label {
+		@extend %db-body-2xs;
+		top: $db-spacing-fixed-2xs;
+		transition: left, top, font-size 0.2s $db-transition-functional-timing;
```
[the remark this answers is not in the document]
nmerget: Okay, maybe you can change this for me?
We should use something similar for input and select

packages/components/src/components/select/model.ts
```diff
@@ -0,0 +1,56 @@
...
+	 * If you don't/can't use children/slots you can pass in the options as an array.
+	 */
+	options?: DBSelectOptionType[];
+}
+
+export type DBSelectOptionType = {
+	disabled?: boolean;
+	label?: string;
+	options?: DBSelectOptionType[];
+	value: string;
```
annsch, line 34: Shouldn't value allow more than just string, e.g. number or object or array?

packages/components/src/components/select/model.ts
```diff
@@ -0,0 +1,56 @@
...
+	 */
+	description?: string;
+
+	/**
+	 * If you don't/can't use children/slots you can pass in the options as an array.
+	 */
+	options?: DBSelectOptionType[];
+}
+
+export type DBSelectOptionType = {
```
annsch, line 30: These nested Props should also be documented in our props table, should we open an issue for that?

packages/components/src/components/select/select.lite.tsx
```diff
@@ -0,0 +1,181 @@
...
+			<div
+				class={state.getClassNames('db-select', props.className)}
+				data-variant={props.variant}
+				data-icon={props.icon}>
+				<Show when={state.stylePath}>
+					<link rel="stylesheet" href={state.stylePath} />
+				</Show>
+				{/* Required has to be true to use floating label */}
+				<select
+					required
```
mfranzke, line 104: `required` isn't necessary for floating labels on selects anymore, compare to https://github.com/db-ui/core/pull/422/commits/23170746cd978d490d5aeeb5a569079da40d9d9e

packages/components/src/components/select/docs/Angular.md
nmerget: fixed

packages/components/src/components/select/select.scss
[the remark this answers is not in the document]
nmerget: We handle this here: https://github.com/db-ui/mono/issues/908

packages/components/src/components/select/model.ts
nmerget: fixed

packages/components/src/components/select/select.lite.tsx
nmerget: fixed

packages/components/src/components/select/model.ts
nmerget: fixed

annsch (approved):

PR #871, opened 2023-04-19
refactor: generalizing some states

mfranzke:
Resolves https://github.com/db-ui/mono/issues/870

- [x] added missing `aria-invalid` attribute to `input`
- [x] extracted `invalid` prop
- [x] added global `describedbyid` prop

### DIFF
--- a/packages/components/src/shared/model.ts
+++ b/packages/components/src/shared/model.ts
@@ -23,6 +23,11 @@ export type GlobalProps = {
 	id?: string;
 
 	/**
+	 * [`aria-describedby`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-describedby) is used to link to the elements that describe the element with the set attribute.
+	 */
+	describedbyid?: string;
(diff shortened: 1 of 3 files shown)


packages/components/src/shared/model.ts
```diff
@@ -22,6 +22,11 @@ export type GlobalProps = {
 	 */
 	id?: string;
 
+	/**
+	 * [`aria-describedby`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-describedby) is used to link to the elements that describe the element with the set attribute.
+	 */
+	describedbyid?: string;
```
nmerget, line 28: Do we use this somewhere?

mfranzke: https://github.com/db-ui/mono/pull/515 would be the first usage, but it's a general attribute that could be used on any component.

nmerget (dismissed):

nmerget (approved):

PR #875, opened 2023-04-20
refactor: combine adaptive variants to reduce copy paste

nmerget:
- Add adaptive variants to foundations 
- Changed Alert, Infotext & Input to use new adaptive variants variable
- Fix issue with double icons for input

### DIFF
--- a/packages/components/src/components/infotext/infotext.scss
+++ b/packages/components/src/components/infotext/infotext.scss
@@ -1,3 +1,4 @@
+@use "sass:map";
 @use "@db-ui/foundations/build/scss/body-styles" as *;
 @use "@db-ui/foundations/build/scss/variables" as *;
 @use "@db-ui/foundations/build/scss/variables.global" as *;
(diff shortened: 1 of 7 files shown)


packages/components/src/components/infotext/infotext.scss
```diff
@@ -59,9 +26,9 @@ $variants: (
 		}
 	}
 
-	@each $name, $colors in $variants {
+	@each $name, $colors in $component-variants {
 		&[data-variant="#{$name}"] {
-			color: nth($colors, 1);
+			color: nth($colors, 4);
```
mfranzke, line 31: what about using an keys for the values within the `$component-variants` lowest level? Than you would be able to more clearly select for a specific entry in here by it's name instead of its index with [`map.get`](https://sass-lang.com/documentation/modules/map#get):

```suggestion
			color: map.get($colors, "successful");
```

or `map.get($colors, "border");`, `map.get($colors, "element");` or `map.get($colors, "on-bg");` for the other three entries.

nmerget: Yes, it's even better, I changed it.

mfranzke (approved):

PR #879, opened 2023-04-20
chore: retrieve link focus border radius from global variable

mfranzke:


packages/components/src/components/link/link.scss
```diff
@@ -44,7 +43,7 @@
 	}
 
 	&:focus-visible {
-		border-radius: 4px;
+		border-radius: component.$default-border-radius;
```
nmerget, line 46: We should keep it consistent and use `as *` so you could write `$default-border-radius`

mfranzke: related change: https://github.com/db-ui/mono/pull/890

nmerget: Now you are able to remove the prefix `component.`

nmerget (approved):

PR #893, opened 2023-04-21
refactor: optimized form validation approach

mfranzke:


packages/components/src/components/input/input.scss
```diff
@@ -326,9 +326,14 @@
 		}
 	}
 
-	input:required:invalid:not(:placeholder-shown),
-	input[aria-invalid="true"] {
-		@extend %adaptive-critical-component;
+	input {
+		&:is(:user-invalid):not([aria-invalid]),
+		// :user-invalid workaround, listing all relevant form validation attributes: https://developer.mozilla.org/en-US/docs/Web/HTML/Constraint_validation#validation-related_attributes
+		&:is(:required, [pattern], [min], [max], [step], [minlength], [maxlength]):invalid:not(:placeholder-shown),
```
mfranzke, line 332: we most likely won't need `:is(:required, [pattern], [min], [max], [step], [minlength], [maxlength])` as the browser could even already determinate whether the field is `:invalid` and this state might even get set programmatically via `setCustomValidity`.

mfranzke: this should probably get encapsulated within a negated `@supports` rule for `:user-invalid` as well, if that is possible.

mfranzke:
merge this one into https://github.com/db-ui/mono/pull/1545

mfranzke:
integrated into https://github.com/db-ui/mono/pull/1545

PR #1001, opened 2023-05-11
chore: add README.md for Power Apps

nmerget:


build-power-apps/README.md
mfranzke, line 1: Even though that our responsibility ends with providing the ZIP file within the workflow at the moment, could you please give some insight into how someone would be able to test these artifacts themselves with Power Apps (how to upload our components and use them within Power Apps)?

build-power-apps/README.md
```diff
@@ -0,0 +1,78 @@
...
+```tsx
+
+useMetadata({
+	...
+	component: {
+		includeIcon: true,
+		hasDisabledProp: true,
+		hasOnClick: true,
+		properties: [
+			{ name: 'children', type: 'SingleLine.Text' },
```
mfranzke, line 20: Could you please provide information on where to find the documentation for these types?

build-power-apps/README.md
```diff
@@ -0,0 +1,78 @@
...
+					{ key: 'Full', name: 'Full', value: 'full' },
+					{ key: 'Auto', name: 'Auto', value: 'auto' }
+				]
+			}
+		]
+	}
+});
+
+```
+
```
mfranzke, line 49:
```suggestion

Please find more information on the schema within these documentation pages:
- Regarding the `type`entries: https://learn.microsoft.com/en-us/power-apps/developer/component-framework/manifest-schema-reference/type

```

build-power-apps/README.md
mfranzke, line 1: Besides the localization, have you already tried out how to influence the displayed controls name out of our package? We should probably provide those as a well to "harden" our controls by providing some kind of a "default". From my understanding this would be the key `display-name-key`, out of the documentation on https://learn.microsoft.com/en-us/power-apps/developer/component-framework/manifest-schema-reference/control

mfranzke: Let's still add this to the `README`, as you do have this knowledge already and we could document even more than what was in scope for the POC.

nmerget: I have no idea about it, just saw it in another example.
We could add a issue for this to do some research.

build-power-apps/README.md
```diff
@@ -0,0 +1,86 @@
...
+<ItemGroup>
+	<--- your new component
+    <ProjectReference Include="..\button\DBButton.pcfproj" />
+    ...
+</ItemGroup>
+```
+
+For build run `npm run build:dotnet` or `npm run build:msbuild`, based on what you installed e.g. [MSBuild](https://learn.microsoft.com/de-de/visualstudio/msbuild/msbuild?view=vs-2022).
+Afterwards you should have a `.zip` file inside `build-power-apps/DBUI/bin/Release/DBUI.zip`.
+
```
mfranzke, line 79:
```suggestion
## How to build

To deploy a new component you need to include it to the `build-power-apps/DBUI/DBUI.cdsproj` file.
Find `ProjectReference` and add your component there:

```text
<ItemGroup>
	<--- your new component
    <ProjectReference Include="..\button\DBButton.pcfproj" />
    ...
</ItemGroup>
```

For build run `npm run build:dotnet` or `npm run build:msbuild`, based on what you installed e.g. [MSBuild](https://learn.microsoft.com/de-de/visualstudio/msbuild/msbuild?view=vs-2022).
Afterwards you should have a `.zip` file inside `build-power-apps/DBUI/bin/Release/DBUI.zip`.

## How to test

Every component has its own `package.json` and testing environment.
For the button we need to do those steps to test:

1. `cd build-power-apps/button`
2. `npm i`
3. `npm run refreshTypes`
4. `npm run start`

This opens a new browser window with a low-code editor and your component.

```

I think that we should change the order on these two.

PR #1048, opened 2023-05-18
refactor: simplify classnames dependency usage

mfranzke:
Resolves https://github.com/db-ui/mono/issues/927


packages/components/src/components/input/input.lite.tsx
```diff
@@ -128,7 +127,10 @@ export default function DBInput(props: DBInputProps) {
 				id={state._id}
 				name={props.name}
 				type={props.type || 'text'}
-				placeholder={props.placeholder ?? DEFAULT_VALUES.placeholder}
+				placeholder={
+					props.placeholder ??
+					state.defaultValues.placeholder.toString()
```
mfranzke, line 132: @nmerget why do you use a `.toString()` here?

packages/components/src/components/brand/brand.lite.tsx
```diff
@@ -54,7 +47,14 @@ export default function DBBrand(props: DBBrandProps) {
...
 	let component: any;
 	// jscpd:ignore-start
-	const state = useStore<DBBrandState>({});
+	const state = useStore<DBBrandState>({
+		defaultValues: {
+			anchorRef: '/',
+			src: './assets/images/db_logo.svg',
+			width: 34,
+			height: 24
+		}
```
mfranzke: What's the reason to move those into state?

packages/components/src/utils/index.ts
```diff
@@ -5,3 +5,31 @@ export const uuid = () => {
...
+			if (index != 0) {
+				result += ' ';
+			}
+
+			if (typeof arg === 'string') {
+				result += arg;
+			} else {
+				for (let key in arg) {
+					if (arg[key]) {
+						result += key;
```
mfranzke, line 27:
```suggestion
						result += ' ' + key;
```

Wouldn't this result in a missing space?

output/angular/ng-package.json
```diff
@@ -6,7 +6,6 @@
 	},
 	"allowedNonPeerDependencies": [
 		"@db-ui/foundations",
-		"@db-ui/components",
-		"classnames"
```
mfranzke, line 10: @nmerget I even also found that place that was introduced by your commit with https://github.com/db-ui/mono/commit/2fa081be70e9bf1976eb949137ee27aff748bdbe – this needs to get removed as well, doesn't it?

packages/components/src/components/input/input.lite.tsx
nmerget: Because of typescript. It can be `string | number `, but placeholder wants a string.

packages/components/src/components/brand/brand.lite.tsx
nmerget: There was an issue in vue, where the mitosis build in `DEFAULT_VALUES ` causing an issue :(

output/angular/ng-package.json
nmerget: yes

packages/components/src/components/input/input.lite.tsx
mfranzke: poor typescript …

packages/components/src/utils/index.ts
nmerget: changed it with `.trim()`

packages/components/src/components/input/input.lite.tsx
mfranzke: @nmerget on the other hand, as this is most likely mainly about these two entries, we could as well remove the `number` type in our declarations, as all of those values end up in an HTML-attribute anyhow and would be a `string` by that output:
https://github.com/db-ui/mono/pull/1048/files#diff-8ad41e5357c9f96eb5d9008d7d9386d0d7e0f30b11bafb7b9de155409c94c85eR54-R55

mfranzke:
@nmerget I like your style.

nmerget (approved):

PR #1107, opened 2023-06-06
fix(radio): added missing value property

mfranzke:
Resolves https://github.com/db-ui/mono/issues/1106


annsch (commented):
@mfranzke dindn't you forget to update Angular.md file as well?

mfranzke:
> @mfranzke dindn't you forget to update Angular.md file as well?

valid idea, but Angular already had the `value` property in its documentation (even though that it didn't work because the property was missing):
https://github.com/db-ui/mono/blob/main/packages/components/src/components/radio/docs/Angular.md?plain=1#L28

annsch (approved):

PR #1167, opened 2023-06-20
feat: add navigation-item component

nmerget:


packages/components/src/components/navigation-item/docs/HTML.md
```diff
@@ -0,0 +1,17 @@
...
+
+```html index.html
+<!-- index.html -->
+...
+<body>
+	<a href="mypath">
+		<div class="db-navigation-item">
+			<button class="db-navigation-item-button">NavigationItem</button>
+		</div>
+	</a>
```
mfranzke, line 15: `button` tags aren't allowed to get nested into an `a` tag – so we either have an `a` or `button` tag for interactive elements, depending on their behaviour & functionality.

nmerget: Removed the button :)

mfranzke: I could still see the button:
image

packages/components/src/components/navigation-item/navigation-item.lite.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+	// jscpd:ignore-end
+
+	return (
+		<li
+			ref={component}
+			aria-haspopup={state.hasAreaPopup}
+			class={state.getClassNames('db-navigation-item', props.className)}
+			data-width={props.width}
+			tabIndex={props.tabIndex || -1}
```
mfranzke, line 73: why would a list item itself need a tabindex at all?

packages/components/src/components/navigation-item/model.ts
```diff
@@ -0,0 +1,52 @@
...
+
+	/**
+	 * The disabled attribute can be set to [keep a user from clicking on the item](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#disabled).
+	 */
+	disabled?: boolean;
+
+	/**
+	 * To show puls indicator for active navigation items.
+	 */
+	isMainMenuItem?: boolean;
```
mfranzke, line 27: This could easily get scoped e.g. by `.db-mainnavigation` within CSS instead.

packages/components/src/components/brand/brand.scss
```diff
@@ -19,6 +19,7 @@
 }
 
 .db-logo {
+	max-width: none;
```
mfranzke, line 22: what for do we need this declaration?

packages/components/src/components/navigation-item/navigation-item.lite.tsx
```diff
@@ -0,0 +1,94 @@
...
+			}
+		}
+	}, [state.initialized]);
+
+	// jscpd:ignore-end
+
+	return (
+		<li
+			ref={component}
+			aria-haspopup={state.hasAreaPopup}
```
mfranzke, line 70: This should be included on an interactive element, so I've like to suggest to include a `button` HTML element within that triggers the action.

packages/components/src/components/navigation-item/navigation-item.lite.tsx
```diff
@@ -0,0 +1,94 @@
...
+			class={state.getClassNames('db-navigation-item', props.className)}
+			data-width={props.width}
+			tabIndex={props.tabIndex || -1}
+			data-main-menu={props.isMainMenuItem}
+			data-icon={state.iconVisible(props.icon) ? props.icon : undefined}
+			data-icon-after={
+				state.iconVisible(props.iconAfter) ? props.iconAfter : undefined
+			}
+			aria-current={props.active ? 'page' : undefined}
+			data-disabled={props.disabled}
```
mfranzke, line 80: if you'll include a `button` in here for the trigger element, this could as well take the `disabled`-attribute than.

packages/components/src/components/navigation-item/navigation-item.lite.tsx
```diff
@@ -0,0 +1,94 @@
...
+			<Show when={state.stylePath}>
+				<link rel="stylesheet" href={state.stylePath} />
+			</Show>
+			{props.children}
+
+			<menu class="db-sub-navigation" id={state.subNavigationId}>
+				<Slot name="sub-navigation"></Slot>
+			</menu>
+
+			<div class="active-indicator" />
```
mfranzke, line 91: As soon as you'll include a `button` element in here, you'll even also be able to replace this HTML element by a `::after { content: "" }` CSS construct even only.

packages/components/src/components/navigation-item/navigation-item.lite.tsx
```diff
@@ -0,0 +1,94 @@
...
+			data-disabled={props.disabled}
+			onClick={(event) => state.handleClick(event)}>
+			<Show when={state.stylePath}>
+				<link rel="stylesheet" href={state.stylePath} />
+			</Show>
+			{props.children}
+
+			<menu class="db-sub-navigation" id={state.subNavigationId}>
+				<Slot name="sub-navigation"></Slot>
+			</menu>
```
mfranzke, line 89: could the rendering of this element get prevented as well if there's no slotted content available ?

packages/components/src/components/navigation-item/navigation-item.scss
```diff
@@ -0,0 +1,161 @@
...
+	@extend %default-background-transition;
+	@extend %transparent-border;
+
+	@include get-variant-bg-color(0);
+	display: inline-flex;
+	cursor: pointer;
+
+	position: relative;
+
+	font-weight: normal;
```
mfranzke, line 18:
```suggestion
```

That's the regular style for fonts.

packages/components/src/components/navigation-item/navigation-item.lite.tsx
nmerget: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-haspopup#description 

I think the `<li>` is the button that triggers the popup. So the best way would be to add additionally https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role?

packages/components/src/components/navigation-item/navigation-item.lite.tsx
nmerget: But then I wouldn't be able to pass in a `<a>` tag and show this in a disabled state right?

packages/components/src/components/navigation-item/navigation-item.lite.tsx
mfranzke: We should prevent using ARIA-roles, as it's kind of an escape solution in case you couldn't change the HTML code and need to still provide semantic meaning (but not functionality) to non-semantic elements. In our case we could fully control the HTML and would even also need to question ourselves why we would need a list-item in the first place in case that we would want it to be a `button` instead.

packages/components/src/components/navigation-item/navigation-item.lite.tsx
mfranzke: I never understood why someone would want to disable an `<a>` tag anyhow. It mostly looks and behaves like a regular text afterwards, so the user either wouldn't recognize a link and/or wouldn't know how to make it active again. Plus deactivating buttons is also quite contentious, and nearly always related to users interaction that would need to happen before or has happened before so that the button gets into that deactivated state. What would be the equivalent for these deactivated links, what has happened or needs to happen for these links to get deactivated and especially what needs to happen for them to get into the "active" state again?

packages/components/src/components/navigation-item/docs/HTML.md
nmerget: I didn't update the `.md` file...

packages/components/src/components/navigation-item/navigation-item.lite.tsx
nmerget: We don't need this anymore, removed it

packages/components/src/components/brand/brand.scss
nmerget: https://github.com/db-ui/mono/pull/1184#discussion_r1243340960

packages/components/src/components/navigation-item/navigation-item.lite.tsx
nmerget: yes, fixed it

packages/components/src/components/navigation-item/navigation-item.scss
nmerget: fixed

packages/components/src/components/navigation-item/navigation-item.lite.tsx
nmerget: As soon as I include a `button` here, I have a nested button whenever I have another sub-menu.

Also we could move this indicator to an own `DBPuls` Component to have the same behaviour for tabs as well

packages/components/src/components/navigation-item/model.ts
nmerget: removed it for now, to test it in other branch

annsch (approved):

packages/components/src/components/navigation-item/docs/HTML.md
```diff
@@ -0,0 +1,20 @@
...
+For general installation and configuration look at the [components](https://www.npmjs.com/package/@db-ui/components) package.
+
+### Use component
+
+```html index.html
+<!-- index.html -->
+...
+<body>
+	<a href="mypath">
+		<li aria-haspopup="false" class="db-navigation-item">
```
mfranzke, line 12: `aria-haspopup` isn't supposed to get set on tags with this role:
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-haspopup#associated_roles

packages/components/src/components/navigation-item/docs/HTML.md
```diff
@@ -0,0 +1,20 @@
...
+			NavigationItem
+
+			<!-- If you want to use it as main-naivgation with pulse-->
+			<div class="active-indicator" />
```
mfranzke, line 16: We should check whether we really need a block level element for this and cannot achieve it either via `::after { content: ""; }` or at least changing it to a `<span>`.

packages/components/src/components/navigation-item/navigation-item.lite.tsx
```diff
@@ -0,0 +1,96 @@
...
+			role={state.hasAreaPopup ? 'button' : ''}
+			aria-haspopup={state.hasAreaPopup}
+			class={state.getClassNames('db-navigation-item', props.className)}
+			data-width={props.width}
+			data-icon={state.iconVisible(props.icon) ? props.icon : undefined}
+			data-icon-after={
+				state.iconVisible(props.iconAfter) ? props.iconAfter : undefined
+			}
+			aria-current={props.active ? 'page' : undefined}
+			data-disabled={props.disabled}
```
mfranzke, line 82: We should use `aria-disabled` instead of `data-disabled`, as the latter doesn't include any semantic meaning which we'd like to enable by using `aria-disabled` instead.

packages/components/src/components/navigation-item/navigation-item.lite.tsx
```diff
@@ -0,0 +1,96 @@
...
+			}
+		}
+	}, [state.initialized]);
+
+	// jscpd:ignore-end
+
+	return (
+		<li
+			ref={component}
+			role={state.hasAreaPopup ? 'button' : ''}
```
mfranzke, line 73: This would lead to having several buttons and menus wrapped into each other, which would either be a nightmare or totally abstracted content especially to screenreader users.

packages/components/src/components/navigation-item/navigation-item.lite.tsx
```diff
@@ -0,0 +1,96 @@
...
+		if (state.initialized && document && state.subNavigationId) {
+			const subNavigationSlot = document?.getElementById(
+				state.subNavigationId
+			) as HTMLMenuElement;
+			if (subNavigationSlot) {
+				const children = subNavigationSlot.children;
+				if (children?.length > 0) {
+					state.hasAreaPopup = true;
+				} else {
+					state.hideSubNavigation = true;
```
mfranzke, line 62: This it not necessarily about showing or hiding depending on a (dynamic like e.g. viewport size) condition, but not having any subitems and that for we wouldn't want to render that part. So it might be more clear and even also separated from the visual meaning to name it e.g. like `lacksSubNavigation`

PR #1179, opened 2023-06-23
refactor(brand): added missing properties defaults

mfranzke:
Resolves https://github.com/db-ui/mono/issues/1177


packages/components/src/components/brand/brand.lite.tsx
```diff
@@ -13,7 +13,9 @@ useMetadata({
 
 const DEFAULT_VALUES = {
 	anchorRef: '/',
-	src: './assets/images/db_logo.svg'
+	src: './assets/images/db_logo.svg',
+	width: 57,
```
nmerget, line 17: Why is this 57?

mfranzke: seems like an outdated number, got the newer dimensions out of Zeplin (took _Mobile_).

nmerget (approved):

PR #1185, opened 2023-06-26
feat: added new (brand component related) Power Apps fields

mfranzke:
Resolves https://github.com/db-ui/mono/issues/1178


packages/components/src/components/brand/brand.lite.tsx
```diff
@@ -2,22 +2,61 @@ import { onMount, Show, useMetadata, useStore } from '@builder.io/mitosis';
...
 	isAttachedToShadowDom: true,
 	component: {
 		// MS Power Apps
 		includeIcon: false,
-		properties: []
+		properties: [
+			{
+				name: 'anchorRef',
+				type: 'SingleLine.URL',
+				displayNameKey: 'Logo Link URL'
```
nmerget, line 21: We shouldn't add it like this. MS uses XML for those displayKeys we should add it such an XML because it supports multiple languages.

mfranzke: Totally agreed, but we could as well provide multiple ways.

mfranzke: Plus I haven't been able to easily identify of how to provide these XML files. Do you have any more information?

nmerget: https://github.com/microsoft/PowerApps-Samples/blob/master/component-framework/IncrementControl/IncrementControl/ControlManifest.Input.xml#L14

I guess this is an example? The documentation is very bad...

mfranzke: I'll get a better example and description of how to proceed with the translations instead by @sunday-out

PR #1186, opened 2023-06-26
chore: update playwright

nmerget:


mfranzke (approved):

mfranzke:
@nmerget when does somebody need to run `update:playwright` script?

nmerget:
> @nmerget when does somebody need to run `update:playwright` script?

We disabled it for dependabot, so we have to do it manually. I don't know which time interval we prefer

mfranzke (approved):

PR #1195, opened 2023-06-28
feat: add simple html files for development (styles)

nmerget:


packages/components/src/components/alert/index.html
mfranzke, line 1: @nmerget these are the generated files in this folder, aren't they? I'm struggling whether we should check those in or whether they should be either get generated on the fly or via ci/cd.

nmerget: I used the generated files, yes. But those html files are just for development. We wouldn't deliver it to someone. It like a playground to test styling and HTML very fast without any boilerplate from a showcase.

mfranzke (approved):

PR #1242, opened 2023-07-10
feat: add stylelint-use-logical to support all languages

nmerget:


packages/components/src/components/drawer/drawer.scss
```diff
@@ -7,63 +7,55 @@
...
 	}
 }
 
 %direction-up {
 	@extend %full-horizontal;
-	top: auto;
-	bottom: 0;
-	left: 0;
-	right: 0;
+	inset: auto 0;
```
mfranzke, line 39:
```suggestion
	inset: auto 0 0;
```

The previous declaration with all four properties doesn't equal the meaning of the proposed `inset` replacement.

packages/components/src/components/drawer/drawer.scss
```diff
@@ -7,63 +7,55 @@
...
+		border-start-start-radius: $default-card-border-radius;
 	}
 }
 %direction-down {
 	@extend %full-horizontal;
-	bottom: auto;
-	top: 0;
-	left: 0;
-	right: 0;
+	inset: logical 0 0 auto;
```
mfranzke, line 48:
```suggestion
	inset: 0 0 auto;
```

What does the `logical` keyword stand for? I'm not aware, that this exists.

nmerget: I did `stylelint **/*.scss --fix` and this was the result, don't know why it did this

mfranzke (approved):

PR #1243, opened 2023-07-10
refactor: removed withText from DBIcon, use DBInfotext instead

nmerget:


annsch:
Hey @nmerget  do you have some more information why you changed this? I assume DBIcon shouldn't be used with text, is this correct? Do you have a link with updated specs or sth similar where I could find som more information?

nmerget:
> Hey @nmerget do you have some more information why you changed this? I assume DBIcon shouldn't be used with text, is this correct? Do you have a link with updated specs or sth similar where I could find som more information?

You created this issue: https://github.com/db-ui/mono/issues/1151

I think I used this `withText` for the `db-button`, but now we use the `data-icon` so the property wasn't used by anything. And I think the correct way to inform a user with icon+text is `DBInfotext`.

annsch (approved):

PR #1251, opened 2023-07-11
refactor(power apps): added default values

mfranzke:
Resolves https://github.com/db-ui/mono/issues/1247


packages/components/_templates/power-apps/new/comp-xml.ejs.t
```diff
@@ -34,7 +34,6 @@ force: true
                   of-type="<%= prop.type %>"
                   <% } else if(prop.type === "Icon"){ -%>
                   of-type="Enum">
-                  <value name="None" display-name-key="None">_</value>
```
nmerget, line 37: Without this, you are not able to remove an icon from e.g. a button again after selecting one

mfranzke: makes sense to me, good catch.

nmerget (approved):

PR #1283, opened 2023-07-17
fix: issue with md files resolving images

nmerget:


docs/images/download/npm-ed1c24.svg
mfranzke, line 1: That's not the same image as it doesn't include the version information as in e.g. https://img.shields.io/npm/v/@db-ui/ngx-components.svg?color=rgb%28237%2C%2028%2C%2036%29

nmerget: Yes I removed the dynamic information. I think it's not necessary

mfranzke:
@nmerget I'm struggling whether it makes sense to check in the transformed files, but instead keep in the URLs, and only have those rewrites being done during build for the GitHub Pages exclusively – both because we only need them on this environment and we could even also keep dynamic contents like we would need them for items like this: https://github.com/db-ui/mono/pull/1283#discussion_r1266416265

nmerget:
> @nmerget I'm struggling whether it makes sense to check in the transformed files, or instead keep in the URLs, and only have those rewrites being done during build for the GitHub Pages exclusively – both because we only need them on this environment and we could even also keep dynamic contents like we would need them for items like this: [#1283 (comment)](https://github.com/db-ui/mono/pull/1283#discussion_r1266416265)

So it's fine to have those links (shield.io) on the default readme on GitHub?

mfranzke (dismissed):

mfranzke (approved):

PR #1288, opened 2023-07-18
Refactor tag

nmerget:
## Proposed changes

The `db-tag` has multiple behaviors (static, button, checkbox, ...). 
To reduce complexity we changed the `db-tag` to be just a wrapper for other `db-*` components: 
- `db-button`
- `db-radio`
- `db-checkbox`
- `span`

## Types of changes


mfranzke:
@nmerget I'm struggling with the aspect what for we would need to use our DB styled and functioning components which enhance existing HTML styles and behaviour instead of using the regular, standard HTML elements for this purpose within this component even only, as we don't seem to need the styling and probably most of the enhanced functionality, but only need the standard HTML elements behaviour and even also hide their appearance.

nmerget:
> @nmerget I'm struggling with the aspect what for we would need to use our DB styled and functioning components which enhance existing HTML styles and behaviour instead of using the regular, standard HTML elements for this purpose within this component even only, as we don't seem to need the styling and probably most of the enhanced functionality, but only need the standard HTML elements behaviour and even also hide their appearance.

I think that's a Gretchen question...

I thought about this, too. But all other component libraries I've used use their own components inside a wrapper component.

We should talk about this with the team. But I think it is easier for a user to write:

`<DBTag><DBRadio>Radio1</DBRadio></DBTag>`

instead of:

`<DBTag><input type="radio" id="radio1"></input><label htmlFor="radio1">Radio1</label ></DBTag>`

packages/components/src/components/navigation-item/navigation-item.scss
```diff
@@ -5,6 +5,7 @@
 @use "@db-ui/foundations/build/scss/animation/animations";
 @use "@db-ui/foundations/build/scss/icon/icon-helpers";
 @use "../../styles/db-puls";
+@use "../../styles/icon-passing" as icon-passing;
```
mfranzke, line 8:
```suggestion
@use "../../styles/icon-passing";
```

packages/components/src/styles/_icon-passing.scss
```diff
@@ -0,0 +1,30 @@
+@use "@db-ui/foundations/build/scss/variables";
+@use "@db-ui/foundations/build/scss/icon/icon-helpers";
+
+@mixin icon-passing($inline-start: variables.$db-spacing-fixed-sm) {
+	&[data-icon] {
```
mfranzke, line 5:
```suggestion
```
This shouldn't be necessary, as you would only selectively use this mixin, without the need to select for that attribute as well.

packages/components/src/components/tag/docs/React.md
```diff
@@ -8,7 +8,13 @@ For general installation and configuration take a look at the [react-components]
...
 import { DBTag } from "@db-ui/react-components";
 
-const App = () => <DBTag>Tag</DBTag>;
+const App = () => <>
+	<DBTag><DBButton>Tag as Button</DBButton></DBTag>
+	<DBTag><DBLink>Tag as Link</DBLink></DBTag>
+	<DBTag><DBCheckbox>Tag as Checkbox</DBCheckbox></DBTag>
+	<DBTag><DBRadio>Tag as Radio</DBRadio></DBTag>
+	<DBTag>Static Tag</DBTag>
+	<>;
```
mfranzke, line 17:
```suggestion
	</>;
```
Isn't it that this needs to be a closing tag ?

showcases/react-showcase/src/components/form/index.tsx
```diff
@@ -182,6 +184,8 @@ const FormComponent = () => {
 					</fieldset>
 				</form>
 			</div>
+
+			<div></div>
```
mfranzke, line 188:
```suggestion
```

packages/components/src/components/tag/tag.scss
```diff
@@ -1,70 +1,152 @@
...
+		@include icon-helpers.is-icon-text-replace();
+		padding: 0;
+
+		label,
+		span {
+			@include icon-helpers.is-icon-text-replace();
+		}
+	}
+
+	&:not(&[data-no-text="true"]) {
```
mfranzke, line 24:
```suggestion
	&:not([data-no-text="true"]) {
```

Is there a reason to include the `db-tag` class in there again ?

mfranzke (approved):

PR #1292, opened 2023-07-18
feat: add "weak" emphasis for drawer

nmerget:
## Proposed changes

Add new attribute for `db-drawer`: `data-emphasis` to control the opacity of the background

## Types of changes

### DIFF
--- a/packages/components/src/styles/_dialog-init.scss
+++ b/packages/components/src/styles/_dialog-init.scss
@@ -1,5 +1,10 @@
 @use "@db-ui/foundations/build/scss/variables" as *;
 
+%backdrop {
+	background-color: $db-colors-neutral-on-bg-enabled;
+	opacity: 0.64;
+}
+
 %dialog-container {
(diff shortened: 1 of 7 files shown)


packages/components/src/styles/_dialog-init.scss
```diff
@@ -1,5 +1,10 @@
 @use "@db-ui/foundations/build/scss/variables" as *;
 
+%backdrop {
+	background-color: $db-colors-neutral-on-bg-enabled;
+	opacity: 0.64;
+}
+
```
mfranzke, line 7: @nmerget extracting those declarations into SCSS placeholders seems to be obsolete to me (right now), doesn't it?

mfranzke (approved):

packages/components/src/styles/_dialog-init.scss
nmerget: Yes, it is. But we might use it in the near future

PR #1298, opened 2023-07-19
refactor: replaced individual staged files linting script by lint-staged

mfranzke:
Resolves https://github.com/db-ui/mono/issues/1297


.lintstagedrc
```diff
@@ -0,0 +1,4 @@
+{
+	"*.{css,scss}": "stylelint --fix",
+	"*.{js,css,scss,md,adoc,ts,tsx}": "prettier --write"
```
mfranzke, line 3: @nmerget @dkolba @annsch I liked this approach by `lint-staged` to even also provide the possibility to restrict the staged files that should get linted (especially for stylelint); are there any files missing for prettier that we regularly use? It's especially nice to prevent some files being prettified that aren't supported (compare to e.g. https://github.com/db-ui/mono/issues/1256).

nmerget: `sh,yml`

nmerget (approved):

PR #1299, opened 2023-07-19
refactor: drawer backdrop (combine noBackdrop and emphasis)

nmerget:
## Proposed changes

Combine 2 attributes to into one to align with design.

## Types of changes


packages/components/src/styles/_dialog-init.scss
```diff
@@ -1,10 +1,17 @@
...
-	background-color: $db-colors-neutral-on-bg-enabled;
+	background-color: $backdrop-color;
 	opacity: 0.64;
 }
 
+%backdrop-weak {
+	background-color: $backdrop-color;
+	opacity: 0.32;
+}
+
```
mfranzke, line 14: What for do you need a SCSS placeholder for this instead of "just" putting those declarations on line 40?

nmerget: Because we should use it in the future for other components as well :P

mfranzke: How do we / will we document the existence of these SCSS mixins, SCSS placeholders and Design Tokens ?

mfranzke: I've opened an issue as a follow up: https://github.com/db-ui/mono/issues/1310

PR #1301, opened 2023-07-19
feat: add badge component

nmerget:
Resolves https://github.com/db-ui/mono/issues/1104


packages/components/src/components/badge/badge.lite.tsx
```diff
@@ -0,0 +1,42 @@
+import { onMount, Show, useMetadata, useStore } from '@builder.io/mitosis';
+import { DBBadgeState, DBBadgeProps } from './model';
+import { cls } from '../../utils';
+
+useMetadata({
+	isAttachedToShadowDom: true,
+	component: {
+		// MS Power Apps
+		includeIcon: false,
+		properties: []
```
mfranzke, line 10: In general this should be possible even also to get used with Power Apps, even though that we wouldn't have it included as a nested component obviously, but only as a "standalone" attachment. What do you think?

nmerget: I think we shouldn't build the power-apps components in the first place. This would be another task in our process.
Moreover this is the perfect example why we should create an own extension for power-apps:
I guess the badge should be a nested component (besides of the standalone) in every other power-apps component and you can make it visible and change the text by an extra property inside the other component. Otherwise it wouldn't be possible to handle the dot on a component.

packages/components/src/components/badge/badge.scss
```diff
@@ -0,0 +1,141 @@
+@use "sass:map";
+@use "@db-ui/foundations/build/scss/init" as *;
```
mfranzke, line 2: what for is this init file necessary to get included? It mainly brings in a lot of declarations that are especially included in the incremental file selecting global tags, but wouldn't be necessary in there.

packages/components/src/components/badge/badge.scss
```diff
@@ -0,0 +1,141 @@
...
+@use "@db-ui/foundations/build/scss/variables" as *;
+@use "@db-ui/foundations/build/scss/variables.global" as *;
+@use "@db-ui/foundations/build/scss/color-placeholder" as *;
+@use "@db-ui/foundations/build/scss/color/color-variants" as *;
+@use "@db-ui/foundations/build/scss/helpers/component" as *;
+@use "@db-ui/foundations/build/scss/body-styles" as *;
+@use "../../styles/button-components";
+
+%tag-colors {
+	// TODO: Color from zeplin does not exist is it neutral-4??
```
mfranzke, line 12: do you have chance to clarify this question with creatives previous to merging this PR?

packages/components/src/components/badge/badge.scss
```diff
@@ -0,0 +1,141 @@
...
+	@extend %tag-colors;
+	@extend %component-border;
+	@extend %db-body-sm;
+	@extend %default-button;
+
+	display: flex;
+	border-radius: 42px; // extreme radius
+	padding-inline: $db-spacing-fixed-xs;
+	block-size: fit-content;
+	pointer-events: none;
```
mfranzke, line 96: a `span` HTML tag wouldn't necessarily have any `pointer-events` out of the box, or am I missing something here?

packages/components/src/components/badge/badge.scss
```diff
@@ -0,0 +1,141 @@
...
+	> span:empty,
+	&[data-no-text="true"] {
+		block-size: var(--badge-size);
+		inline-size: var(--badge-size);
+		padding: 0;
+	}
+
+	&:has(.db-icon) {
+		padding-block: $db-spacing-fixed-2xs;
+		padding-inline: $db-spacing-fixed-2xs;
```
mfranzke, line 108:
```suggestion
		padding: $db-spacing-fixed-2xs;
```

Doesn't this mean the general `padding`?

packages/components/src/components/badge/badge.scss
```diff
@@ -0,0 +1,141 @@
...
+
+		&:empty,
+		> span:empty,
+		&[data-no-text="true"] {
+			padding: 0;
+		}
+
+		&:has(.db-icon) {
+			padding-block: $db-spacing-fixed-3xs;
+			padding-inline: $db-spacing-fixed-3xs;
```
mfranzke, line 129:
```suggestion
			padding: $db-spacing-fixed-3xs;
```
Doesn't this mean the general `padding`?

packages/components/src/components/badge/docs/HTML.md
```diff
@@ -0,0 +1,13 @@
...
+
+For general installation and configuration look at the [components](https://www.npmjs.com/package/@db-ui/components) package.
+
+### Use component
+
+```html index.html
+<!-- index.html -->
+...
+<body>
+	<div class="db-badge">Badge</div>
```
mfranzke, line 11:
```suggestion
	<span class="db-badge">Badge</span>
```

Our component itself is mainly a span even only.

packages/components/src/styles/db-ui-components.scss
```diff
@@ -70,3 +73,9 @@ db-alert,
 dbalert {
 	inline-size: inherit;
 }
+
+// for db-badge
+
+*:has(.db-badge[data-placement^="corner"]) {
```
mfranzke, line 79:
```suggestion
:has(.db-badge[data-placement^="corner"]) {
```

Just to prevent an asterisk for once (as it should be optional in this case).

showcases/react-showcase/src/components/badge/index.tsx
```diff
@@ -0,0 +1,111 @@
...
+					<DBInfotext variant="informational" size="small">
+						{children}
+					</DBInfotext>
+				)}
+			</>
+		)}
+
+		{placement && placement !== 'inline' && !example && (
+			<>
+				<DBButton icon="account" variant="outlined" noText>
```
mfranzke, line 40: `noText` is mentioned to be an Angular workaround – so why is it used on this example as well?:
https://github.com/db-ui/mono/pull/1301/files#diff-87778bd8d4faa639f834ff2aedc6b53de727fb2072a121295310d681c2f6d972R10-R14

showcases/vue-showcase/src/components/badge/Badge.vue
```diff
@@ -0,0 +1,100 @@
...
+	<DefaultComponent title="DBBadge" :variants="defaultComponentVariants">
+		<template
+			#example="{ exampleIndex, variantIndex, exampleName, exampleProps }"
+		>
+			<template v-if="!exampleProps.placement && !exampleProps.example">
+				<DBBadge
+					:variant="exampleProps.variant"
+					:emphasis="exampleProps.emphasis"
+					:size="exampleProps.size"
+					:noText="exampleProps.noContent"
```
mfranzke, line 22: `noText` is mentioned to be an Angular workaround – so why is it used on this example as well?:
https://github.com/db-ui/mono/pull/1301/files#diff-87778bd8d4faa639f834ff2aedc6b53de727fb2072a121295310d681c2f6d972R10-R14

packages/components/src/components/badge/model.ts
```diff
@@ -0,0 +1,37 @@
...
+	EmphasisProps,
+	GlobalProps,
+	GlobalState,
+	SizeProps
+} from '../../shared/model';
+
+export interface DBBadgeDefaultProps {
+	/**
+	 * Workaround property for angular to force a dot style without any text in it.
+	 * Angular adds a comment with binding to the component which disables the :empty css selector.
```
mfranzke, line 12: The comment itself shouldn't be a problem for the `:empty` selector, but the additional whitespaces that are included in this component are, compare to https://developer.mozilla.org/en-US/docs/Web/CSS/:empty

mfranzke, line 12:
```suggestion
	 * Define the text within the badge to get hidden.
```

packages/components/src/components/badge/model.ts
```diff
@@ -0,0 +1,37 @@
...
+	 */
+	noText?: boolean;
+
+	/**
+	 * The corner attributes changes the position to absolute and adds a transform based on the placement.
```
mfranzke, line 17:
```suggestion
	 * The `placement` attributes `corner-*` values change the position to absolute and adds a transform based on the placement.
```

packages/components/src/components/badge/badge.scss
nmerget: Yes, for now, it is neutral-4. They changed the colors and at that time the tokens weren't up to date. 
But we gonna change the colors again with the upcoming dark-mode, so we will change this maybe again in the future.

__snapshots__/badge/showcase/chromium/regular/neutral-0/DBBadge-should-match-screenshot.png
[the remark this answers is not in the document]
nmerget: We need to align this with UX they deliver the overview and I hope they have some ideas why they show it like this.

The better question is: What's the difference between the Tag and Badge with only the label. Because it is just the border-radius. So we should ask ourselves, do we need the Badge with the label inline? Or should a user use a tag for this purpose and the badge is only for the absolute placement?

packages/components/src/components/badge/index.html
```diff
@@ -0,0 +1,16 @@
...
+		<title>DBBadge</title>
+		<link rel="stylesheet" href="/styles/db-ui-42.css" />
+	</head>
+	<body style="padding: var(--db-spacing-fixed-md)">
+		<db-badge
+			><span class="db-badge"
+				><!--binding-->
+				<span>test</span>
+			</span></db-badge
+		>
```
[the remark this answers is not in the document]
nmerget: Yes, I used it to test Angular faster. This is the example why `:empty` did not work...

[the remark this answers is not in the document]
mfranzke: > Yes, I used it to test Angular faster. This is the example why `:empty` did not work...

Well, because it's not empty, it contains the text "test" – this wouldn't work in any framework as well.

showcases/react-showcase/src/components/badge/index.tsx
nmerget: But that's the `<DBButton>` here, it'S valid in this case :P

packages/components/src/components/badge/model.ts
nmerget: It's not the whitespace...

Angular adds `<!--bindings={}-->` to every custom component. Which breaks `:empty` :(

image

packages/components/src/components/badge/index.html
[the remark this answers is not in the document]
nmerget: You are right, the `index.html` is just a playground. I wrote "test" and removed it while testing. I guess the last state was with "test"

__snapshots__/badge/showcase/chromium/regular/neutral-0/DBBadge-should-match-screenshot.png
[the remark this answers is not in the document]
nmerget: I guess we will discuss this in a meeting

packages/components/src/components/badge/badge.scss
nmerget: fixed

packages/components/src/components/badge/badge.scss
nmerget: fixed

packages/components/src/components/badge/badge.scss
nmerget: fixed

packages/components/src/components/badge/docs/HTML.md
nmerget: fixed

packages/components/src/styles/db-ui-components.scss
nmerget: fixed

showcases/showcase-styles.css
```diff
@@ -79,3 +80,16 @@
 	flex-wrap: wrap;
 	gap: var(--db-spacing-fixed-md);
 }
+
+.badge-inline-container {
+	display: flex;
+	gap: var(--db-spacing-fixed-sm);
+	padding-block: var(--db-spacing-fixed-xs);
+	padding-inline: var(--db-spacing-fixed-sm);
```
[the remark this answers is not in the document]
nmerget: fixed

showcases/vue-showcase/src/components/badge/Badge.vue
nmerget: fixed

packages/components/src/components/badge/badge.scss
nmerget: fixed

packages/components/src/components/badge/model.ts
mfranzke: But it's not the comment itself, compare to: https://jsbin.com/vokomeq/edit?html,css,output

mfranzke: I've removed the whitespaces out of the including template with commit 027d0885765c8c4f25f50a8d15cf049381c02f5d and now the selector even also works on `:empty`:
image

showcases/react-showcase/src/components/badge/index.tsx
mfranzke: that was a drill.

mfranzke (approved):

mfranzke:
Disabled auto-merge again for creative colleagues being able to review and approve.

PR #1371, opened 2023-08-07
docs(angular): removed incorrect `:empty` pseudo selector assumption

mfranzke:
Resolves https://github.com/db-ui/mono/issues/1370


nmerget:
But why shouldn't we inform the developer about the `:empty`? It does not work in angular but in other frameworks. I used it and had problems. So somehow I want to write down my experience :S

mfranzke:
> But why shouldn't we inform the developer about the `:empty`? It does not work in angular but in other frameworks. I used it and had problems. So somehow I want to write down my experience :S

Because it's not related to Angular (which "only" adds a comment, that isn't influenced by `:empty`), but adding whitespaces that are actually related to `:empty`.

nmerget:
It's not about the whitespaces ^.^

Like I wrote here: https://github.com/db-ui/mono/pull/1301#discussion_r1286803964

The problem are the comments inside custom components

nmerget (approved):

nmerget (approved):

Issue #1372, opened 2023-08-07
Incorrect local font names


mfranzke:
### Which generators are impacted?

- [X] All
- [ ] HTML
- [ ] React
- [ ] Angular
- [ ] Vue
- [ ] Web components
- [ ] Power Apps

### Reproduction case

Open https://db-ui.github.io/mono/ and you'll recognize that we're still loading DB Web Fonts that should actually get referenced from locally installed fonts.

### Expected Behaviour

Locally installed fonts would be the source for font rendering, so that those DB Web Fonts wouldn't need to get loaded on DB managed devices, on which those are preinstalled.

### Screenshots

<img width="1767" alt="image" src="https://github.com/db-ui/mono/assets/787658/d3c82a67-80b6-4e5a-9700-4692d8ab23f9">

### Browser version

None

### Add any other context about the problem here.

Font overview: https://marketingportal.extranet.deutschebahn.com/marketingportal/Marke-und-Design/Basiselemente/Typografie/Fuer-digitale-Anwendungen

PR #1373, opened 2023-08-07
fix: defined correct local font names #1372

mfranzke:
Resolves https://github.com/db-ui/mono/issues/1372

### DIFF
--- a/packages/foundations/scss/_font-faces.scss
+++ b/packages/foundations/scss/_font-faces.scss
@@ -24,7 +24,7 @@
 	font-style: normal;
 	font-weight: $db-font-db-screensans-regular-font-weight;
 	src:
-		local($db-font-db-screensans-regular-name),
+		local($db-font-db-screensans-regular-local-name),
 		url($fonts-path + $db-font-db-screensans-regular-woff2) format("woff2");
 }
 
--- a/packages/foundations/tokens/font.jsonc
+++ b/packages/foundations/tokens/font.jsonc
@@ -7,6 +7,12 @@
 					"target": "web"
 				}
 			},
+			"local-name": {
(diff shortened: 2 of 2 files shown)


packages/components/src/fonts.html
```diff
@@ -0,0 +1,43 @@
+<!doctype html>
```
nmerget, line 1: We should add another vite app in foundations with this file and maybe colors & icons. So we could watch it separately from components?

mfranzke: I've opened a followUp task for this, to get this "fix" out of the door asap:
https://github.com/db-ui/mono/issues/1405

packages/foundations/scss/_font-faces.scss
```diff
@@ -6,7 +6,8 @@
 	font-style: normal;
 	font-weight: $db-font-db-screenhead-light-font-weight;
 	src:
-		local($db-font-db-screenhead-light-name),
+		local($db-font-db-screenhead-light-local-name),
+		local($db-font-db-screenhead-light-local-name-mac),
```
mfranzke, line 10: I'm struggling whether we should change the order of these, as the `*-name-mac`-variants include the name of the files according to the original names (`DB Screen Sans`), whereas the `*-name*` only include `DB Sans` etc.
So the cascade might be better to have `DB Screen Sans` and then `DB Sans`

mfranzke: Or we call these variables `*-name` and `*-name-alternative` (for the shorter Windows names).

nmerget: Feel free to rename it :)

nmerget (approved):

PR #1396, opened 2023-08-10
feat: add dynamic colors to change them via css-properties

nmerget:


__snapshots__/alert/showcase/chromium/regular/neutral-0/DBAlert-should-match-screenshot.png
mfranzke, line 1: @nmerget the alert has lost it's white background color.

__snapshots__/link/showcase/chromium/regular/neutral-0/DBLink-should-match-screenshot.png
mfranzke, line 1: The space in between text and icon has been reduced.

packages/foundations/scss/_variables.global.scss
```diff
@@ -23,3 +25,95 @@ $db-spacing-responsive-sm: var(--db-spacing-responsive-sm);
 $db-spacing-responsive-md: var(--db-spacing-responsive-md);
 $db-spacing-responsive-lg: var(--db-spacing-responsive-lg);
 $db-spacing-responsive-xl: var(--db-spacing-responsive-xl);
+
+// Variants for adaptive components like input, select, alert, ...
+
+// neutral
+$db-colors-neutral-bg: var(
+		--neutral-bg-enabled,
```
mfranzke, line 33: Let's prefix all of the variables that we introduce ourselves.

packages/components/src/components/alert/alert.scss
```diff
@@ -54,7 +54,6 @@
 }
 
 .db-alert {
-	@extend %db-bg-neutral-0;
```
mfranzke, line 57: The specification still includes a white background color.

packages/components/src/components/checkbox/checkbox.scss
```diff
@@ -10,7 +10,6 @@
 $font-size-height: calc(var(--db-base-font-size) * var(--db-base-line-height));
 
 .db-checkbox {
-	@extend %default-color;
```
mfranzke, line 13: The specification still includes a white background color.

__snapshots__/alert/showcase/chromium/regular/neutral-0/DBAlert-should-match-screenshot.png
nmerget: So the alert should be "adaptive" as default. Before we used the neutral-0 (#ffffff) as a fixed value.

At the moment it lays above neutral-1 so it is not 100% white. 
Did you mean this?

packages/foundations/scss/_variables.global.scss
nmerget: I had it with `db` as a prefix, but I think it would be better to use the future name of the design-system as a prefix...

packages/components/src/components/alert/alert.scss
nmerget: 'specification': Do you ment the zeplin screens with this?

It is on a white background, but it should be adaptive, based on the undelaying background

__snapshots__/link/showcase/chromium/regular/neutral-0/DBLink-should-match-screenshot.png
nmerget: fixed

packages/foundations/scss/_variables.global.scss
nmerget: We should change it in the future

__snapshots__/link/showcase/mobile-chrome/regular/neutral-0/DBLink-should-match-screenshot.png
mfranzke, line 1: @nmerget the links text and icon spacing slightly changes, seems like related to the changes out of this branch.

https://github.com/db-ui/mono/assets/787658/ca2d0564-0dfc-43b2-841c-07e291e2a9ed

packages/components/src/components/input/input.scss
```diff
@@ -76,6 +114,8 @@
 	--icon-margin-after: 0;
 	--db-input-label-max-width: 80%;
 
+	@extend %input-color-variants;
+
```
mfranzke, line 118: Why do you add a new SCSS placeholder, but only use it once? Is it a preparation for some further usage in the future?

packages/components/src/components/link/link.scss
```diff
@@ -74,12 +77,20 @@
...
-		@extend %db-primary-text-ia;
+		color: $db-colors-primary;
+
+		&:hover {
+			color: $db-colors-primary-hover;
+		}
+
+		&:active {
+			color: $db-colors-primary-pressed;
+		}
```
mfranzke, line 88: Isn't there a SCSS placeholder, that already encapsulates these three different states (regular, hover and active, generated out of `generateInteractiveVariants` function?

packages/foundations/scripts/color-classes-generator.js
mfranzke, line 1: interesting move, a de-abstraction …

packages/components/src/components/input/input.scss
nmerget: It's just for gaining a better readability, is there another option to solve this?

packages/components/src/components/link/link.scss
nmerget: I had a placeholder for this problem before, but it is only required for link and only for the brand/primary color.
Furthermore, this will be renamed to `primary-text-*` which is a special case for the Link with the brand color.

I would keep it inside the Link for now, because this "text-color" should not be used elsewhere.

__snapshots__/link/showcase/mobile-chrome/regular/neutral-0/DBLink-should-match-screenshot.png
nmerget: fixed the spacing for the link

mfranzke (approved):

PR #1422, opened 2023-08-17
fix: all colors in components to work with dark-mode

nmerget:


__snapshots__/card/component/chromium/DBCard-should-match-screenshot-for-color-variant-neutral-1.png
mfranzke, line 1: The white background-color removal is intended (due to neutral)?

__snapshots__/card/component/chromium/DBCard-should-match-screenshot-for-color-variant-neutral-3.png
mfranzke, line 1: The white background-color removal is intended (due to neutral)?

__snapshots__/card/component/chromium/DBCard-should-match-screenshot-for-color-variant-neutral-4.png
mfranzke, line 1: The white background-color removal is intended (due to neutral)?

__snapshots__/card/component/chromium/DBCard-should-match-screenshot-for-color-variant-secondary.png
mfranzke, line 1: The white background-color removal is intended (I can't find any variants for the Cards in Zeplin at the moment)?

__snapshots__/card/component/chromium/DBCard-should-match-screenshot-for-color-variant-neutral-1.png
nmerget: Good catch, this variant doesn't exist anymore

__snapshots__/card/component/chromium/DBCard-should-match-screenshot-for-color-variant-neutral-4.png
nmerget: https://github.com/db-ui/mono/pull/1422#discussion_r1298043108

__snapshots__/card/component/chromium/DBCard-should-match-screenshot-for-color-variant-secondary.png
nmerget: https://github.com/db-ui/mono/pull/1422#discussion_r1298043108

__snapshots__/card/component/chromium/DBCard-should-match-screenshot-for-color-variant-neutral-3.png
nmerget: https://github.com/db-ui/mono/pull/1422#discussion_r1298043108

packages/components/src/components/link/link.scss
```diff
@@ -24,11 +24,6 @@
 	text-underline-position: from-font;
 	text-decoration-thickness: to-rem($pxValue: 1);
 
-	// Changing color on hover should be animated by a transition
-	transition:
-		outline 0.06s,
-		color $db-transition-color;
-
```
mfranzke, line 31: why have you removed this? This doesn't look equal anymore on https://db-ui.github.io/mono/review/fix-colors-dark-mode/components/link/overview and https://db-ui.github.io/mono/version/v0.0.32/components/link/overview

packages/components/src/components/select/select.scss
```diff
@@ -150,6 +150,10 @@
 		margin-block: $db-spacing-fixed-2xs 0;
 	}
 
+	option {
+		background-color: var(--db-current-bg-color);
+	}
```
mfranzke, line 155: what's that declaration for? In general you can't style the `option` elements within `select` elements.

packages/foundations/scss/db-ui-foundations.scss
```diff
@@ -5,8 +5,17 @@
...
 :root {
 	@extend %db-ui-regular;
-	@extend %db-bg-neutral-0;
+	@extend %db-bg-neutral;
+}
+
+[class^="db-"] {
+	@extend %default-color-transition;
+
+	::before {
```
mfranzke, line 18: why only `::before`, but not `::after`? Is it that you had some specific elements / constellations in mind?

packages/foundations/scss/_init.scss
```diff
@@ -32,7 +32,12 @@ html {
 }
 
 // Make sure every elements has the current color from container above
-body {
+body,
+button,
+input,
+textarea,
+summary,
+select {
```
mfranzke, line 40:
```suggestion
button,
input,
textarea,
summary,
select {
    color: inherit;
}
body,
```

I'm struggling with the aspect whether we would be more flexible with an approach like this.

packages/foundations/scss/icon/_icons.placeholder.scss
```diff
@@ -1,12 +1,14 @@
...
 @use "../helpers/functions" as *;
+@use "../variables.global" as *;
 
 %icon {
 	// Hiding icon from screenreaders
 	-webkit-alt: "";
 	alt: "";
 
 	content: var(--icon-glyph);
+	color: $db-colors-neutral-on-bg;
```
mfranzke, line 11:
```suggestion
	color: inherit;
```

In most cases, I could think about that this would mean a color inheritance instead.

packages/components/src/components/select/select.scss
nmerget: Why shouldn't it work?
It's working at least for chrome :O

mfranzke: probably only on **Sunday** mornings …

mfranzke: I meant those options and their background-colors:
image

packages/components/src/components/link/link.scss
nmerget: https://github.com/db-ui/mono/blob/fix-colors-dark-mode/packages/foundations/scss/helpers/_component.scss#L36

I moved this to a global file and used it for every component. It's the same variable behind, so it should be the same

packages/components/src/components/select/select.scss
nmerget: image

Okay I guess it's an OS thing. Works for windows

nmerget: Maybe the best option for this would be a custom options list. Other frameworks support this as well, a default styled list with buttons and the native one

packages/foundations/scss/icon/_icons.placeholder.scss
nmerget: fixed

packages/foundations/scss/_init.scss
nmerget: fixed

packages/foundations/scss/db-ui-foundations.scss
nmerget: fixed

mfranzke (approved):

PR #1423, opened 2023-08-17
refactor: further adaptions for linting

mfranzke:


.husky/pre-commit
```diff
@@ -5,5 +5,6 @@
 
 npx validate-branch-name
 npm run scripts:check-commit-mail
+# TODO: Those two include redundant lintings
 npx lint-staged
 npm run lint
```
mfranzke, line 10: We need to consolidate some of the lintings that are done here.

nmerget (approved):

PR #1451, opened 2023-08-22
fix: colors for tag to work with dark-mode

nmerget:


packages/components/src/styles/tag-components.scss
```diff
@@ -89,5 +89,10 @@
 				}
 			}
 		}
+
+		&[data-emphasis="strong"] {
+			/* We force white as color for strong tags */
+			color: #fff;
```
mfranzke, line 95: Does this mean "always" as in "both in light and dark mode"? And for all different themes, without the possibility for overwrites?

nmerget: Yes, I don't know if we will support this in the future, but we keep white for now.
We won't get an issue here as well because we force the theme to use colors with a contrast of more than 4.5 which works with white anytime.

PR #1454, opened 2023-08-22
fix: tag colors for emphasis strong

nmerget:


packages/components/src/styles/tag-components.scss
```diff
@@ -16,24 +23,32 @@
```
mfranzke, line 23:
```suggestion
&[data-emphasis] {
	@include overwrite-strong-white();
}
```

Probably I'm misunderstanding something here, but wouldn't this declaration be simpler?

nmerget: But we need this only for [data-emphasis="strong"]

And this color would be overwritten by the following [data-variant="critical"][data-emphasis="strong"]. So I had to included it everytime

mfranzke (approved):

PR #1466, opened 2023-08-23
refactor: icon variants

nmerget:
## Proposed changes

- Refactored how variants change the icon in components
- Removed types from `shared/model.ts` for iconVariants
- Refactored `icon-helper.scss` 

## Types of changes


.github/workflows/03-e2e-merge-reports.yml
```diff
@@ -11,6 +11,7 @@ jobs:
   merge-playwright-reports:
     name: 🎭 Merge Playwright Reports
     runs-on: ubuntu-latest
+    if: always()
```
annsch, line 14: if always? why do wen need this ;)

packages/components/src/components/input/input.lite.tsx
```diff
@@ -149,6 +135,7 @@ export default function DBInput(props: DBInputProps) {
 				minLength={props.minLength}
 				max={props.max}
 				min={props.min}
+				readOnly={props.readonly}
```
annsch, line 138: shouldn't this be lowercase? https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/readonly

nmerget: JSX is using it with camelCase all the other frameworks will render it in lowercase anyways

.github/workflows/03-e2e-merge-reports.yml
nmerget: Good point I moved it to another branch: https://github.com/db-ui/mono/pull/1475

packages/components/src/components/input/input.lite.tsx
mfranzke: I'm not sure whether question and answer match – the attribute `readOnly` that @annsch mentioned here renders in HTML to `read-only`, doesn't it? So this is not necessary that much about lowercase vs. uppercase in the resulting rendering, but about the expectation for whether it's expected to get rendered as `readonly` or `read-only` in the resulting code, isn't it?

nmerget: Why should it be `read-only`? :O

nmerget: If I use `readonly` the typescript compiler for React is throwing an error. But it works in angular an vue because the dom will make them lowercase anyways

packages/components/src/components/alert/alert.scss
```diff
@@ -64,6 +66,10 @@
 	// TODO: probably extract this to an overwrite or external file
 	text-align: start; // workaround for power apps
 
+	@include icons.no-icon() {
```
annsch, line 69: should we rename this to has-no-icon?

packages/components/src/components/button/button.lite.tsx
```diff
@@ -9,10 +9,6 @@ useMetadata({
 		includeIcon: true,
 		hasDisabledProp: true,
 		hasOnClick: true,
-		canvasSize: {
-			height: 'fixed', // 'fixed', 'controlled'
-			width: 'dynamic' // 'fixed', 'dynamic' (requires width property), 'controlled'
-		},
```
annsch, line 15: same error

packages/components/src/components/button/button.lite.tsx
```diff
@@ -45,9 +41,7 @@ useMetadata({
 			{ name: 'noText', type: 'TwoOptions' },
 			{
 				name: 'width',
-				powerAppsName: 'autoWidth', // width property is reserved in power apps
```
annsch, line 48: same

packages/components/src/components/button/button.lite.tsx
```diff
@@ -45,9 +41,7 @@ useMetadata({
...
 				type: 'Enum',
-				defaultValue: 'auto',
```
annsch, line 50: same

packages/components/src/shared/model.ts
```diff
@@ -148,6 +147,19 @@ export type FormCheckProps = {
...
+	 * The disabled attribute can be set to keep a user from edit on the form element
+	 */
+	readonly?: boolean;
+};
+
+export type FormMessageProps = {
+	message?: string;
+	messageIcon?: IconTypes;
+	messageWrap?: string;
+};
```
annsch, line 161: add decscriptional comments for props

packages/components/src/styles/form-components.scss
```diff
@@ -1,14 +1,102 @@
...
+			max-inline-size: 25ch;
+			text-overflow: ellipsis;
+			white-space: nowrap;
+			overflow: hidden;
+		}
+	}
+}
+
+%disabled-form-element {
+	// TODO: Evaluate whether those could get moved to ../_form-elements.scss
```
annsch, line 46: can be removed

packages/components/src/components/alert/alert.lite.tsx
```diff
@@ -13,10 +11,6 @@ useMetadata({
 		// MS Power Apps
 		includeIcon: true,
 		hasOnClick: true,
-		canvasSize: {
-			height: 'fixed', // 'fixed', 'controlled'
-			width: 'controlled' // 'fixed', 'dynamic' (requires width property), 'controlled'
-		},
```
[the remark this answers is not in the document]
nmerget: fixed

packages/components/src/components/button/button.lite.tsx
nmerget: fixed

packages/components/src/components/alert/alert.scss
nmerget: fixed

packages/components/src/components/button/button.lite.tsx
nmerget: fixed

packages/components/src/components/button/button.lite.tsx
nmerget: fixed

packages/components/src/shared/model.ts
```diff
@@ -148,6 +147,19 @@ export type FormCheckProps = {
...
 
+export type FormReadonlyProps = {
+	/**
+	 * The disabled attribute can be set to keep a user from edit on the form element
+	 */
+	readonly?: boolean;
+};
+
+export type FormMessageProps = {
+	message?: string;
```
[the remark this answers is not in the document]
nmerget: fixed

packages/components/src/shared/model.ts
nmerget: fixed

annsch (approved):

annsch (approved):

PR #1477, opened 2023-08-23
fix(Power Apps): font-size overwrite to the MS UI via our global selector

mfranzke:
Resolves https://github.com/db-ui/mono/issues/1473


packages/foundations/scss/_tonality.scss
```diff
@@ -12,7 +12,7 @@ body {
 	@extend %db-component-variables-md;
 }
 
-* {
+:where([class^="db-"], [class^="db-"] *) {
```
nmerget, line 15: Would this selector work on normal tags Like <span>, <p>? Inside our regular web apps?

Moreover this will only add css variables as font-size and line-height. The css variables are set via `:root` in `foundations.scss`.
Maybe we could change the selector there with a `:root:not(:has("some-ms-selector"))`?

nmerget: This would be my suggestion.

`:root {
	@extend %db-bg-neutral;

	&:not(:has(.app-canvas)) {
		@extend %db-ui-regular;
	}
}

.app-canvas {
	@extend %db-ui-regular;
}`

`:where([class^="db-"], [class^="db-"] *) {` only works when you use `<DBPage>`, without it only the db-* Komponents work

mfranzke: I think, scoping it to db-* would be fine in general, but this ticket is more about handling other contexts of our work, like in this case being embedded into MS Power Apps.

So I like your suggestion, let's go with this. Could you please commit your proposal, so that I only need to approve it ?

nmerget: It's your branch I need to approve it :P

nmerget (approved):

mfranzke:
@nmerget could you please create a new release afterwards and upload it to MS platform as well ?

nmerget:
> @nmerget could you please create a new release afterwards and upload it to MS platform as well ?

I would wait for this one as well https://github.com/db-ui/mono/pull/1479

PR #1488, opened 2023-08-25
fix: issue with mobile header wrong padding and not hidden on desktop

nmerget:
## Proposed changes

The mobile header had some issues:

- Padding was not applied correctly because of missing css-vars
- The mobile header had only `display: none`

## Types of changes


packages/components/src/components/header/header.scss
```diff
@@ -184,14 +184,17 @@
 
 [data-hide-on="mobile"] {
 	display: none;
+	visibility: hidden;
```
annsch, line 187: why do we need both here? Aren't this two different types of hiding elements? 
_display:none turns off the layout of the elements, so they are not rendered. visibility:hidden hides the elements without changing their layouts_

packages/components/src/components/header/header.scss
```diff
@@ -184,14 +184,17 @@
...
 	@include screen("m") {
 		display: inherit;
+		visibility: visible;
 	}
 }
 
 [data-hide-on="desktop"] {
 	@include screen("m") {
 		display: none;
+		visibility: hidden;
```
annsch, line 198: same here

packages/components/src/components/header/header.scss
nmerget: You are right.
My intention was to hide the mobile navigation because the slots and the ids of the components inside the slots are duplicated. But we won't be able to fix this issue without JS. Maybe we can rework the header in the future for now I get 2 elements when I use some id in the DOM/ e2e-Tests

packages/components/src/components/header/header.scss
nmerget: fixed

mfranzke (approved):

PR #1489, opened 2023-08-25
refactor: possible simplification

mfranzke:
## Proposed changes

Code simplification.
Even also slightly optimized output of fallback `Math` function, as the first two characters would be `0.` 

## Types of changes


packages/components/src/utils/index.ts
```diff
@@ -1,23 +1,7 @@
...
-						.join('-');
-				}
-			}
-		}
-	} catch (error) {
-		console.warn(error);
-	}
-
-	return Math.random().toString();
+	return window?.crypto?.randomUUID() ?? window?.crypto?.getRandomValues(new Uint32Array(3))?.join('-') ?? Math.random().toString();
```
nmerget, line 4: This wouldn't work. 

In cypress, where I found this issue, is a `window.crypto` but there is not function `randomUUID` which would cause the problem again.

Moreover with the try/catch it's really bullet proof

mfranzke: I don't like try/catch (for whatever reason).

mfranzke: what for do we use cypress?

mfranzke: > In cypress, where I found this issue, is a window.crypto but there is not function randomUUID which would cause the problem again.

You're right, I was too sloppy for this one …
I've just provided a new suggestion.

annsch (approved):

PR #1497, opened 2023-08-28
feat: add css nano with postcss

nmerget:
## Proposed changes

Add `cssnano` via `postcss` to minify CSS output

## Types of changes


packages/components/package.json
```diff
@@ -16,23 +16,24 @@
...
-    "build:mv-styles": "cpr src build -o -f \"(.ts|.tsx|.md)$\"",
-    "build:sass": "sass build --no-source-map --load-path=node_modules/ --load-path=../../node_modules/ --future-deprecation=import",
-    "build:z-directives": "cpr overrides/angular ../../output/angular -o",
-    "build:z-post": "node scripts/post-build/index.js",
-    "build:z2-docs": "react-docgen ../../output/react/src/components/**/*.tsx -o ../../output/docs.json -i *.spec.tsx",
-    "compile:angular": "mitosis build -c mitosis-angular.config.js && node -e \"require('./scripts/post-build/angular.js')(true)\" && npm run build:z-directives && cpr ../../output/tmp/angular/src ../../output/angular/src -o",
+    "build:01_mitosis": "mitosis build",
+    "build:02_mv-styles": "cpr src build -o -f \"(.ts|.tsx|.md)$\"",
+    "build:03_sass": "sass build --no-source-map --load-path=node_modules/ --load-path=../../node_modules/ --future-deprecation=import",
+    "build:04_postcss": "postcss build/**/*.css --replace",
```
mfranzke, line 22: especially those two dependend on each other and couldn't get run in parallel, but only sequently. What about moving `postcss` into a new `scripts`-"namespace", like e.g. `optimize:postcss`? And/or are all of the other `build` scripts even also not being possible to get run in parallel?

mfranzke (approved):

packages/components/package.json
nmerget: fixed

mfranzke (dismissed):

packages/foundations/postcss.config.js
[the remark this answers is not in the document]
nmerget: Had to revert it, because it clashes with our nextjs showcase. There is postcss with a default config...

mfranzke (approved):

PR #1501, opened 2023-08-29
test: clean icons

nmerget:
## Proposed changes

Test icons with cleaned path

## Types of changes


mfranzke:
what do you mean by "cleaned path"? How have you cleaned them up? Is this even already an automated process? And why does the cleanup result in actually _more_ file contents?

image

PR #1508, opened 2023-08-30
fix: missing input types

nmerget:
## Proposed changes

- Add missing types for input
- Add step for input types with date/time

## Types of changes

### DIFF
--- a/packages/components/src/components/input/model.ts
+++ b/packages/components/src/components/input/model.ts
@@ -22,18 +22,24 @@ export type DBInputDefaultProps = {
 	dataList?: KeyValueType[];
 	dataListId?: string;
 	type?:
-		| 'text'
-		| 'search'
-		| 'number'
-		| 'tel'
-		| 'url'
-		| 'email'
-		| 'password'
-		| 'hidden'
+		| 'color'
 		| 'date'
 		| 'datetime-local'
+		| 'email'
+		| 'file' // TODO: move this to own component
+		| 'hidden'
+		| 'month'
+		| 'number'
+		| 'password'
+		| 'range' // TODO: move this to own component
+		| 'search'
+		| 'tel'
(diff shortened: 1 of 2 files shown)


packages/components/src/components/input/model.ts
```diff
@@ -22,18 +22,24 @@ export type DBInputDefaultProps = {
...
+		| 'hidden'
+		| 'month'
 		| 'number'
+		| 'password'
+		| 'range' // TODO: move this to own component
+		| 'search'
 		| 'tel'
+		| 'text'
+		| 'time'
 		| 'url'
```
annsch, line 39: do we really want to provide input for all those types?

mfranzke: why not ?

packages/components/src/components/input/model.ts
```diff
@@ -22,18 +22,24 @@ export type DBInputDefaultProps = {
...
 	dataListId?: string;
 	type?:
-		| 'text'
-		| 'search'
+		| 'color'
+		| 'date'
+		| 'datetime-local'
+		| 'email'
+		| 'file' // TODO: move this to own component
+		| 'hidden'
```
mfranzke, line 30: I don't necessarily see a reason this for this one, except consistency. `input[type=hidden]` wouldn't be a UI component, but only a visually hidden HTML element.

What do you think?

mfranzke (approved):

packages/components/src/components/input/model.ts
nmerget: We need to handle all those properly in the future.
But for now a migration wouldn't be possible if we don't provide them :(

mfranzke: @annsch what was the reason for your question on this topic?

PR #1532, opened 2023-09-06
feat: select

mfranzke:
## Proposed changes

Resolves https://github.com/db-ui/mono/issues/1365

## Types of changes


showcases/angular-showcase/src/app/components/form/form.component.html
mfranzke, line 1: this doesn't work so far, but we'll put this into a follow up story.

mfranzke: https://github.com/db-ui/mono/issues/1537

annsch:
@mfranzke it looks like the asterisk has a bit too much space:

Bildschirmfoto 2023-09-08 um 09 43 27

mfranzke:
> @mfranzke it looks like the asterisk has a bit too much space:
> 
> Bildschirmfoto 2023-09-08 um 09 43 27

I'll cover this within https://github.com/db-ui/mono/pull/1545, as we should generalize it through all the different form elements.

PR #1533, opened 2023-09-06
refactor: scripts

mfranzke:
## Proposed changes

We should further differentiate the scripts we're providing:
- Test screenshots should be possible to get generated more specifically, either easily and simply or by doing a rebuild after facing problems.
- `clean` should do a `git clean` again, as the current simplified one does lead to problems on a quite frequent basis.

## Types of changes


package.json
```diff
@@ -37,12 +37,13 @@
...
-    "clean:patternhub-pages": "rimraf -g showcases/patternhub/pages/components/*/*",
-    "clean:patternhub-src-ts": "rimraf -g showcases/patternhub/components/src/**/*.ts",
-    "clean:patternhub-src-tsx": "rimraf -g showcases/patternhub/components/src/**/*.tsx",
+    "clean": "git clean -dfx --exclude=.env",
+    "clean:incrementally": "npm-run-all -p clean:incrementally:*",
+    "clean:incrementally:components-build": "rimraf packages/components/build",
+    "clean:incrementally:outputs": "rimraf -g output/**/src",
+    "clean:incrementally:patternhub-pages": "rimraf -g showcases/patternhub/pages/components/*/*",
+    "clean:incrementally:patternhub-src-ts": "rimraf -g showcases/patternhub/components/src/**/*.ts",
+    "clean:incrementally:patternhub-src-tsx": "rimraf -g showcases/patternhub/components/src/**/*.tsx",
```
annsch, line 46: For me as a user it's not clear, why we do have all those clean scripts. Should we explain them anywhere, e.g. a README?
Esp. the difference between the patternhub clean scripts

annsch, line 46: For me as a user it's not clear, why we do have all those clean scripts. Should we explain them anywhere, e.g. a README?
Esp. the difference between the patternhub clean scripts

package.json
```diff
@@ -59,7 +60,8 @@
     "lint:xo": "cross-env TIMING=1 NODE_OPTIONS=\"--max-old-space-size=4096\" xo",
     "postbuild:02_components": "cpr output/power-apps build-power-apps -o",
     "prepare": "node -e \"try { require('husky').install() } catch (e) {}\"",
-    "regenerate:screenshots": "git clean -dfx --exclude=.env && npm i && npm run build && npm run build-showcases:react && docker-compose -f ./e2e/docker-compose.regenerate.yml up --build",
+    "regenerate:screenshots": "npm run build && npm run build-showcases:react && docker-compose -f ./e2e/docker-compose.regenerate.yml up",
+    "regenerate:screenshots:rebuild": "npm run clean && npm i && npm run regenerate:screenshots -- --build",
```
annsch, line 64: LGTM

package.json
mfranzke: This would need to get clarified with Nico. I'll open another issue for this clarification.

mfranzke:
@annsch, as it looks good to you, would you please also leave your approval here?

annsch (approved):

PR #1540, opened 2023-09-07
fix(input): search type floating label

mfranzke:
## Proposed changes

Resolves https://github.com/db-ui/mono/issues/1539

## Types of changes


packages/components/src/components/input/input.scss
```diff
@@ -290,13 +290,6 @@
 				@include icon("search");
 			}
 
-			+ label {
-				transform: translate(
-					$paddingLeft,
-					calc(-1 * var(--db-sizing-md))
-				);
-			}
-
```
mfranzke, line 299: @annsch was there a specific reason why search would have needed to behave differently than the other text-based ones ?

annsch: no, this seems to be old stuff that hasn't been deleted after some refactoring according dbInput.

dkolba (approved):

PR #1545, opened 2023-09-08
refactor: form elements

mfranzke:
## Proposed changes

The form elements needed to get some refactoring regarding the following aspects:
- validation
- Label above (default) vs. floating labels
- displaying the different types in our documentation
- further showcase examples
- …

## Types of changes

## ToDos

- [ ] Remove https://github.com/db-ui/mono/pull/1609/files#diff-cdcd29b5b468be8ed75b9d9c490928e7d3e126141256cc5cfc31caab9bed60f3R8-R28


showcases/vue-showcase/src/components/textarea/Textarea.vue
```diff
@@ -1,7 +1,7 @@
 <script setup lang="ts">
 import DefaultComponent from "../DefaultComponent.vue";
 import defaultComponentVariants from "../../../../shared/textarea.json";
-import { DBTextarea } from "../../../../../output/vue/vue3/src";
+import { DBInput, DBTextarea } from "../../../../../output/vue/vue3/src";
```
mfranzke, line 4:
```suggestion
import { DBTextarea } from "../../../../../output/vue/vue3/src";
```

Maybe I'm missing something here, but is this a leftover?

packages/components/src/components/textarea/textarea.scss
```diff
@@ -22,42 +23,113 @@
...
+	&::-webkit-scrollbar-button:single-button {
+		@include color-variants.bg-transparent-interactive(false);
+		background-size: $scrollbar-width;
+		background-repeat: no-repeat;
+		background-position: center;
+		block-size: form-components.$floating-label-size;
+	}
+	/* Up */
+	&::-webkit-scrollbar-button:single-button:vertical:decrement {
+		background-image: url("data:image/svg+xml;base64,PHN2ZyBpZD0iaWNvbiIgaGVpZ2h0PSIxNiIgdmlld0JveD0iMCAwIDE2IDE2IiB3aWR0aD0iMTYiIHhtbG5zPSJodHRwOi8vd3d3LnczLm9yZy8yMDAwL3N2ZyI+PHBhdGggZD0ibTEwLjQ3MiA4LjY5Ni0zLjUzMiAzLjUzNGEuOTk4Ljk5OCAwIDEgMSAtMS40MTMtMS40MTNsMi44MjUtMi44MjUtMi44MjUtMi44MjVhLjk5OC45OTggMCAxIDEgMS40MTMtMS40MTNsMy41MzIgMy41MzJhLjk5OS45OTkgMCAwIDEgMCAxLjQxMnoiIGZpbGw9InZhcigtLWRiLWljb24tY29sb3IsIGN1cnJlbnRDb2xvcikiIGZpbGwtcnVsZT0iZXZlbm9kZCIgdHJhbnNmb3JtPSJtYXRyaXgoMCAtMSAtMSAwIDE1Ljk5IDE1Ljk5KSIvPjwvc3ZnPg0K");
```
mfranzke, line 76: what about using some base64-inline-plugins e.g. from PostCSS? Than it would be clearer, which icons we're using in here.

packages/components/src/components/input/input.scss
```diff
@@ -9,329 +9,90 @@
...
-					transition-duration: 0.01s !important;
-				}
-			}
-
-			// Hiding inputs placeholder for floating label
-			&::placeholder {
-				opacity: 0;
-			}
+	&:has(datalist) {
+		@extend %dropdown-icon;
```
milan-w, line 84: I don't know what's specified in the %dropdown-icon mixin and can't find the declaration in this repo, so maybe it's already fixed.  
But right now, when using an input field with a datalist, the dropdown-icon suggests a dropdown action but is not clickable or interactive at all.  
I'd suggest adding `pointer-events: none;` here to enable usage of the underling input field, when clicking on the dropdown icon. But then again, I have no idea how accessible this solution is for keyboard or screen reader navigation.

__snapshots__/textarea/component/mobile-chrome/DBTextarea-should-match-screenshot.png
mfranzke, line 1: Der Text-Kontrast sieht mir auf dem Text/Labels dieses Elements zu gering aus.

nmerget: Based on the Zeplin specs everything is right. The label has a 0.75 opacity, which wasn't the case before

packages/components/src/components/textarea/textarea.scss
mfranzke: https://github.com/db-ui/mono/issues/1725

nmerget (approved):

PR #1566, opened 2023-09-14
refactor(form element): infotext needs to get connected

mfranzke:
## Proposed changes

Resolves https://github.com/db-ui/mono/issues/1564

## Types of changes


packages/components/src/components/input/input.lite.tsx
```diff
@@ -181,7 +182,8 @@ export default function DBInput(props: DBInputProps) {
 				<DBInfotext
 					size="small"
 					variant={props.variant}
-					icon={getMessageIcon(props.variant, props.messageIcon)}>
+					icon={getMessageIcon(props.variant, props.messageIcon)}
+					id={state._id + '-infotext'}>
```
nmerget, line 186: We should have a shared state for the form elements with `state.messageId` and initialize it at `onMount`

nmerget (approved):

PR #1577, opened 2023-09-16
refactor: reactivate commitlint

mfranzke:
## Proposed changes

We should reactivate this linting tool to lint our commit messages. The [previous PR that deactivated `commitlint`](https://github.com/db-ui/mono/pull/969) sadly doesn't state that much about the (exact) reason why it was deactivated previously, so we need to monitor future dependabot updates.
Furthermore adapted the double-dash to divide our parameters (or in this case the related command) by `husky` examples: https://github.com/typicode/husky/blob/main/.husky/commit-msg

## Types of changes


.husky/commit-msg
```diff
@@ -4,4 +4,4 @@
 . "$(dirname -- "$0")/_/husky.sh"
 
 # TODO: There is some peerDependency not working with type:module from commitlint
```
annsch, line 6: could we remove the comment according the failing peer dependency?

nmerget (approved):

PR #1588, opened 2023-09-19
chore: add dev research for tooltip

nmerget:
## Proposed changes

dev research for tooltip

## Types of changes


annsch (approved):

docs/research/tooltip.md
mfranzke, line 1: > There are two ways to use a tooltip:
> 1. As wrapping tag `<tooltip content="xyz"><button>Test</button></tooltip>`
> 2. As property `<button tooltip/title="xyz">Test</button>`

- why not provide both possibilities? The latter would be restricted not to be able to include an arrow.
- A wrapping tag wouldn't be possible to get connected via [aria-describedby](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-describedby to the element that the tooltip gets displayed for, compare to https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tooltip_role

mfranzke: @nmerget, I've opened an issue for my review:
https://github.com/db-ui/mono/issues/1589

PR #1626, opened 2023-09-28
refactor(focus): implement two colored outline

mfranzke:
## Proposed changes

Resolves https://github.com/db-ui/mono/issues/1620

## Types of changes

## ToDos

- [x] 	Replace color values `#309fd1`, `#fff` and `#242629` by their related design tokens
- [x] Go through each and every element to identify possible collisions


packages/foundations/scss/_init.scss
```diff
@@ -73,6 +81,8 @@ iframe {
 			--db-focus-outline-offset,
 			var(--db-focus-outline-size)
 		);
+		box-shadow: 0 0 0
```
nmerget, line 84: We need to add box-shadow to default transition

packages/foundations/scss/_init.scss
```diff
@@ -3,9 +3,9 @@
 
 :root {
 	// TODO: probably move these to the tokens as well
-	--db-focus-outline-offset: 1px;
-	--db-focus-outline-color: #{variables.$db-colors-informational};
-	--db-focus-outline-size: max(2px, 0.08em);
+	--db-focus-outline-offset: 2px;
+	--db-focus-outline-color: #fff;
```
nmerget, line 7: I don't like that we have a fixed color here. We should ask UX which token we should provide, to make it overwriteable with the theme-builder.

packages/foundations/scss/_init.scss
mfranzke: Let's align, as I not fully understand what you mean. But in general this declaration is at least still missing a styling API for either further declarations of box-shadow, and/or modifications from the outside besides our necessary declaration for the "outline" defined as a box-shadow, meaning to have e.g. another css variable in this box-shadow-chain.

packages/foundations/scss/_init.scss
mfranzke: couldn't agree more.

packages/foundations/scss/_init.scss
nmerget: Sorry I meant the [%default-transition](https://github.com/db-ui/mono/blob/main/packages/foundations/scss/helpers/_component.scss#L34) to animate this new focus state

mfranzke: > Sorry I meant the [%default-transition](https://github.com/db-ui/mono/blob/main/packages/foundations/scss/helpers/_component.scss#L34) to animate this new focus state

@nmerget could you please provide a suggestion or just commit the relevant change?

packages/foundations/scss/_init.scss
```diff
@@ -87,15 +91,43 @@ iframe {
...
+			outline
+				var(
+					--db-transition-duration-ultra-fast,
+					#{variables.$db-transition-duration-ultra-fast}
+				),
+			box-shadow
+				var(
+					--db-transition-duration-ultra-fast,
+					#{variables.$db-transition-duration-ultra-fast}
+				);
```
mfranzke, line 125: The `outline` and `box-shadow` stylings on `focus-visible` include some fallback values for their CSS Custom Properties, which are currently including these very same CSS Custom Properties as well, which doesn't make any sense.

mfranzke: https://github.com/db-ui/mono/issues/1784

packages/foundations/scss/_variables.scss
```diff
@@ -57,3 +59,6 @@ $db-transition-functional-straight: #{$db-transition-duration-medium} #{$db-tran
 
 $db-transition-show-straight: var(--db-transition-show-straight);
 $db-transition-hide-straight: var(--db-transition-hide-straight);
+
+/* Border radius */
+$db-border-radius: helpers.to-rem(4);
```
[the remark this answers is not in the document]
mfranzke: > Maybe the best way would be to move the border radius to the theme builder as well and consume it as a css-property

@nmerget I think that you would need to take care of this aspect, wouldn't you?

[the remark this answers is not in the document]
nmerget: https://github.com/db-ui/mono/pull/1790

Yes :)

nmerget (approved):

PR #1649, opened 2023-10-05
fix: use outlined icons as fallback for solid

nmerget:
## Proposed changes

We don't provide all `solid` icons. If a user imports `DBInfotext` this leads to a missing icon which seems like an error. We should deliver the default icons as fallback.

## Types of changes


mfranzke:
This PR would currently let each icon font explode in size (most of them four times bigger than before). I'd like to further understand the problem, especially as outline is the default and in this case I don't get why the missing icons in `solid` variant would be a problem.

nmerget:
You can see it here: https://db-ui.github.io/mono/review/docs-founctions/foundations/icons/overview

The problem is that we don't have every icon with the solid variant. 
We use another `font-familiy` for the solid icons to load the `solid-x.woff2` file but we do not bundle all icons in this. 

For `DBInfotext` we enforce the solid icons. Using infotext with a non-solid icon shows nothing which looks like an error. But we need to provide the full list of icons for the infotext as well. As a workaround, we use the outline icons as a fallback.

The sizes of the `woff2` files should be the same now. We only load 1 file based on the `font-family` * `font-weight` you are using which is ~40kb each. That's the trade-off to ensure the components work as we wish. The best solution would be to (a) always use outline icons for all components and solid only as extension or to create always a solid variant of an icon.

mfranzke:
> For `DBInfotext` we enforce the solid icons. Using infotext with a non-solid icon shows nothing which looks like an error. But we need to provide the full list of icons for the infotext as well. As a workaround, we use the outline icons as a fallback.

I think to remember that we've had another discussion on the infotext component about not having icons enforced to be `solid`. So this PR is obsolete, isn't it?

nmerget:
> > For `DBInfotext` we enforce the solid icons. Using infotext with a non-solid icon shows nothing which looks like an error. But we need to provide the full list of icons for the infotext as well. As a workaround, we use the outline icons as a fallback.
> 
> @nmerget I think to remember that we've had another discussion on the infotext component about not having icons enforced to be `solid`. So this PR is obsolete, isn't it?

No, it isn't. I think anyways it's the best option to show a fallback for solid icons if there isn't one. The only downside is that the woff file is a little bit larger

mfranzke (approved):

PR #1658, opened 2023-10-06
feat: add popover

nmerget:
## Proposed changes

Add `DBPopover` component.

## Types of changes

## Backlog

- [x] `border-weak` is missing on the element
- [x] The popover (and [tooltip](https://github.com/db-ui/mono/pull/1659)) need to get displayed on `:focus` as well if they are expected to get displayed on `:hover`

https://github.com/db-ui/mono/issues/1732 <- escape key


packages/components/src/components/popover/popover.lite.tsx
```diff
@@ -0,0 +1,53 @@
...
+	onMount(() => {
+		state._id = props.id || 'popover-' + uuid();
+		if (props.stylePath) {
+			state.stylePath = props.stylePath;
+		}
+	});
+	// jscpd:ignore-end
+
+	return (
+		<div
```
mfranzke, line 35:
```suggestion
		<span
```

`div` isn't possible to get used at all places / contexts, so we should switch to an inline-element like e.g. `span`.

packages/components/src/components/popover/popover.lite.tsx
```diff
@@ -0,0 +1,53 @@
...
+			data-open={props.open}
+			data-delay={props.delay}
+			data-width={props.width}
+			data-placement={props.placement}
+			onClick={(event) => state.handleClick(event)}>
+			<Show when={state.stylePath}>
+				<link rel="stylesheet" href={state.stylePath} />
+			</Show>
+			{props.children}
+		</div>
```
mfranzke, line 51:
```suggestion
		</span>
```


`div` isn't possible to get used at all places / contexts, so we should switch to an inline-element like e.g. `span`.

packages/components/src/components/popover/model.ts
```diff
@@ -0,0 +1,38 @@
...
+	 */
+	delay?: 'none' | 'slow' | 'fast';
+	/**
+	 * Disable animation with 'plain' behaviour
+	 */
+	behaviour?: 'animated' | 'plain';
+	/**
+	 * Use open to disable the default hover to use it on click or other trigger.
+	 */
+	open?: boolean;
```
mfranzke, line 22: I'm struggling with the naming of these parameters; they might have been adapted out of other contexts in our system, but especially `open` even already got a meaning in the HTML standard itself nowadays, like e.g. on [`details`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details#attributes) and [`dialog`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog#attributes) elements meaning to progressively control their `open` state by having them displayed/opened initially.

On the other hand, `behaviour` currently "only" controls a configuration similar to `boolean` for specifically for animations.

At first I thought that the "behaviour" of the tooltip (`hover`&`focus` or `click`) would get configured via the property named like this.

So what about moving the configuration option `open` to `behaviour` and the current `behaviour` configuration property to a property called like e.g. `animation` instead? Seems to be more clear to me than.

packages/components/src/components/popover/popover.lite.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+	return (
+		<div
+			ref={component}
+			id={state._id}
+			class={cls('db-popover', props.className)}
+			data-spacing={props.spacing}
+			data-gap={props.gap}
+			data-behaviour={props.behaviour}
+			data-open={props.open}
```
mfranzke, line 42: Additionally (or instead) we should add `role="tooltip"` in case that we have the `:hover`&`:focus` behaviour for this component.

packages/components/src/components/popover/model.ts
nmerget: The suggestion for `animation` is fine, I used the naming from Zeplin, but you are absolutely right here.

Is there a valid option to open a popover on focus?

We still need the `open` for `behaviour="click"` otherwise the user cannot control the popover from the outside or we would need to use `:focus-within` which could lead to other issues

packages/components/src/components/popover/popover.lite.tsx
nmerget: But even the new [popover-target](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/popover) is using a div :O

Can you tell me in detail your concern about a good old div?

packages/components/src/components/popover/popover.lite.tsx
mfranzke: `div` is not the first choice, but the last option when choosing an element; any other element that provides semantic meaning and/or glorifies the diversity of the Hypertext Markup Language should be preferred instead of `div`.

In this special case it's about that `div` couldn't get used in any context, as it's a block-level-element, so it wouldn't be allowed in e.g. `button, whereas `span`, `i` etc. would as they are `inline`-elements.

packages/components/src/components/popover/popover.lite.tsx
mfranzke: `div` is not the first choice, but the last option when choosing an element; any other element that provides semantic meaning and/or glorifies the diversity of the Hypertext Markup Language should be preferred instead of `div`.

In this special case it's about that `div` couldn't get used in any context, as it's a block-level-element, so it wouldn't be allowed in e.g. `button, whereas `span`, `i` etc. would as they are `inline`-elements.

packages/components/src/components/popover/model.ts
nmerget: fixed

packages/components/src/components/popover/popover.lite.tsx
nmerget: fixed

packages/components/src/components/navigation-item/navigation-item.lite.tsx
```diff
@@ -105,6 +105,7 @@ export default function DBNavigationItem(props: DBNavigationItemProps) {
 					{props.children}
 				</button>
 
+				{/* TODO: Consider using popover here */}
```
mfranzke, line 108: @nmerget should we create a follow up issue out of this?

nmerget: https://en.wikipedia.org/wiki/Eating_your_own_dog_food

nmerget: I think we should open up a new issue, to consider it. But I guess we could use it here

nmerget: https://github.com/db-ui/mono/issues/1733

packages/foundations/scss/animation/_animations.scss
```diff
@@ -89,3 +89,51 @@
 		transform: translateY(0%);
 	}
 }
+
+@keyframes popover-to-bottom {
+	0% {
+		opacity: 0;
+		transform: translateY(-25%);
```
mfranzke, line 96: what about using a CSS Variable to differentiate on the only differentiating part in here (`translateY(-25%)`)?

mfranzke (approved):

packages/foundations/scss/animation/_animations.scss
mfranzke: @nmerget I knew that you would love it.

mfranzke (approved):

PR #1659, opened 2023-10-06
feat: add tooltip component

nmerget:
## Proposed changes

Add `DBTooltip` component.

## Types of changes

## Backlog

- [ ] Connecting tooltip and the element it describes by [`aria-describedby`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-describedby)


packages/components/src/components/tooltip/tooltip.lite.tsx
```diff
@@ -0,0 +1,53 @@
...
+		state._id = props.id || 'tooltip-' + uuid();
+		if (props.stylePath) {
+			state.stylePath = props.stylePath;
+		}
+	});
+	// jscpd:ignore-end
+
+	// TODO: Shall we check if only <span>, <p> or direct text was passed as children?
+	return (
+		<div
```
mfranzke, line 36:
```suggestion
		<span
```


`div` isn't possible to get used at all places / contexts, so we should switch to an inline-element like e.g. `span`.

packages/components/src/components/tooltip/tooltip.lite.tsx
```diff
@@ -0,0 +1,53 @@
...
+			data-delay={props.delay}
+			data-width={props.width}
+			data-placement={props.placement}
+			data-gap="true"
+			onClick={(event) => state.handleClick(event)}>
+			<Show when={state.stylePath}>
+				<link rel="stylesheet" href={state.stylePath} />
+			</Show>
+			{props.children}
+		</div>
```
mfranzke, line 51:
```suggestion
		</span>
```

`div` isn't possible to get used at all places / contexts, so we should switch to an inline-element like e.g. `span`.

packages/components/src/styles/db-ui-components.scss
```diff
@@ -111,3 +118,12 @@ dbalert {
 		@extend %show-popover;
 	}
 }
+
+:has(> db-tooltip > .db-tooltip):hover,
+:has(> dbtooltip > .db-tooltip):hover,
+:has(> .db-tooltip):hover {
```
mfranzke, line 124:
```suggestion
:has(.db-tooltip):hover {
```

packages/components/src/shared/model.ts
```diff
@@ -115,6 +115,21 @@ export type WidthProps = {
...
+	 */
+	delay?: 'none' | 'slow' | 'fast';
+	/**
+	 * Disable animation with 'plain' behaviour
+	 */
+	behaviour?: 'animated' | 'plain';
+	/**
+	 * Use fixed with for default max-width
+	 */
+	width?: 'auto' | 'fixed';
```
mfranzke, line 130: We might want to provide a number additionally or instead of `fixed` for devs to easily overwrite the default 200px max-width.

packages/components/src/components/tooltip/tooltip.lite.tsx
```diff
@@ -0,0 +1,55 @@
...
+			className={cls('db-tooltip', props.className)}
+			id={state._id}
+			data-emphasis={props.emphasis}
+			data-behaviour={props.behaviour}
+			data-delay={props.delay}
+			data-width={props.width}
+			data-variant={props.variant}
+			data-placement={props.placement}
+			// TODO: clarify this attribute and we need to set it statically
+			data-gap="true"
```
mfranzke, line 47: Let's align on this attribute and why we need to set it this way statically.

mfranzke (approved):

packages/components/src/shared/model.ts
mfranzke: https://github.com/db-ui/mono/issues/1710

packages/components/src/styles/db-ui-components.scss
mfranzke: https://github.com/db-ui/mono/issues/1709

PR #1763, opened 2023-11-01
feat: add additional popover placements

nmerget:
## Proposed changes

Add additional placements for popover and tooltip.

## Types of changes

### DIFF
--- a/packages/components/src/components/popover/popover.spec.tsx
+++ b/packages/components/src/components/popover/popover.spec.tsx
@@ -31,3 +31,14 @@ test.describe('DBPopover', () => {
 	test.use({ viewport: DEFAULT_VIEWPORT });
 	testComponent();
 });
+
+test.describe('DBPopover', () => {
+	test('should not have any A11y issues', async ({ page, mount }) => {
+		await mount(comp);
+		const accessibilityScanResults = await new AxeBuilder({ page })
+			.include('.db-popover')
+			.analyze();
+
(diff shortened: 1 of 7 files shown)


packages/components/src/components/popover/popover.spec.tsx
```diff
@@ -31,14 +31,3 @@ test.describe('DBPopover', () => {
...
-test.describe('DBPopover', () => {
-	test('should not have any A11y issues', async ({ page, mount }) => {
-		await mount(comp);
-		const accessibilityScanResults = await new AxeBuilder({ page })
-			.include('.db-popover')
-			.analyze();
-
-		expect(accessibilityScanResults.violations).toEqual([]);
-	});
-});
```
mfranzke, line 44: why do you remove this whole part (at least I wouldn't expected some change like this to happen within a PR like this)?

nmerget: I had issues with the local e2e testing.

I guess we had the issue here before as well, at the moment we can remove this test.

nmerget: https://github.com/db-ui/mono/actions/runs/6735460700/job/18308760379?pr=1765

The issue was there before

nmerget: https://github.com/db-ui/mono/pull/1767 <- own PR for this

mfranzke (approved):

PR #1765, opened 2023-11-02
fix: made dark theme optional to prevent breaking apps which are migr…

nmerget:
…ating

## Proposed changes

If a user has "prefer-dark-theme" it may break an app which is migrating, which hasn't a dark theme before. So we should split `db-ui-42` to provide an only light theme styling

## Types of changes


mfranzke:
closed in favour of https://github.com/db-ui/mono/pull/1776

PR #1767, opened 2023-11-02
fix: removed bad a11y issue from popover

nmerget:
## Types of changes

### DIFF
--- a/packages/components/src/components/popover/popover.spec.tsx
+++ b/packages/components/src/components/popover/popover.spec.tsx
@@ -31,14 +31,3 @@ test.describe('DBPopover', () => {
 	test.use({ viewport: DEFAULT_VIEWPORT });
 	testComponent();
 });
-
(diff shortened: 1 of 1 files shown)


mfranzke:
@nmerget this does remove the testing all at once; to my understanding, this issue is actually a false positive, as it's identifying the colors incorrectly, isn't it? So deactivating doesn't necessarily feel like a good / final solution to me; we might have other unidentified issues and would also need to analyze why this is reporting incorrectly, as removing that test could only be a temporary hack.

mfranzke:
… and for whatever reason, even with the test included again, this didn't break anymore: https://github.com/db-ui/mono/pull/1763/checks

mfranzke:
are you still working on this feature in the connected branch? Or is it that you don't follow up on this topic at all?
In case of the latter, you could even also delete the branch via that button in the PR:
image

PR #1798, opened 2023-11-08
fix: issue with form elements new user-valid

nmerget:
## Proposed changes

Goto https://db-ui.github.io/mono/review/main/components/input/overview?page=tonality with the new chrome version and type in 1 input. After focus out it shouldn't be green without setting it required

## Types of changes


packages/components/src/styles/_form-components.scss
```diff
@@ -139,9 +139,9 @@ $floating-label-size: calc(
 		$boolean: "false";
 	}
 
-	&:has(input:not([aria-invalid]):user-#{$key}),
-	&:has(textarea:not([aria-invalid]):user-#{$key}),
-	&:has(select:not([aria-invalid]):user-#{$key}) {
+	&:has(input:not([aria-invalid]):required:user-#{$key}),
+	&:has(textarea:not([aria-invalid]):required:user-#{$key}),
+	&:has(select:not([aria-invalid]):required:user-#{$key}) {
```
mfranzke, line 144: the validity is not just about required, but even also about `pattern`, `min` and further attributes.

nmerget: okay so we need an array with all the properties and use `@each` to handle all of them?

mfranzke: I'm struggling about the expected behaviour on this one. Actually the field is valid so it works correctly, doesn't it? Or is our concept/briefing that it should only get marked as valid in case that there are any validation configurations defined for it?

nmerget: Yes I think it should be only if there is a validation config set up, I gonna ask UX

nmerget: It should be set explicitly, I will try to use the `@each` variant with all properties

nmerget: @mfranzke You have to look at this branch as well. There are some issues here with the other properties like `min` for input and textarea. Furthermore the `user-valid` isn't working for `radio`, `checkbox`, `select`. Maybe there is the same problem for `type=date` etc?

packages/components/src/styles/_form-components.scss
```diff
@@ -125,39 +126,54 @@ $floating-label-size: calc(
...
 		@content;
 	}
 
-	// :user-valid workaround
-	@supports not selector(:user-#{$key}) {
-		&:has(input:not([aria-invalid]):required:#{$key}),
-		&:has(textarea:not([aria-invalid]):required:#{$key}),
-		&:has(select:not([aria-invalid]):required:#{$key}) {
-			@content;
-		}
```
mfranzke, line 154: @nmerget why do you remove this in total? We might still have some time in which we would want to provide this for yet non-supporting browsers.

packages/components/src/styles/_form-components.scss
mfranzke: I've set up a page with all native elements so that we could use this a basis for testing: https://jsbin.com/ruwanaz/edit?html,css,output

mfranzke (dismissed):

PR #1809, opened 2023-11-10
feat: add `labelVariant=hidden` for `input`, `select`, `textarea`

nmerget:
## Proposed changes

- add `labelVariant` Prop for `input`, `select`, `textarea`
- refactored `labelHidden` for `checkbox` and `radio` to match with `labelVariant`

Resolves https://github.com/db-ui/mono/issues/1712

## Types of changes


packages/components/src/shared/model.ts
```diff
@@ -229,14 +229,14 @@ export type FormCheckProps = {
...
-	labelHidden?: boolean;
+	labelVariant?: 'hidden';
 };
 
 export type FormMessageProps = {
 	/**
 	 * Change the variant of the label to float
 	 */
-	labelVariant?: 'above' | 'floating';
+	labelVariant?: 'above' | 'floating' | 'hidden';
```
mfranzke, line 239: ♥

mfranzke, line 239: ♥

packages/components/src/styles/_form-components.scss
```diff
@@ -236,6 +236,13 @@ $floating-label-size: calc(
 		overflow: hidden;
 	}
 
+	&[data-label-variant="hidden"] {
+		label {
+			font-size: 0;
```
mfranzke, line 241: we should avoid `font-size: 0;` as this actually let's this alternative text gets hidden from some screenreaders.

showcases/shared/select.json
```diff
@@ -532,6 +532,24 @@
...
+						{
+							"value": "Option 1"
+						},
+						{
+							"value": "Option 2"
+						}
+					],
+					"label": "Label",
+					"labelVariant": "hidden"
+				}
```
mfranzke, line 552:
```suggestion
```

mfranzke: This should be within the "content" section, shouldn't it ?

mfranzke (approved):

mfranzke (approved):

PR #1826, opened 2023-11-14
refactor: update to Node 20

mfranzke:
## Proposed changes

Resolves https://github.com/db-ui/mono/issues/1823

## Types of changes


mfranzke:
The problem coming up is related to the comment `node --experimental-specifier-resolution=node scripts/generate-docs-mdx.js` being run within the `showcases/patternhub` folder.

showcases/patternhub/scripts/generate-docs-mdx.js
```diff
@@ -1,6 +1,6 @@
-/* eslint-disable unicorn/prefer-top-level-await, no-await-in-loop */
+/* eslint-disable unicorn/prefer-top-level-await, unicorn/prefer-node-protocol, no-await-in-loop */
 
-import FS from 'node:fs';
+import FS from 'fs';
```
mfranzke, line 3: what's the reason for these changes?

nmerget: If I use `esbuild` wiht `--platform=node` I get some warnings about `hooks` and a missmatch between React version. If I use `--platform=browser` it works as expected.
Don't ask me why...
But `xo --fix` would add `node:fs` which would break a "browser" build

mfranzke: > If I use `esbuild` wiht `--platform=node` I get some warnings about `hooks` and a missmatch between React version. If I use `--platform=browser` it works as expected. Don't ask me why... But `xo --fix` would add `node:fs` which would break a "browser" build

Seems to be a Windows topic …

mfranzke: Solved the remaining problem reported by the pipeline through this solution:
https://github.com/evanw/esbuild/issues/1921#issuecomment-1152991694

nmerget (approved):

PR #1829, opened 2023-11-14
feat: upload icon assets via GitHub Actions

nmerget:
## Types of changes


package.json
```diff
@@ -83,6 +83,7 @@
     "@commitlint/cli": "18.4.2",
     "@commitlint/config-conventional": "18.4.2",
     "@playwright/test": "1.35.1",
+    "adm-zip": "0.5.10",
```
mfranzke, line 86: what's the reason to use this zip library instead of [some others](https://www.npmjs.com/search?ranking=optimal&q=zip)? There are other ones with more downloads, more frequent updates and a stable "non-0.x" version, like e.g. [jszip](https://www.npmjs.com/package/jszip) or [archiver](https://www.npmjs.com/package/archiver).

nmerget: I saw it on a GH Action: https://github.com/vimtor/action-zip/blob/master/index.js#L3

I like it, because it is very simple to zip an entire folder and save it directly. 
With the other packages I need to use other dependencies as well to achieve the same goal.

mfranzke (approved):

PR #1832, opened 2023-11-15
fix: lint-stages issue with generate icon fonts

nmerget:
## Proposed changes

Resolves https://github.com/db-ui/mono/issues/1827

## Types of changes


mfranzke (approved):

.lintstagedrc.js
```diff
@@ -1,6 +1,6 @@
 export default {
 	'packages/foundations/assets/icons/functional/**/*.svg': () =>
```
mfranzke, line 2: What's the reason for this function here anyhow?

nmerget: As a function it runs only once. Otherwise the filepath will be added to the command which would trigger the `generate-icon-font` multiple times based on the `.svg`s you changed

PR #1865, opened 2023-11-17
refactor: changed react components to use proper typescript types

nmerget:
## Proposed changes

- add typescript types for components to use them in react
- e.g. enables `DBInput` to use `aria-xxx` or `onMouseEnter` directly

## Types of changes


packages/components/src/components/button/button.lite.tsx
```diff
@@ -1,63 +1,13 @@
...
-			}
-		]
-	}
+	isAttachedToShadowDom: true
 });
 
 export default function DBButton(props: DBButtonProps) {
+	const ref = useRef<HTMLButtonElement>(null);
 	// This is used as forwardRef
 	let component: any;
```
mfranzke, line 12: @nmerget did you leave these in there intentionally ?

packages/components/src/shared/model.ts
```diff
@@ -273,7 +272,6 @@ export type LinkProps = {
 		| undefined;
 	disabled?: boolean;
 	href?: string;
-	hreflang?: string;
```
mfranzke, line 296: @nmerget what's the reason for this removal ?

nmerget: TS said, it is not a valid attribute

nmerget: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a?retiredLocale=de#hreflang

_No built-in functionality._ 

I don't know how to interpret this

showcases/react-showcase/src/components/form/index.tsx
```diff
@@ -87,12 +90,13 @@ const FormComponent = () => {
 							type="date"
 						/>
 						<DBTextarea
+							aria-atomic
```
mfranzke, line 93: what's the reason to introduce this attribute?

packages/components/src/shared/model.ts
mfranzke: Me neither – at least it's included here: https://github.com/search?q=repo%3Amicrosoft%2FTypeScript%20hreflang&type=code

showcases/react-showcase/src/components/form/index.tsx
nmerget: I tested if `aria-*` attributes are passed.

To be honest, `atomic` sounds cool, so I used this for the test :P

mfranzke: totally agreed – but let's use [`aria-label`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label) in this case, as it has more meaning and a clear use case for this element.

mfranzke (approved):

mfranzke (approved):

PR #1866, opened 2023-11-17
fix: we need to actually prettify JS files as well

mfranzke:
## Proposed changes

it's not done by just running `xo`

Resolves https://github.com/db-ui/mono/issues/1847

## Types of changes


.lintstagedrc.js
```diff
@@ -3,6 +3,6 @@ export default {
 		'npm run update:icon-fonts',
 	'*.md': 'markdownlint -c .markdown-lint.yml',
 	'*.{css,scss}': 'stylelint --fix',
-	'*.{js,ts,tsx}': 'xo --fix',
+	'*.{js,ts,tsx}': ['prettier --write', 'xo --fix'],
 	'!*.{js,ts,tsx}': 'prettier --write --ignore-unknown'
```
nmerget, line 7: So this is obsolete, right?

nmerget (approved):

PR #1937, opened 2023-12-01
refactor: brand component

### LINKED ISSUE #1574: Header: why is the logo always supposed to get a link?
Currently there's no way around for the logo within `DB Brand` to have a link, even if none is provided by the property there's still the `href` of `/` being enforced. Is this a conceptual requirement, or could we switch back to having this optional for the developers, depending on their needs and requirements?

nmerget:
## Proposed changes

closes #1574

- reworked brand completely 

## Types of changes


mfranzke:
set to draft as the pipeline fails.

showcases/angular-showcase/src/app/components/brand/brand.component.html
```diff
@@ -0,0 +1,21 @@
...
+	<ng-template
+		#exampleTemplate
+		let-exampleProps="exampleProps"
+		let-exampleName="exampleName"
+		let-exampleIndex="exampleIndex"
+		let-variantIndex="variantIndex"
+	>
+		<db-brand [hideLogo]="exampleProps.hideLogo">
+			<img
+				*ngIf="exampleProps.customLogo"
```
d-koppenhagen, line 15: Should be refactored to use Control Flow here

showcases/angular-showcase/src/app/components/brand/brand.component.ts
```diff
@@ -0,0 +1,10 @@
+import { Component } from '@angular/core';
+import defaultComponentVariants from '../../../../../shared/brand.json';
+
+@Component({
+	selector: 'app-button',
+	templateUrl: './brand.component.html'
```
d-koppenhagen, line 6:
```suggestion
	templateUrl: './brand.component.html',
	standalone: true,
```

showcases/angular-showcase/src/app/app.module.ts
```diff
@@ -60,11 +60,13 @@ import { NavItemComponent } from './nav-item/nav-item.component';
 import { MainNavigationComponent } from './components/main-navigation/main-navigation.component';
 import { TextareaComponent } from './components/textarea/textarea.component';
 import { IconComponent } from './components/icon/icon.component';
+import { BrandComponent } from './components/brand/brand.component';
 
 @NgModule({
```
d-koppenhagen, line 65: NgModule was removed. Component must be imported directly in the parent Standalone Component

showcases/angular-showcase/src/app/components/brand/brand.component.html
nmerget: fixed

showcases/angular-showcase/src/app/components/brand/brand.component.ts
nmerget: fixed

showcases/angular-showcase/src/app/app.module.ts
nmerget: fixed

__snapshots__/brand/showcase/webkit/regular/neutral/DBBrand-should-match-screenshot.png
mfranzke, line 1: is "Custom Logo" with a logo included like the others actually something that our users would understand?

showcases/react-showcase/src/components/brand/index.tsx
```diff
@@ -1,39 +1,22 @@
...
+	hideLogo,
+	customLogo
+}: DBBrandProps & { customLogo: boolean }) => (
+	<DBBrand hideLogo={hideLogo}>
+		{customLogo && (
+			<img
+				src={`${
+					process?.env?.NEXT_PUBLIC_BASE_PATH ?? '/react-showcase'
+				}/assets/images/db_logo.svg`}
+			/>
```
mfranzke, line 18:
```suggestion
				}/assets/images/db_logo.svg`}
				alt=""
			/>
```

nmerget: fixed

showcases/vue-showcase/src/components/brand/Brand.vue
```diff
@@ -0,0 +1,28 @@
+<script setup lang="ts">
+import DefaultComponent from "../DefaultComponent.vue";
+import defaultComponentVariants from "../../../../shared/brand.json";
+import { DBBrand } from "../../../../../output/vue/src";
+import imageUrl from "/assets/images/placeholder.jpg";
+
+const log = (exampleName: string) => {
+	// eslint-disable-next-line no-alert
+	alert(exampleName);
+};
```
mfranzke, line 10: is this a leftover?

packages/components/src/components/brand/brand.scss
```diff
@@ -1,25 +1,23 @@
...
-	a {
-		@extend %brand-display;
-		color: inherit;
-		// Deactivate underline on site name being made clickable by added to the link as well
-		text-decoration: none;
+	&::before {
+		--db-icon-font-size: #{variables.$db-sizing-md};
+		--db-icon-margin-end: 0;
+		--db-current-icon-color: #{colors.$db-brand-origin-enabled};
 	}
```
mfranzke, line 14: wouldn't we still need this if somebody wraps a link around our element (as it's the new approach)?

mfranzke: yes, it seems like we do:
image

showcases/vue-showcase/src/components/brand/Brand.vue
nmerget: yes was a leftover, removed it

packages/components/src/components/brand/brand.scss
nmerget: fixed

packages/components/src/components/brand/brand.scss
```diff
@@ -1,25 +1,33 @@
...
 	font-weight: 700;
 
-	a {
-		@extend %brand-display;
-		color: inherit;
-		// Deactivate underline on site name being made clickable by added to the link as well
-		text-decoration: none;
+	// Deactivate underline on site name being made clickable by added to the link as well
+	text-decoration: none;
+
```
mfranzke, line 19:
```suggestion
```

what for do you need this declaration? It needs to get overwritten on the hyperlink on the outside (only), doesn't it?

packages/components/src/components/brand/brand.scss
```diff
@@ -1,25 +1,33 @@
...
 
-%brand-display {
-	display: flex;
-	gap: variables.$db-spacing-fixed-md;
-	align-items: center;
+:root {
+	* > :has(.db-brand) {
+		// Deactivate underline on site name being made clickable by added to the link as well
+		text-decoration: none;
+	}
```
mfranzke, line 8:
```suggestion
a > :has(.db-brand) {
	// Deactivate underline on site name being made clickable by added to the link as well
	text-decoration: none;
}
```

we could be more restrictive on this one and even also don't need to prefix it by `:root` (or I don't understand the reason).

packages/components/src/components/brand/brand.scss
nmerget: You could add the `<a>` as child into the brand so only the text would be a link. Maybe it's not a use-case, I don't know

mfranzke: > You could add the `<a>` as child into the brand so only the text would be a link. Maybe it's not a use-case, I don't know

valid aspect

mfranzke: I've added `& a` to `a:has(&)` for that.

mfranzke (approved):

PR #1950, opened 2023-12-04
fix: issue with releasing icon assets

nmerget:
## Proposed changes

closes #1948 

## Types of changes

### DIFF
--- a/.github/scripts/release/upload-asset.js
+++ b/.github/scripts/release/upload-asset.js
@@ -11,6 +11,25 @@ const uploadAsset = async ({
 	assetPath
 }) => {
 	const { repo, owner } = context.repo;
+
+	const uploadRelease = await github.rest.repos.getRelease({
+		owner,
+		repo,
+		release_id
+	});
+
+	const foundAsset = uploadRelease?.assets?.find(
+		(asset) => asset.name === assetName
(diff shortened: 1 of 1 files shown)


.github/scripts/release/upload-asset.js
```diff
@@ -11,6 +11,25 @@ const uploadAsset = async ({
...
+		owner,
+		repo,
+		release_id
+	});
+
+	const foundAsset = uploadRelease?.assets?.find(
+		(asset) => asset.name === assetName
+	);
+
+	if (foundAsset) {
```
mfranzke, line 25: As you're only checking for an existing release file with this and the new code before to delete the item afterwards, wouldn't it be possible to delete it directly without the previous check? Or would the deletion without a previous check result in a broken pipeline ?

nmerget: I'm not 100% sure, but calling a DELETE endpoint should result in a non 200 Status code, which would exit the pipeline.
The alternative would be a try catch block, but it would also catch any other error for the endpoint and that isn't what we want. 
The GET + DELETE is the best option I guess, why do you have concerns about it?

mfranzke (approved):

PR #2011, opened 2024-01-02
feat: add anchor headings for patternhub

### LINKED ISSUE #1966: patternhub: anchor links not being generated
Currently anchor links that we use within our documents (e.g. see the list "We currently support" on the page https://db-ui.github.io/mono/review/main/foundations/readme) and for which the anchors are automatically generated by GitHub for displaying markdown files content, aren't being generated on the resulting pages by patternhub, compare to e.g. https://db-ui.github.io/mono/review/main/foundations/readme#scss in this case.

We should align on both how we'd like to include anchors in our documents in general, and even also how to solve this problem specifically.

nmerget:
## Proposed changes

closes #1966 

## Types of changes


mfranzke:
@nmerget this current solution renders a `button` element within the `a`-elements within the headlines. The `button` needs to get removed.

mfranzke:
@nmerget I would have expected that you would need a [`scroll-margin-top` CSS declaration](https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-margin-top) for the anchor to get scrolled to below the fixed positioned header. Some other technique seems to be used here, that might be part of the problem regarding the strangely scrolling header.

showcases/patternhub/components/link-header/index.tsx
```diff
@@ -0,0 +1,19 @@
+import Link from 'next/link';
+import { DBIcon } from '../src';
+
+const LinkHeader = ({ id }) => {
+	if (id) {
+		return (
+			<>
+				<a className="anchor" id={id} />
```
mfranzke, line 8:
```suggestion
```

The `id`-attribute is now included twice in the resulting HTML code:
```html
<h3 id="css">CSS<a class="anchor" id="css"></a><a href="/mono/review/feat-anchor-links/foundations/readme#css"><button class="db-button is-icon-text-replace" data-icon="add_link" data-variant="text">css</button></a></h3>
```

We wouldn't need it as an exclusive `a`-element, but could rely on the `id`-attribute of the headline itself.

showcases/patternhub/styles/globals.scss
```diff
@@ -317,3 +326,32 @@ div[class^="ch-"] {
...
+	}
+}
+
+.db-header {
+	&[data-rerender="true"] {
+		display: none;
+	}
+	&[data-rerender="false"] {
+		display: inherit;
+	}
```
mfranzke, line 356:
```suggestion
	&[data-rerender="false"] {
		display: inherit;
	}
```

do you actually need this one?

showcases/patternhub/styles/globals.scss
```diff
@@ -317,3 +326,32 @@ div[class^="ch-"] {
...
+		margin-block: auto;
+		margin-inline-start: variables.$db-spacing-fixed-xs;
+		visibility: hidden;
+	}
+
+	&:hover {
+		a {
+			visibility: visible;
+		}
+	}
```
mfranzke, line 347: couldn't we make this part of our headline component ?

showcases/patternhub/styles/globals.scss
```diff
@@ -5,6 +5,15 @@
 
 @forward "@db-ui/components/build/styles/db-ui-42-webpack";
 
+html {
+	position: fixed;
+	inset: 0;
+}
```
mfranzke, line 11: let's talk about those declarations, as I don't get their reason and it might be too hard to align on this via comments.

showcases/patternhub/styles/globals.scss
nmerget: Which "headline component" do you mean? I think this is a special case for patternhub only

showcases/patternhub/styles/globals.scss
nmerget: https://stackoverflow.com/questions/65953805/scrollintoview-breaks-the-overflow-scroll/65955689#65955689

This resolves the issue, otherwise, there wouldn't be a solution if you want the scrollbar only on the `<main>` element.

showcases/patternhub/components/link-header/index.tsx
nmerget: I think it was some leftover 👍

showcases/patternhub/styles/globals.scss
nmerget: I think it was some leftover 👍

showcases/patternhub/styles/globals.scss
mfranzke: that was actually another "test". We don't have that headline component. Seems to me like we need one. A component that we provide that actually includes the opportunity to have these anchor-`id`s as well as the links you're including in here. This might be beneficial for any documentation related webpage.

nmerget: I think this wouldn't be part of the core library. Maybe we can add something like this to an extension.

mfranzke (dismissed):

mfranzke (approved):

showcases/patternhub/components/link-header/index.tsx
```diff
@@ -0,0 +1,16 @@
+import Link from 'next/link';
+import { DBIcon } from '../src';
+
+const LinkHeader = ({ id }) => {
+	if (id) {
+		return (
+			<Link href={`#${id}`} scroll={false}>
+				<DBIcon icon="add_link">{id}</DBIcon>
```
mfranzke, line 8:
```suggestion
				<DBIcon icon="add_link"></DBIcon>
```

I don't understand what for what `id` would be included in there.

showcases/patternhub/components/link-header/index.tsx
```diff
@@ -0,0 +1,16 @@
...
+			</Link>
```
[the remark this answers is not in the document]
nmerget: That's the reason why I add `{id}` to DBIcon, but it has `aria-hidden` :(

I guess the solution is to add it to the link and hide it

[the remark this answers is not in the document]
nmerget: fixed

mfranzke (dismissed):

mfranzke (approved):

PR #2016, opened 2024-01-03
docs: add remaining migration guides

### LINKED ISSUE #1936: Some migration guides are missing
- [ ] link
- [ ] input
- [x] brand
- [ ] icon
- [ ] checkbox
- [ ] radio
- [ ] header
- [ ] page
- [ ] main-navigation

nmerget:
## Proposed changes

closes #1936 

## Types of changes


packages/components/src/components/checkbox/docs/Migration.md
```diff
@@ -0,0 +1,18 @@
...
+
+| Before         | Status | After         | Description |
+| -------------- | :----: | ------------- | ----------- |
+| `elm-checkbox` |   🔁   | `db-checkbox` |             |
+
+### id
+
+| Before     | Status | After | Description |
+| ---------- | :----: | ----- | ----------- |
+| `input_id` |   🔁   | `id`  |             |
```
mfranzke, line 18: @nmerget I just recognized that we use the attribute `id` on the outer components element that we actually transfer to the native element – this might lead to that the `id` would be on two elements afterwards (at least for frameworks like e.g. Angular in which context the outer element would still render), wouldn't it ?

packages/components/src/components/icon/docs/Migration.md
```diff
@@ -0,0 +1,18 @@
...
+
+| Before | Status | After     | Description |
+| ------ | :----: | --------- | ----------- |
+|        |   🔁   | `db-icon` |             |
+
+### variant
+
+| Before    | Status | After | Description                                                                                                             |
+| --------- | :----: | ----- | ----------------------------------------------------------------------------------------------------------------------- |
+| `variant` |   🔁   | ❌    | icon variants are controlled by `font-family:var(--db-icon-font-family)` and `font-weight: var(--db-icon-font-weight);` |
```
mfranzke, line 18: I just recognized that we wouldn't provide the `variant`-attribute any more. Wouldn't this still be possible, as it would be beneficial regarding developer convenience?

mfranzke: created https://github.com/db-ui/mono/issues/2049 for this question.

packages/components/src/components/checkbox/docs/Migration.md
mfranzke: created https://github.com/db-ui/mono/issues/2050 for this question.

packages/components/src/components/icon/docs/Migration.md
nmerget: Right, we introduced `data-icon-weight` & `data-icon-variant` we should expose this to `DBIcon`

packages/components/src/components/checkbox/docs/Migration.md
nmerget: image

You are right, that's a problem...
But this is an issue for every component, we need to discuss this to find the best solution

mfranzke: > image
> 
> You are right, that's a problem... But this is an issue for every component, we need to discuss this to find the best solution

https://github.com/db-ui/mono/issues/2017

packages/components/src/components/icon/docs/Migration.md
nmerget: https://github.com/db-ui/mono/pull/2052

mfranzke (approved):

PR #2019, opened 2024-01-03
feat: add autocomplete and caret-color for form elements

### LINKED ISSUE #909: Feat: styling :autofill & caret-color with DB Colors
- https://developer.mozilla.org/en-US/docs/Web/CSS/:autofill
- https://developer.mozilla.org/en-US/docs/Web/CSS/caret-color (we had this colored in red with Guidelines 2)

We should discuss this with our design colleagues first (which color should we use)

nmerget:
## Proposed changes

closes #909

## Types of changes

### DIFF
--- a/packages/components/src/styles/_form-components.scss
+++ b/packages/components/src/styles/_form-components.scss
@@ -321,6 +321,7 @@ $floating-label-size: calc(
 		&[data-variant="#{$name}"] #{$selector},
 		&[data-variant="#{$name}"] [id$="-placeholder"] {
 			color: map.get($colors, "on-bg");
+			caret-color: map.get($colors, "element");
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


mfranzke (approved):

packages/components/src/styles/_form-components.scss
mfranzke, line 1: currently this leads to incorrect color contrasts.

nmerget: How did you check this? It is a semi-transparent background I got some "range". I changed it to "on-bg" color, so this should be valid

mfranzke: I've measured the color of background on pointer devices (like mouse) focused input fields and compared it against the defined caret-color, like e.g. https://coolors.co/contrast-checker/0099d1-eff5f8; please ignore their outcome regarding an expected `4.5:1`, as this color-contrast is only meant for text on background, where we measure object color – caret – to background-color here, which only needs `3:1`.

mfranzke (approved):

PR #2024, opened 2024-01-04
chore: tidy-up scripts

### LINKED ISSUE #1541: npm clean scripts
For me as a user it's not clear, why we do have all those clean scripts. Should we explain them anywhere, e.g. a README?
Esp. the difference between the patternhub clean scripts

_Originally posted by @annsch in https://github.com/db-ui/mono/pull/1533#discussion_r1318328242_

### LINKED ISSUE #1958: docs(e2e): align docs and scripts
Currently within the [documentation](https://github.com/db-ui/mono/blob/main/e2e/README.md) we're mentioning to run `npm run build-showcases`, whereas within the [`scripts`](https://github.com/db-ui/mono/blob/main/package.json#L64) we're only building the react showcase. We should align this and even also also replace the README parts by these npm `scripts` instead of individual other partial scripts.

nmerget:
## Proposed changes

Moved some `scripts` from root `package.json` to `scripts/package.json` to have a better overview.
Most of the scripts are needed because of [npm-run-all](https://www.npmjs.com/package/npm-run-all), as we can't run scripts in parallel on npm workspaces yet.

closes #1541
closes #1958

## Types of changes


e2e/README.md
```diff
@@ -2,7 +2,7 @@
 
 This directory provides `docker-compose.yml` to test or regenerate screenshots.
 
--   run `npm run build && npm run build-showcases` from your root directory
+-   run `npm run build && npm run build --workspace=react-showcase` from your root directory
```
mfranzke, line 5: This is not a direct replacement, as `build-showcases` previously ran `npm-run-all -p build-showcases:*` meaning to build all showcases. `npm run build --workspace=react-showcase` now only builds the react showcase, so is this actually enough for e2e?

nmerget: Yes, we just use one showcase to generate the screenshots, which is the react-showcase.

package.json
```diff
@@ -19,65 +19,27 @@
...
-    "build-outputs:ngx-components": "npm run build --workspace=@db-ui/ngx-components",
-    "build-outputs:react-components": "npm run build --workspace=@db-ui/react-components",
-    "build-outputs:v-components": "npm run build --workspace=@db-ui/v-components",
-    "build-outputs:web-components": "npm run build --workspace=@db-ui/web-components",
-    "build-showcases": "npm-run-all -p build-showcases:*",
-    "build-showcases:angular": "npm run build --workspace=angular-showcase",
-    "build-showcases:patternhub": "npm run build --workspace=db-patternhub",
-    "build-showcases:react": "npm run build --workspace=react-showcase",
-    "build-showcases:vanilla": "npm run build --workspace=vanilla-showcase",
-    "build-showcases:vue": "npm run build --workspace=vue-showcase",
```
mfranzke, line 38: Even though that I totally agree that this huge amount of scripts within the `package.json` are way too many, "just" moving them to a workspace does only partly solve the problem and even also abstract the process that would get executed from the main `package.json`; and I agree on the initial reason (Most of the scripts are needed because of [npm-run-all](https://www.npmjs.com/package/npm-run-all)), but we should probably further experiment with solutions like e.g. [`@jwplayer/npm-run-ws`](https://www.npmjs.com/package/@jwplayer/npm-run-ws) until Node provides concurrent workspace script execution out of the box.

nmerget: The package has 42 downloads a week ^.^

mfranzke: > The package has 42 downloads a week ^.^

we could give it a boost. Or choose an alternative.

nmerget: we need the build script primarily for the pipeline
I don't think we should add a new dependency just because it isn't "nice looking" :S

mfranzke:
@nmerget I'm still struggling with "just" moving our scripts to a sub-package. Out of https://github.com/npm/rfcs/issues/190 I found https://github.com/google/wireit – what do you think about that for some consolidation ?

scripts/dev.js
```diff
@@ -34,7 +34,7 @@ inquirer
 	])
 
 	.then((answers) => {
-		let startCommand = 'npm-run-all -p start:foundations dev:sass';
+		let startCommand = 'npm-run-all -p dev:foundations dev:sass';
```
mfranzke, line 37: this might be misleading, as this `dev:foundations` actually doesn't trigger an `npm run dev`, but `npm run start`. What about with leaving this script name within the `start` namespace ?

package.json
```diff
@@ -19,65 +19,27 @@
...
     "lint:jscpd": "jscpd --exitCode 1 .",
     "lint:markdownlint": "markdownlint -c .markdown-lint.yml **/*.md",
     "lint:stylelint": "stylelint **/*.scss",
     "lint:xo": "cross-env TIMING=1 NODE_OPTIONS=\"--max-old-space-size=4096\" xo",
     "prepare": "node -e \"try { require('husky').install() } catch (e) {}\"",
-    "regenerate:screenshots": "npm run build && npm run build-showcases:react && docker-compose -f ./e2e/docker-compose.regenerate.yml up",
+    "regenerate:screenshots": "npm run build && npm run build --workspace=react-showcase && docker-compose -f ./e2e/docker-compose.regenerate.yml up",
     "regenerate:screenshots:rebuild": "npm run clean && npm i && npm run regenerate:screenshots -- --build",
-    "scripts:check-commit-mail": "node scripts/check-commit-mail.js",
-    "scripts:md-resolve-svg": "node scripts/md-resolve-svg.js",
```
mfranzke, line 67: it seems you've totally removed this `scripts:md-resolve-svg` script, whereas it's still being called within `.github/workflows/01-build-showcases.yml`

mfranzke (dismissed):

scripts/dev.js
nmerget: fixed

mfranzke (approved):

PR #2025, opened 2024-01-04
chore: optimize `normalize` dependencies usage

### LINKED ISSUE #1494: Remove `@csstools/normalize.css` from foundations
- We have a dependency for foundations that [normalizes css](https://github.com/db-ui/mono/blob/main/packages/foundations/package.json#L31) but a lot of those normalizes are overwritten later
- We should remove the dependency and check which of the normalizes matters for us and which ones we should adapt with our own variables (tonality etc.)

nmerget:
## Proposed changes

closes #1494 

## Types of changes


mfranzke:
This is currently just removing normalize, but not replacing it with another solution that would ensure we stay up to date.

packages/foundations/scss/_normalize.scss
mfranzke, line 1: This is currently just removing normalize, but not replacing it with another solution that would ensure we stay up to date.

nmerget: The postcss plugin didn't work, maybe you can check it as well?
I don't like this hard dependency just for some CSS...

nmerget: fixed it

mfranzke (approved):

PR #2040, opened 2024-01-05
refactor(icons): prevent the textual representation from being announced

mfranzke:
by providing an empty alternative text

## Proposed changes

Resolves https://github.com/db-ui/mono/issues/2036

## Types of changes

### DIFF
--- a/packages/foundations/scss/icons/_icon-helpers.scss
+++ b/packages/foundations/scss/icons/_icon-helpers.scss
@@ -61,6 +61,8 @@ $default-icon-font-size: var(--db-icon-font-size, #{$default-icon-size-rem});
 			@if $position == "after" {
 				--db-icon-after: "#{$icon}";
 				content: $default-icon-content-after;
+				content: $default-icon-content-after /
+					""; // Hiding icon from screenreaders, https://www.w3.org/TR/css-content-3/#alt
 				margin-inline-start: $default-icon-margin-start;
 			}
 		}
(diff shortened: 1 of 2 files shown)


nmerget:
Do we need the mixin [here](https://github.com/db-ui/mono/pull/2040/files#diff-ec80ae73206b805aaf8bcae303151372274b5769ade789a0ae9539419ee35450R72) as well? A screen-reader might speak "none" otherwise

mfranzke:
> Do we need the mixin [here](https://github.com/db-ui/mono/pull/2040/files#diff-ec80ae73206b805aaf8bcae303151372274b5769ade789a0ae9539419ee35450R72) as well? A screen-reader might speak "none" otherwise

Have you actually tested this? By the specs, the element wouldn't get rendered:
https://developer.mozilla.org/en-US/docs/Web/CSS/content?retiredLocale=de#none

nmerget:
> > Do we need the mixin [here](https://github.com/db-ui/mono/pull/2040/files#diff-ec80ae73206b805aaf8bcae303151372274b5769ade789a0ae9539419ee35450R72) as well? A screen-reader might speak "none" otherwise
> 
> Have you actually tested this? Judged by the specs, the pseudo-element wouldn't get rendered at all: https://developer.mozilla.org/en-US/docs/Web/CSS/content?retiredLocale=de#none

It was a question. But yes I tested it and it is removed from DOM completely

nmerget (approved):

PR #2044, opened 2024-01-06
refactor: `playwright` update

mfranzke:
## Types of changes


.github/dependabot.yml
```diff
@@ -19,10 +19,6 @@ updates:
       time: "02:00"
       timezone: "Europe/Berlin"
     ignore:
-      # disable playwright auto update for now, because we need to update them manually anyways
```
nmerget, line 22: We can't do this automaticlly, because we have the same dependency for github workflows and docker, which wouldn't be updated by dependabot

mfranzke: > We can't do this automaticlly, because we have the same dependency for github workflows and docker, which wouldn't be updated by dependabot

agreed. couldn't we still automate this via dependabot by using a hook and running the script than during its update process ?

nmerget: This may be possible, but it would take some time to integrate. Do we need to update playwright all the time? We could update it manually whenever we need it

mfranzke: > This may be possible, but it would take some time to integrate. Do we need to update playwright all the time? We could update it manually whenever we need it

this sounds like you would accept a manual process instead of automating *EVERYTHING*. Who are you and what did you do to Nico ?

nmerget: 😱

mfranzke: seems to me like we're doomed.

nmerget: add the auto-update (I hope this works :P)

mfranzke: > add the auto-update (I hope this works :P)

what could possibly go wrong ?

.github/scripts/update-playwright.js
```diff
@@ -0,0 +1,45 @@
...
+				'.github/workflows/02-e2e.yml',
+				'.github/workflows/02-e2e-showcases.yml',
+				'e2e/Dockerfile'
+			]
+		},
+		{
+			from: `@playwright/test": "${oldPlaywrightVersion}"`,
+			to: `@playwright/test": "${version}"`,
+			files: ['e2e/package.json']
+		},
```
mfranzke, line 30: couldn't we handle this as a regular dependabot update instead ?

mfranzke: https://github.com/db-ui/mono/issues/2062

PR #2067, opened 2024-01-10
feat: add tab component

snahn2209:
## Proposed changes

add new component

Resolves https://github.com/db-ui/mono/issues/2093

## Types of changes


packages/components/src/components/tabs/tabs.spec.tsx
```diff
@@ -0,0 +1,67 @@
...
+		await expect(component).toHaveScreenshot();
+	});
+};
+
+test.describe('DBTabs', () => {
+	test.use({ viewport: DEFAULT_VIEWPORT });
+	testComponent();
+});
+
+// TODO: AXE has problems with tabs...
```
mfranzke, line 56: do you want to fix this now or should we open a follow up ticket ?

nmerget: A follow-up...

packages/components/scripts/post-build/angular.js
```diff
@@ -185,6 +185,15 @@ module.exports = (tmp) => {
 			}
 		];
 
+		if (component.config?.angular?.initValues) {
+			component.config?.angular?.initValues.forEach((init) => {
+				replacements.push({
+					from: `["${init.key}"];`,
+					to: `["${init.key}"] = ${init.value === '' ? '""' : init.value};`
```
mfranzke, line 192:
```suggestion
					to: `["${init.key}"] = ${init.value || '""'};`
```

wouldn't this mean the same, but be simpler ?

showcases/shared/tab.json
mfranzke, line 1: I'm struggeling whether it makes sense to provide a documentation for these, as this component wouldn't get used standalone, would it? For Sketch this most likely makes perfectly sense, because of how they create those patterns based on the feasibility of those tools, but is it necessary / relevant for the tech/dev context ?

packages/components/src/shared/model.ts
```diff
@@ -391,6 +395,37 @@ export type CloseEventState = {
...
+	/**
+	 * The name of the tab list, is required for grouping multiple tabs together. Otherwise, content won't switch by clicking the tabs.
+	 */
+	name: string;
+
+	/**
+	 * The index of the tab, is required for grouping multiple tabs together. Otherwise, content won't switch by clicking the tabs.
+	 */
+	index: number;
+};
```
mfranzke, line 415: I'm struggling to have those included as mandatory parameters that the consuming users would need to put into the tab component. At least the name should be something that we could create within the `tabs` component and than provide to the `tab` components on an component instance level.

packages/components/src/components/tabs/tabs.spec.tsx
mfranzke: https://github.com/db-ui/mono/issues/2230

packages/components/src/components/tabs/tabs.lite.tsx
```diff
@@ -0,0 +1,205 @@
...
+
+	onUpdate(() => {
+		if (ref && state.initialized) {
+			const childTabLists = ref.getElementsByClassName('db-tab-list');
+
+			if (childTabLists?.length > 0) {
+				const firstTabList = childTabLists.item(0);
+				if (firstTabList) {
+					firstTabList.setAttribute(
+						'aria-orientation',
```
mfranzke, line 81: ♥️

packages/components/src/components/tabs/tabs.lite.tsx
```diff
@@ -0,0 +1,205 @@
...
+								) {
+									if (
+										(props.initialSelectedIndex ===
+											undefined &&
+											index === 0) ||
+										props.initialSelectedIndex === index
+									) {
+										input[0].click();
+									}
+								}
```
mfranzke, line 128: why not combine these two `if` blocks?

packages/components/src/components/tabs/docs/Vue.md
```diff
@@ -0,0 +1,18 @@
...
+### Use component
+
+```vue App.vue
+<!-- App.vue -->
+<script>
+import { DBTabs } from "@db-ui/v-components";
+</script>
+
+<template>
+	<DBTabs>Tabs</DBTabs>
```
mfranzke, line 14: I know that we've aligned about not exposing too much component schema within these code blocks. But on the other hand `<DBTabs>Tabs</DBTabs>` wouldn't ever be valid code, would it?

packages/components/scripts/post-build/angular.js
nmerget: No, it wouldn't. In your suggestion `init.value` is undefined, but it should be an empty string, because you can use `undefined` as value

packages/components/src/components/tabs/docs/Vue.md
nmerget: All the `.md` files aren't updated at the moment. This would be the next step after merging this branch.

showcases/shared/tab.json
nmerget: Good point, I guess we discuss it and remove it if we don't need it here

packages/components/src/shared/model.ts
nmerget: You are absolutely right :)

I removed the "name"+"index" stuff the tab won't be used standalone without the tabs parent so we let it handle the id stuff

nmerget (approved):

PR #2113, opened 2024-01-19
refactor: fix nav item spacing

snahn2209:
## Proposed changes

in navigation item component: added missing spacing between text and icons

## Types of changes


packages/components/src/components/navigation-item/navigation-item.scss
```diff
@@ -105,6 +105,10 @@
 			@include icons.set-icon("chevron_right", "after");
 
 			font-weight: inherit;
+
+			&::after {
+				margin-inline-start: variables.$db-spacing-fixed-sm;
```
nmerget, line 110: There is the prop `width=full` where we can't set the margin to `sm`, may you check this again?

nmerget (approved):

PR #2114, opened 2024-01-19
docs(@db-ui/ngx-components): change term to content projection instead of slot

### LINKED ISSUE #1498: docs(angular): use term "content projection" instead of "slots" in angular docs
Developers using the angular framework may be confused by the term "slot". This concept is called content projection in angular:
https://angular.io/guide/content-projection

I think it would be better to use the angular vocabulary in the related docs to avoid confusion.

d-koppenhagen:
The term `slot` is typically not used in the angular world. It's called `content projection` here.
see: https://angular.dev/guide/components/content-projection

closes #1498

## Types of changes

### DIFF
--- a/packages/components/src/components/accordion-item/docs/Angular.md
+++ b/packages/components/src/components/accordion-item/docs/Angular.md
@@ -18,7 +18,7 @@ import { DBAccordionItemModule } from '@db-ui/ngx-components';
 
 ### Use component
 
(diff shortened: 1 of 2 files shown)


mfranzke (approved):

mfranzke:
@d-koppenhagen thanks a lot for your contribution !

Issue #237, opened 2023-01-23
feat: provide individual components for angular as `standalone` components


mfranzke:
With the [introduction of Standalone Components with Angular 14](https://angular.io/guide/standalone-components), components can be imported and tree-shaked individually without the need of a wrapping module such as `DBUIElementsModule`.

It would be great if all components can be marked as `standalone` and imported individually.

compare to https://github.com/db-ui/elements/issues/605

annsch:
Actually this feature/option seems to be not fully implemented: 
https://github.com/BuilderIO/mitosis/issues/960
We should observe their solution first.

annsch:
@nmerget according to Dannys Feedback is this really nice to have or a must have?

nmerget:
You are right, it's Angular so we could add the label "not supported" :P

We can remove the label, let's finish the input for Angular and Vue.

d-koppenhagen:
I will give this issue a push since we ware st Angular 17 now. Standalone components are the recommended way in Angular now. Angular will drop or at least deprecate NgModules in the future.

d-koppenhagen:
Another point: It produces less boilerplate code and makes integration and testing easier.

d-koppenhagen:
If you give me a quick into about the mechanisms of mitosis here, I can probably also take over.

d-koppenhagen:
@annsch I think I managed it. Fingers crossed.

PR #2184, opened 2024-02-07
refactor(ngx-components)!: completely switch to standalone

d-koppenhagen:
## Proposed changes

**BREAKING CHANGE:**
All Angular components are `standalone` now.
SCAM (Single Component Angular Modules) have been removed.
please import the standalone components directly in your `imports` section.

```ts app.component.ts
//app.component.ts
import { DBButton } from '@db-ui/ngx-components';

@Component({
  // ...
  imports: [
    // ...,
    DBButton
  ],
  // ...
})
```

closes [#237](https://github.com/db-ui/mono/issues/237)

## Types of changes

## Further comments

This PR includes https://github.com/db-ui/mono/pull/2183 as well and sits on top of it.


packages/components/mitosis.config.js
```diff
@@ -7,7 +7,8 @@ module.exports = {
 			typescript: true
 		},
 		angular: {
-			typescript: true
+			typescript: true,
+			standalone: true
```
nmerget, line 11: There is a `mitosis-angular.config.js` besides this file, which need the `standalone:true` as well

d-koppenhagen:
Not sure why the playwright tests are failing. The want to access the route via HasLocation (e. g. `/#/04/accordion`). This locations strategy works when running the app in dev mode and when building+serving it locally with production build options: `npm run build && npm run preview`.
The strange thing is: when starting playwright locally in UI mode (`--ui`), it seems like an old version of the angular app is started and it uses the "normal" path location strategy. In fact, it always checks against the root route since the child routes are not loaded. Any idea?

showcases/angular-showcase/src/app/components/navigation-item/navigation-item.component.html
```diff
@@ -19,31 +19,33 @@
...
 		>
 			<ng-container *dbNavigationContent>
-				<ng-container *ngIf="exampleProps.areaPopup">{{
-					exampleName
-				}}</ng-container>
-				<a href="#" *ngIf="!exampleProps.areaPopup">{{
-					exampleName
-				}}</a>
+				@if (exampleProps.areaPopup) {
+					{{ exampleName }}
```
nmerget, line 23: Based on the snapshots, the exampleName is missing:

neutral-actual

d-koppenhagen: Ah, I see. Now I also managed to run it locally (I forgot to re-build the outputs). I am on it

nmerget:
> Not sure why the playwright tests are failing. The want to access the route via HasLocation (e. g. `/#/04/accordion`). This locations strategy works when running the app in dev mode and when building+serving it locally with production build options: `npm run build && npm run preview`. The strange thing is: when starting playwright locally in UI mode (`--ui`), it seems like an old version of the angular app is started and it uses the "normal" path location strategy. In fact, it always checks against the root route since the child routes are not loaded. Any idea?

I guess the tests are working only some of them are failing: https://github.com/db-ui/mono/actions/runs/7826104309/job/21351476929?pr=2184#step:10:20

It's only the navigation-item that has an issue. If the routing doesn't work in the CI all Tests would fail.

packages/components/mitosis.config.js
d-koppenhagen: I though this too, but it works when only setting it in this file? What's the other file for?

nmerget (approved):

PR #2189, opened 2024-02-08
fix: uuid for option keys

### LINKED ISSUE #2188: fix(db-select): options by prop leads to duplicated keys

- [ ] All
- [ ] HTML
- [X] React
- [ ] Angular
- [ ] Vue
- [ ] Web components
- [ ] Power Apps

### Reproduction case

when using `<DBSelect>` with props `option` the generated key isn't unique. This leeds to an error message in the developer  console:

```
Warning: Each child in a list should have a unique "key" prop.
```

### Expected Behaviour

Generated key should be unique

### Screenshots

_No response_

### Browser version

None

_No response_

annsch:
## Proposed changes

fixes #2188

## Types of changes


packages/components/src/components/select/select.lite.tsx
```diff
@@ -128,17 +127,14 @@ export default function DBSelect(props: DBSelectProps) {
 							<>
 								<Show when={option.options}>
 									<optgroup
-										key={'optgroup-' + option.value}
+										key={'optgroup-' + uuid()}
```
nmerget, line 130: I would suggest that we pass an `id` to the option and use `uuid()` as fallback:

`model.ts
export type DBSelectOptionType = {
id?:string
}
`

annsch: @nmerget fixed this for options

packages/components/src/components/select/select.lite.tsx
```diff
@@ -137,7 +136,9 @@ export default function DBSelect(props: DBSelectProps) {
 												<option
 													key={
 														'option-' +
-														optgroupOption.value
+														optgroupOption.id
+															? optgroupOption.id
+															: uuid()
```
mfranzke, line 141:
```suggestion
														optgroupOption.id || uuid()
```

mfranzke:
@annsch is this _ready for review_?

annsch:
> @annsch is this _ready for review_?

@mfranzke yes, please

packages/components/src/components/select/select.lite.tsx
nmerget: > 

@annsch and I talked and we don't need a custom `id` on the `key={}`. `uuid()` is totally fine

nmerget (approved):

nmerget (approved):

PR #2231, opened 2024-02-18
feat: auto-determine nav item active state

### LINKED ISSUE #2200: `db-navigation-item`: determine active state from `aria-current` of `routerLink`
Currently setting the active state of a `db-navigation-item` must be realized by setting the `active` input to a boolean value.
This values is determined by accessing the Angular Router like this:

```ts
import { Router, RouterLink } from '@angular/router';
import { Component, Input } from '@angular/core';
import { DBNavigationItem } from '@db-ui/ngx-components';

@Component({
  selector: 'app-nav-item',
  template: `<db-navigation-item [active]="isActive('/')">
    <a routerLink="/">Home</a>
  </db-navigation-item>
  <db-navigation-item [active]="isActive('imprint')">
    <a routerLink="/imprint">Imprint</a>
  </db-navigation-item>`,
  imports: [RouterLink, DBNavigationItem, NavigationContentDirective],
  standalone: true
})
export class NavItemComponent {
  constructor(private readonly router: Router) {}

  isActive = (path: string) =>
    path === '/'
      ? this.router.url === '/'
      : this.router.url.includes(this.navItem.path);
}
```

This is quite complicated, since we must:
- Inject the Router Instance
- Two different place where we have to evaluate the route path (`routerLink`, `<db-navigation-item>`: `active`)
- a component method / function which checks the active state

Furthermore it's recommended to set the `aria-current` attribute for navigation items to improve the accessibility.

Therefore, Angular already provides of the `ariaCurrentWhenActive` attribute to automatically mark a link (`<a>`), when the current route matches with the target.

```html
<a routerLink="/" ariaCurrentWhenActive="page">Home</a>
```

It should be possible to find out from the parent component  `DBNavigationItem` if it has a child component where the `aria-current` attribute it defined (as a result of the `ariaCurrentWhenActive` directive on the `routerLink`). In this case, the `active` state could be inferred from this. The precedence should be lower compared to actively setting the attribute on the component.

The solution could look like this:

```ts
export class DbNavigationItemComponent implements AfterContentInit {
  @Optional() @ContentChild('routerLinkWithAriaCurrent', { read: ElementRef }) routerLinkElement?: ElementRef;

  active: boolean = false;

  constructor(private renderer: Renderer2) {}

  ngAfterContentInit(): void {
    this.checkAriaCurrent();
  }

  private checkAriaCurrent(): void {
    if (this.routerLinkElement) {
      const ariaCurrent = this.renderer.getAttribute(this.routerLinkElement.nativeElement, 'aria-current');
      this.active = ariaCurrent === 'page';
    }
  }
}
```

<a routerLink="/" routerLinkActive="active" ariaCurrentWhenActive="page">Home Page</a>

d-koppenhagen:
It will make the the `db-navigation-item`'s `active` prop  optional. It highlights an active link element with it's children. It still works as before but in addition it checks for child elements. They styles like active elements when having the `aria-current` attr set. It allows to use framework build-in aria-current mechanisms.

closes #2200

## Types of changes

### DIFF
--- a/packages/components/src/components/main-navigation/main-navigation.scss
+++ b/packages/components/src/components/main-navigation/main-navigation.scss
@@ -13,6 +13,10 @@
 		inline-size: auto;
 	}
 
+	.db-navigation-item:has([aria-current="page"]):first-of-type button {
+		font-weight: 700;
+	}
+
 	& > menu {
 		display: flex;
 		flex-direction: column;
@@ -82,6 +86,17 @@
 						}
 					}
 				}
+			}
+		}
+	}
+
+	:has(.db-navigation-item [aria-current="page"]) {
+		// add puls for main-navigation items
+		@extend %show-db-puls;
+
+		menu {
+			&::after {
+				display: none;
 			}
 		}
 	}
--- a/packages/components/src/components/main-navigation/docs/Angular.md
+++ b/packages/components/src/components/main-navigation/docs/Angular.md
@@ -11,7 +11,10 @@ import { DBMainNavigation } from '@db-ui/ngx-components';
 @Component({
   // ...
   standalone: true,
-  imports: [..., DBMainNavigation],
+  imports: [
+	// ...,
+    DBMainNavigation
+  ],
   // ...
 })
 ```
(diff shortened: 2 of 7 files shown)


packages/components/src/components/main-navigation/main-navigation.scss
```diff
@@ -13,6 +13,10 @@
 		inline-size: auto;
 	}
 
+	.db-navigation-item:has([aria-current="page"]):first-of-type button {
+		font-weight: 700;
```
nmerget, line 17: image

You can remove this selector and place the `font-weight:700` into the `:has(.db-navigation-item [aria-current="page"]) {` selector on line 93. It should be bold when a child is selected

packages/components/src/components/main-navigation/main-navigation.scss
```diff
@@ -85,4 +89,15 @@
 			}
 		}
 	}
+
+	:has(.db-navigation-item [aria-current="page"]) {
```
nmerget, line 93: `.db-navigation-item:has([aria-current="page"]):not(:has(menu)),
	:has(.db-navigation-item [aria-current="page"]) {`
	
We could add something like this for all navigation-items without a sub-navigation.

d-koppenhagen: Yes! To be honest I am not that deep into the SCSS structure here. The current state was more like a breakthrough for checking if my idea worked.
Feels free to edit / commit some changes you'd like. I am sure your code for scss styles looks better than mine!

mfranzke: I've add a follow up issue for this: https://github.com/db-ui/mono/issues/2254

nmerget:
@d-koppenhagen I created a new branch with some minor changes for the `scss` https://github.com/db-ui/mono/pull/2259, hope it's fine if we close this one?

mfranzke:
closed in favour of https://github.com/db-ui/mono/pull/2259

PR #2278, opened 2024-02-29
docs(other design systems): added new entry

mfranzke:
## Types of changes


docs/scripts/component-research/design-systems.js
```diff
@@ -51,6 +51,11 @@ export const designSystems = [
 		url: 'https://build.washingtonpost.com/',
 		componentUrl: 'https://build.washingtonpost.com/components/accordion'
 	},
+	{
```
nmerget, line 54: Do we need this as research for new components?

mfranzke: sure, why not ?

nmerget: Because we have already 15 Design Systems, adding more to the list would only increase the amount of work.

Furthermore, the DS is ugly :P

mfranzke: > Because we have already 15 Design Systems, adding more to the list would only increase the amount of work.
> 
> Furthermore, the DS is ugly :P

I agree on that topic in general, but actually this Design System is quite mature and often mentioned regarding aspects like accessibility for quite a while. So if you feel that we might have too many entries in our list, we should probably kick out some other entry instead.

nmerget (approved):

PR #2282, opened 2024-03-01
fix: global font-size issue and add documentation

### LINKED ISSUE #2252: Global style definitions

- [ ] All
- [ ] HTML
- [X] React
- [ ] Angular
- [ ] Vue
- [ ] Web components
- [ ] Power Apps

### Reproduction case

It seems a global style definition coming from `@db-ui/components/build/styles/db-ui-42-rollup` (
image
) is making font-size inheritance impossible unless we explicitly use `font-size: inherit | actual value` to child elements (which acutally beats the purpose of inheritance in CSS).

Steps to reproduce:

1. Import `@db-ui/components/build/styles/db-ui-42-rollup` from the DB UI library in order to gain access to combined dependencies like variables, colors and tonality and all components
2. Generate a simple HTML block containing nested elements where the root element has a defined font-size to a value X, e.g. 			
```html
<style type="text/css">
  .test-class {
    font-size: 12px;
  }	
</style>
<div className="test-class">
  <ul>
    <li>List item 1</li>
    <li>List item 2</li>
  </ul>
  <p>Test paragraph</p>
</div>
```

### Expected Behaviour

`font-size` inheritance should be working without having to use the `inherit` keyword or similar.

### Screenshots

I've attached screenshots in the "Reproduction case". If you want, in the future I can attach them here only, just let me know.

### Browser version

None

_No response_

nmerget:
## Proposed changes

closes #2252 

## Types of changes


packages/foundations/scss/fonts/_index.scss
mfranzke, line 1: why not name the file `packages/foundations/scss/fonts/_font.scss` as `packages/foundations/scss/fonts/_index.scss` directly ?

packages/foundations/scss/_init.scss
```diff
@@ -37,6 +38,12 @@ html {
 
 // global
 * {
+	font-size: inherit;
+	line-height: inherit;
```
mfranzke, line 42: in general, this is even already how the browser handles these styles, by inheriting these from their parent element. But most likely you're setting these because of the CSS Variables computed values, aren't you? Could you please add a comment in the CSS code, so that we remember in the future as well?

packages/foundations/scss/fonts/_index.scss
nmerget: We should name it `_font-sizes`...
I want to add another `css` file in the future to control the font-family like we did it with the icons like `[data-font="my-constom-font"]`

mfranzke: ah, got it, you would want to include further files in there in the future …

__snapshots__/accordion-item/showcase/chromium/regular/neutral/DBAccordionItem-should-match-screenshot.png
mfranzke, line 1: @nmerget is the disabled state on accordions actually supposed to not get an opacity or dimmed colors any more ?

nmerget: deleted the tokens accidentally... reverted them

mfranzke (approved):

PR #2331, opened 2024-03-15
fix: minor issues with sub-navigation

nmerget:
## Proposed changes

Mobile sub-navigation didn't use `overflow:auto`

## Types of changes


packages/foundations/scss/_screen-sizes.scss
```diff
@@ -1,10 +1,10 @@
...
-$db-breakpoint-xs: functions.to-rem(360);
-$db-breakpoint-sm: functions.to-rem(720);
-$db-breakpoint-md: functions.to-rem(1024);
-$db-breakpoint-lg: functions.to-rem(1440);
-$db-breakpoint-xl: functions.to-rem(1920);
+$db-breakpoint-xs: 360px;
+$db-breakpoint-sm: 720px;
+$db-breakpoint-md: 1024px;
+$db-breakpoint-lg: 1440px;
+$db-breakpoint-xl: 1920px;
```
mfranzke, line 7: why do you change these general breakpoints from relative to absolute values?

packages/foundations/scss/_screen-sizes.scss
```diff
@@ -38,7 +38,7 @@ $db-breakpoint-xl: functions.to-rem(1920);
 			@content;
 		}
 
-		@media screen and (max-width: $screen) {
+		@media screen and (max-width: $screen - 1) {
```
mfranzke, line 41:
```suggestion
		@media screen and (width < $screen) {
```

packages/foundations/scss/_screen-sizes.scss
nmerget: To do a calculation in the media-query below. But this is obsolete with your suggestion using `<`.

What's the best value for breakpoints? `px`, `rem` or `em`?

nmerget: Found a good answer: https://zellwk.com/blog/media-query-units/

packages/foundations/scss/_screen-sizes.scss
```diff
@@ -1,10 +1,10 @@
-@use "helpers/functions";
+@use "helpers";
```
mfranzke, line 1: why not leave this with including "functions" even only ?

packages/foundations/scss/_screen-sizes.scss
```diff
@@ -27,7 +27,7 @@ $db-breakpoint-xl: functions.to-rem(1920);
 	}
 
 	@if ($width == "min") {
-		@media screen and (min-width: $screen) {
+		@media screen and (width >= $screen) {
```
mfranzke, line 30:
```suggestion
		@media screen and (min-width: $screen) {
```

The argument to leave this as it was is that it's downward compatible.

mfranzke (approved):

packages/foundations/scss/_screen-sizes.scss
nmerget: It doesn't matter because it will only use the function `to-em`.
But we get an alignment for all files. We did it already for other files like `color` or `fonts` etc.

packages/foundations/scss/_screen-sizes.scss
nmerget: That's sad, it shouldn't break at all.

But I think we can skip it, it doesn't work on [IE](https://caniuse.com/css-media-range-syntax) but we don't support it anyways...

mfranzke: It's not just IE, it's even also some others, even though that all of these are under 1% market share:
https://caniuse.com/css-media-range-syntax

Nevertheless, if one of these might be more "resilient" than the other, I would have preferred that one.

nmerget: fixed

mfranzke (approved):

PR #2337, opened 2024-03-18
fix: wrong/missing react attributes

### LINKED ISSUE #2315: Different props available but not applied/working

- [ ] All
- [ ] HTML
- [X] React
- [ ] Angular
- [ ] Vue
- [ ] Web components
- [ ] Power Apps

### Reproduction case

1. use the DBTooltip component
2. assign some CSS values to its "style" prop 
3. CSS assigned to the "style" prop is not applied

### Expected Behaviour

The CSS values given to the "style" prop is applied.

### Screenshots

_No response_

### Browser version

None

There is a bigger list of props available for the DBTooltip component (probably for other DB components as well) which are not working / are mistakenly there. Some such props are `as`, `color`, `result`, `style` and so on.
The list of available props should contain only props which are supported by the component.

nmerget:
## Proposed changes

closes #2315

## Types of changes


__snapshots__/accordion-item/showcase/chromium/regular/neutral/DBAccordionItem-should-match-screenshot.png
mfranzke, line 1: @nmerget the "(Default) Collapsed" label within the regular, collapsed accordion isn't getting displayed anymore. Most likely this isn't on purpose.

packages/components/src/components/accordion-item/accordion-item.lite.tsx
```diff
@@ -54,9 +54,9 @@ export default function DBAccordionItem(props: DBAccordionItemProps) {
 			open={state._open}
 			name={props.name}>
 			<summary onClick={(event) => state.toggle(event)}>
-				<Show when={props.title}>{props.title}</Show>
-				<Show when={!props.title}>
-					<Slot name="title" />
+				<Show when={props.headline}>{props.headline}</Show>
+				<Show when={!props.headline}>
+					<Slot name="headline" />
 				</Show>
```
mfranzke, line 60:
```suggestion
				<Slot name="headline">{props.headline}</Slot>
```

Wouldn't this work as well with Mitosis? At least I think to remember this by Web Components slots. It doesn't do exactly the same, doesn't check for the `props.headline` property, but is a little simpler.

nmerget: It's a little bit confusing...
At the moment Mitosis uses `slotHeadline` for React which isn't the case for new versions. I changed it already in the [beta branch](https://github.com/db-ui/mono/blob/feat-color-palette/packages/components/src/components/accordion-item/model.ts#L28).

Is your solution working?

mfranzke: > Is your solution working?

I haven't tried it, I wanted to either ask for your experience/knowledge or trigger your curiosity.

nmerget: ` <ng-content> element cannot have content. [plugin angular-compiler]

    ../../output/angular/src/components/accordion-item/accordion-item.ts:28:8:
      28 │         <ng-content select="[headline]">{{headline}}</ng-content>
`

Nope it's not working for angular

__snapshots__/accordion-item/showcase/chromium/regular/neutral/DBAccordionItem-should-match-screenshot.png
nmerget: https://github.com/db-ui/mono/pull/2337/files#diff-65cfdea3944d52150c5413bfc9852e938d057e4ebe66d343e5555ae4771210c2R52 <- the `title` wasn't renamed

packages/components/src/components/accordion-item/accordion-item.lite.tsx
mfranzke: okay. Let's move to Web Components.

__snapshots__/accordion-item/showcase/chromium/regular/neutral/DBAccordionItem-should-match-screenshot.png
nmerget: fixed

mfranzke (approved):

PR #2351, opened 2024-03-19
fix: issue with generating uuid for every fragment

nmerget:
## Proposed changes

We added a `uuid()` for every `React.Fragment` in https://github.com/db-ui/mono/pull/2189.
This is causing a lot of rerenders which is bad.

 - Removed all unused Fragment + uuid implementations, and only add the uuid for fragments inside `select`.
 - Removed `stylePath` because it was adding Fragment, while we don't use it anymore.

## Types of changes

### DIFF
--- a/packages/components/src/components/popover/popover.lite.tsx
+++ b/packages/components/src/components/popover/popover.lite.tsx
@@ -1,13 +1,6 @@
-import {
-	onMount,
-	Show,
-	useMetadata,
-	useRef,
-	useStore
-} from '@builder.io/mitosis';
-import { DBPopoverState, DBPopoverProps } from './model';
-import { cls, uuid } from '../../utils';
-import { DEFAULT_ID } from '../../shared/constants';
+import { useMetadata, useRef, useStore } from '@builder.io/mitosis';
+import { DBPopoverProps, DBPopoverState } from './model';
+import { cls } from '../../utils';
 import { ClickEvent } from '../../shared/model';
 
 useMetadata({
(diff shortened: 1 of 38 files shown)


packages/components/src/components/popover/popover.lite.tsx
```diff
@@ -47,9 +33,7 @@ export default function DBPopover(props: DBPopoverProps) {
 			onClick={(event: ClickEvent<HTMLElement>) =>
 				state.handleClick(event)
 			}>
-			<Show when={state.stylePath}>
-				<link rel="stylesheet" href={state.stylePath} />
-			</Show>
+			{' '}
```
mfranzke, line 36:
```suggestion
```

I assume that this is a leftover – or have you included this one intentionally ?

packages/components/src/components/section/section.lite.tsx
```diff
@@ -34,9 +31,7 @@ export default function DBSection(props: DBSectionProps) {
 			id={state._id}
 			className={cls('db-section', props.className)}
 			data-size={props.size || 'medium'}>
-			<Show when={state.stylePath}>
-				<link rel="stylesheet" href={state.stylePath} />
-			</Show>
+			{' '}
```
mfranzke, line 34:
```suggestion
```
@nmerget I assume that this is a leftover – or have you included this one intentionally ?

packages/components/src/components/tooltip/tooltip.lite.tsx
```diff
@@ -50,9 +35,7 @@ export default function DBTooltip(props: DBTooltipProps) {
 			onClick={(event: ClickEvent<HTMLElement>) =>
 				state.handleClick(event)
 			}>
-			<Show when={state.stylePath}>
-				<link rel="stylesheet" href={state.stylePath} />
-			</Show>
+			{' '}
```
mfranzke, line 38:
```suggestion
```

@nmerget I assume that this is a leftover – or have you included this one intentionally ?

packages/components/src/components/popover/popover.lite.tsx
nmerget: fixed

packages/components/src/components/tooltip/tooltip.lite.tsx
nmerget: fixed

packages/components/src/components/section/section.lite.tsx
nmerget: fixed

mfranzke (approved):

PR #2361, opened 2024-03-20
fix: issue with accordion with single behaviour

nmerget:
## Proposed changes

- `DBAccordion` `behaviour="single"` was broken, because `accordion-item` had no id

## Types of changes


mfranzke:
@nmerget too bad that we cannot reproduce how this bug got introduced …

mfranzke (approved):

PR #2373, opened 2024-03-21
chore: add new workflow to regenerate snapshots on failure

nmerget:
## Types of changes


.github/workflows/default.yml
```diff
@@ -43,6 +43,11 @@ jobs:
     uses: ./.github/workflows/02-e2e-showcases.yml
     needs: [build-showcases]
 
+  regenerate-snapshots:
+    if: always() && (needs.test-components.result == 'failure' || needs.test-foundations.result == 'failure' || needs.test-showcases.result == 'failure')
+    uses: ./.github/workflows/02-e2e-regenerate.yml
+    needs: [test-components, test-showcases, test-foundations]
+
```
mfranzke, line 50: @nmerget to my understanding, it would be possible to split those up and only run them conditionally if either components, foundations or showcases fails, but not running all of them, if only some of them fail, wouldn't it ?

nmerget: We could do this, but then we need to merge the `__snapshots__` together for the case where two or all of them would fail. It's not a big deal if we run them all, most of the time is used by setting up the playwright image

mfranzke (approved):

PR #2378, opened 2024-03-21
Docs mention routers for frameworks

bruno-sch:
## Proposed changes

Adds docs for the use of router links in the corresponding frameworks (Angular, React, Vue).

## Types of changes


packages/components/docs/router-usage.md
```diff
@@ -0,0 +1,53 @@
+# Router usage
+
+> **Note:** At the moment, router links can only be used within the **DBNavigationItem** component!
+
+Unfortunately, the **DBLink** and **DBButton** components do not support router links. Explanation: Router links are usually rendered as anchor tags, which are not permitted within anchor or button tags.
```
nmerget, line 5: That's a good hint, we can also add a workaround comment here. The fallback would be `HTML` with the classes and `data-` attributes like this:

`<FrameworkLink class="db-button" data-variant="brand" to="my-subpage">Subpage</FrameworkLink>`

showcases/patternhub/data/routes.ts
```diff
@@ -6,6 +6,10 @@ export type NavigationItem = {
 };
 
 const componentChildren: NavigationItem[] = [
+	{
```
nmerget, line 9: Maybe we can move this to the line below [`readme`](https://github.com/db-ui/mono/blob/main/showcases/patternhub/data/routes.ts#L234)? The array should be for components ^.^

showcases/patternhub/package.json
```diff
@@ -19,7 +19,7 @@
     "next:start": "next start",
     "open": "open-cli http://localhost:3000/mono/sub",
     "prebuild:app": "cpr ../../node_modules/iframe-resizer/js/ public/iframe-resizer -o",
-    "start": "cross-env NEXT_PUBLIC_BASE_PATH=/mono/sub npm-run-all compile next:start"
+    "start": "cross-env NEXT_PUBLIC_BASE_PATH=/mono/sub npm-run-all compile next:dev"
```
mfranzke, line 22: is this change done on purpose ?

packages/components/docs/router-usage.md
mfranzke: test

nmerget (approved):

PR #2472, opened 2024-04-08
Feat tabs design changes no span

nmerget:
## Proposed changes

removed `span`

## Types of changes


packages/components/src/components/tab/tab.lite.tsx
```diff
@@ -35,10 +35,9 @@ export default function DBTab(props: DBTabProps) {
 	}, [ref, state.initialized]);
 
 	return (
-		<li className={cls('db-tab', props.className)}>
+		<li className={cls('db-tab', props.className)} role="tab">
```
mfranzke, line 38: Is the `role="tab"`-attribute correctly placed on this element or would it need to get placed on the `input` in this case (most likely it needs to be on the interactive element I think)?

mfranzke: https://github.com/db-ui/mono/issues/2476

PR #2484, opened 2024-04-10
docs: update overview for checkbox and radio

### LINKED ISSUE #2469: docs(radio & checkbox): incorrect structure
States:
– `Disabled` is an Interaction State

nmerget:
## Proposed changes

closes #2469

## Types of changes


showcases/shared/checkbox.json
```diff
@@ -35,48 +53,47 @@
...
 				}
 			},
 			{
-				"name": "Disabled",
+				"name": "Checked - Valid",
 				"props": {
 					"name": "States",
-					"disabled": true
+					"required": true,
+					"checked": true
```
mfranzke, line 74: this is making it "implicitly" valid – what about using `valid` property instead? If we'd like to propose this as a solution, we should also show it within our docs, as those would also be the basis for the HTML code underneath (to be adapted by developers).

mfranzke, line 74: this is making it "implicitly" valid – what about using `valid` property instead? If we'd like to propose this as a solution, we should also show it within our docs, as those would also be the basis for the HTML code underneath (to be adapted by developers).

nmerget: We want to automate most of the use-cases and `:user-valid` works like this for checkbox and radio. So we should keep it like this. In a talk with our design colleagues, we decided that we don't want to promote the `valid/invalid` to much, because we would get a custom behavior again

mfranzke: The programmatic validation is documented via https://github.com/db-ui/mono/blob/feat-color-palette/packages/components/docs/Validation.md

mfranzke (approved):

PR #2549, opened 2024-04-18
fix: db-page performance issue with :has

### LINKED ISSUE #2508: Performance heavy CSS selectors

- [ ] All
- [ ] HTML
- [X] React
- [ ] Angular
- [ ] Vue
- [ ] Web components
- [ ] Power Apps

### Reproduction case

```jsx
// Test.tsx
import { ReactNode, useState } from 'react';
import 'src/assets/db-ui-42-rollup.css';

const elements: Array<ReactNode> = [];

for (let i = 0, l = 8600; i < l; i++) {
	elements.push(<div key={i}>element {i}</div>);
}

export const Test = () => {
	const [value, setValue] = useState('');

	return (
		<div>
			<textarea value={value} onChange={(e) => setValue(e.target.value)}></textarea>
			<div
				className="test"
				style={{ display: 'grid', gridTemplateColumns: 'repeat(240, 1fr)', gap: 10 }}
			>
				{elements}
			</div>
		</div>
	);
};

// main.tsx
import { createRoot } from 'react-dom/client';
import { Test } from 'src/Test';

// React18 application setup code (see https://reactjs.org/docs/hello-world.html)
const root = createRoot(document.getElementById('root')!);
root.render(<Test />);
```

### Expected Behaviour

No delays are expected when typing something into the textarea element.

### Screenshots

_No response_

### Browser version

None

I managed to pinpoint the issue. I think it comes from the `/node_modules/@db-ui/components/build/components/page/page.scss` which generates this CSS code:

```css
:root:has(.fixed-header-footer) #app,
:root:has(.fixed-header-footer) #root,
:root:has(.fixed-header-footer) body,
:root:has(db-page[type=fixedHeaderFooter]) #app,
:root:has(db-page[type=fixedHeaderFooter]) #root,
:root:has(db-page[type=fixedHeaderFooter]) body {
	block-size: 100%
}
```

With this style, here is the performance snapshot while typing "asd" multiple times in the textarea:

And if we adjust the generated CSS to something like this:

```css
:root:has(.fixed-header-footer) #app,
/*:root:has(.fixed-header-footer) #root,
:root:has(.fixed-header-footer) body,*/
:root:has(db-page[type=fixedHeaderFooter]) #app
/*:root:has(db-page[type=fixedHeaderFooter]) #root,
:root:has(db-page[type=fixedHeaderFooter]) body*/ {
	block-size: 100%
}
```

we get perfect performance (no delay) while typing in the textarea element. Here performance snapshot with the commented out CSS code: 

I agree, we should not have that many DOM elements, but the issue appears with lower numbers as well.
Since there is a fix to it, would you please consider re-writing those selectors to be more performant?

nmerget:
## Proposed changes

closes #2508 

## Types of changes


__snapshots__/accordion-item/showcase/chromium/regular/neutral/DBAccordionItem-should-match-screenshot.png
mfranzke, line 1: @nmerget I don't think that this is looking as expected, is it?

nmerget: fixed

packages/components/src/components/page/page.lite.tsx
```diff
@@ -29,16 +30,24 @@ export default function DBPage(props: DBPageProps) {
...
 			state.fontsLoaded = true;
 		}
+
+		if (
+			props.documentOverflow === 'hidden' ||
+			(props.variant === 'fixed' && props.documentOverflow !== 'auto')
+		) {
+			document.documentElement.style.blockSize = '100%';
+			document.documentElement.style.overflow = 'hidden';
+		}
```
bruno-sch, line 40: If these styles are set onMount first, doesn't that create a possible **layout shift** after the pageload? Especially with SSR...

And could you perhaps explain why the styles are set to the documentElement and not to the body?

packages/components/src/components/page/model.ts
```diff
@@ -1,14 +1,21 @@
...
+	/**
+	 * The documentOverflow sets the overflow:hidden/auto to the root document
+	 */
+	documentOverflow?: 'hidden' | 'auto';
 	/**
 	 * Set this to have a transition with opacity to avoid layout-shifts https://simonhearne.com/2021/layout-shifts-webfonts/
 	 */
 	fadeIn?: boolean;
+	footer?: unknown;
+	header?: unknown;
```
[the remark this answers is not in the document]
nmerget: fixed

packages/components/src/components/page/model.ts
```diff
@@ -1,14 +1,21 @@
...
+
+	/**
+	 * The variant=fixed uses flex-box to make header and footer static
+	 */
+	variant?: 'default' | 'fixed';
```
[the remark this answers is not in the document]
nmerget: Shall we discuss this in another issue?

packages/components/src/components/page/page.lite.tsx
nmerget: Add a comment and move it to `OnInit`. I think it wouldn't work for SSR anyhow because of `document`

packages/components/src/components/page/model.ts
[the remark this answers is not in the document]
bruno-sch: Yes!

packages/components/src/components/page/page.lite.tsx
```diff
@@ -19,6 +20,20 @@ export default function DBPage(props: DBPageProps) {
...
+			(props.documentOverflow === 'hidden' ||
+				(props.variant === 'fixed' &&
+					props.documentOverflow !== 'auto'))
+		) {
+			// We need to set this to `html` element that the flex-box solution works
+			// See https://stackoverflow.com/a/43710216 - Approach 1 - flexbox
+			document.documentElement.style.blockSize = '100%';
+			document.documentElement.style.overflow = 'hidden';
+		}
+	});
```
bruno-sch, line 35: Still unsure whether this approach is the best solution...

for example:
**onInit** is transformed in vue to the hook **created**. at first this looks good: with ssr the code is not executed because document is undefined. 

with the ssr idea, pre-rendered html will be delivered, which ideally will not be modified in the browser until the first interaction. the styling is applied directly via css.

whether styles are written to dom elements via js in the created or mounted hook, sooner or later there will probably be a possible layout shift.

my idea would be to ask users to add these two lines of css in their own project css if variant fixed is needed.

nmerget: You are right it could be an issue with SSR. But we need to postpone this until we have a working showcase otherwise it's only guessing.

The `DBPage` is only an optional component, it's nothing you need to use but right now it adds some performance issues, for all components. So we should fix this first

bruno-sch (approved):

PR #2550, opened 2024-04-18
feat: add invisible variant for drawer

nmerget:
## Proposed changes

closes #2478

## Types of changes

### DIFF
--- a/packages/components/src/styles/_dialog-init.scss
+++ b/packages/components/src/styles/_dialog-init.scss
@@ -1,20 +1,5 @@
 @use "@db-ui/foundations/build/scss/colors";
 
-$backdrop-color: var(
-	--db-current-color-enabled,
-	var(--db-neutral-on-bg-enabled, #242629)
-);
-
-%backdrop {
-	background-color: $backdrop-color;
-	opacity: 0.64;
-}
-
-%backdrop-weak {
-	background-color: $backdrop-color;
-	opacity: 0.32;
-}
-
 %dialog-container {
 	position: fixed;
 	inset: 0;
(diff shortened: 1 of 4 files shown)


packages/components/src/styles/_dialog-init.scss
```diff
@@ -1,20 +1,10 @@
 @use "@db-ui/foundations/build/scss/colors";
 
 $backdrop-color: var(
-	--db-current-color-enabled,
-	var(--db-neutral-on-bg-enabled, #242629)
+	#{colors.$db-current-color-enabled},
+	var(#{colors.$db-neutral-on-bg-enabled}, #242629)
```
bruno-sch, line 5: Maybe you can explain it to me: What is the idea behind using a HEX value as an "unsafe" fallback?

packages/components/src/styles/_dialog-init.scss
```diff
@@ -34,13 +24,20 @@ dialog {
 
 	&:not([data-backdrop="none"]) {
 		&::backdrop {
-			@extend %backdrop;
+			background-color: $backdrop-color;
+			opacity: 0.64;
```
bruno-sch, line 28: The value for opacity is repeated at various points. Wouldn't it make sense to use global variables for this?

showcases/patternhub/styles/globals.scss
```diff
@@ -397,7 +397,8 @@ h6 {
 
 		.backdrop {
 			border-radius: variables.$db-border-radius-sm;
-			@extend %backdrop;
+			background-color: dialog-init.$backdrop-color;
+			opacity: 0.64;
```
bruno-sch, line 401: The value for opacity is repeated at various points. Wouldn't it make sense to use global variables for this?

packages/components/src/styles/_dialog-init.scss
nmerget: removed it, was a leftover of an old implementation

packages/components/src/styles/_dialog-init.scss
nmerget: fixed

showcases/patternhub/styles/globals.scss
nmerget: fixed

bruno-sch (approved):

packages/components/src/styles/_dialog-init.scss
```diff
@@ -34,13 +27,20 @@ dialog {
...
+			opacity: $backdrop-opacity-strong;
 		}
-	}
 
-	&[data-backdrop="weak"] {
-		&::backdrop {
-			@extend %backdrop-weak;
+		&[data-backdrop="invisible"] {
+			&::backdrop {
+				opacity: 0;
```
mfranzke, line 36: doesn't this totally hide it ? What's the difference in between `dialog.showModal()` against `dialog.show()` in this case anyhow ?

Issue #2196, opened 2024-02-09
Navigation: implement save triangle pattern


d-koppenhagen:
See:
https://medium.com/@nielsmanders/the-context-menu-safety-triangle-411c75065374

PR #2560, opened 2024-04-22
Feat navigation add save triangle pattern

bruno-sch:
## Types of changes

Fixes: #2196


packages/components/scripts/post-build/components.js
```diff
@@ -118,6 +118,20 @@ const getComponents = () => [
...
+			angular: [
+				{
+					from: 'navigationItemSafeTriangle = undefined;',
+					to: 'navigationItemSafeTriangle: undefined | NavigationItemSafeTriangle = undefined;'
+				}
+			],
+			vue: [
+				{
+					from: 'navigationItemSafeTriangle: undefined',
+					to: 'navigationItemSafeTriangle: undefined as undefined | NavigationItemSafeTriangle'
```
nmerget, line 131: `undefined as undefined` looks wired, is this intended?

packages/components/src/components/navigation-item/navigation-item.lite.tsx
```diff
@@ -36,9 +39,40 @@ export default function DBNavigationItem(props: DBNavigationItemProps) {
...
 			}
 		},
+
 		handleBackClick: (event: ClickEvent<HTMLButtonElement>) => {
 			event.stopPropagation();
 			state.isSubNavigationExpanded = false;
+		},
+
+		updateSubNavigationState: () => {
+			if (props.areaPopup !== undefined) {
```
nmerget, line 49: Do we need `state.initialized` here as well?

packages/components/src/utils/navigation.ts
```diff
@@ -6,6 +9,124 @@ export const isEventTargetNavigationItem = (event: unknown): boolean => {
...
+	public onMouseLeave() {
+		this.triangleData = undefined;
+	}
+
+	public onMouseMove(event: MouseEvent) {
+		if (!this.initialized || !this.triangleData) {
+			return;
+		}
+
+		const getTriangleTipX = (mouseX: number): number => {
```
nmerget, line 78: Can you move this to the class as private function? Otherwise it will be generated on every `onMouseMove`

packages/components/scripts/post-build/components.js
bruno-sch: Yes, intended!

Otherwise `onMouseEnter` would not be available at the following position `@mouseenter="navigationItemSafeTriangle?.onMouseEnter()"` in the template.

The `NavigationItemSafeTriangle` type is therefore required here. And `undefined` as well, otherwise the value here cannot be set initially with `undefined`.

--> `navigationItemSafeTriangle: undefined as undefined | NavigationItemSafeTriangle,`

packages/components/src/components/navigation-item/navigation-item.lite.tsx
bruno-sch: Did not cause any problems during testing. Why do you think it would be necessary here?

packages/components/src/utils/navigation.ts
```diff
@@ -6,6 +9,124 @@ export const isEventTargetNavigationItem = (event: unknown): boolean => {
...
+			if (this.triangleData.outsideVX === 'right') {
+				// vertical flipped triangle needs an inverted x pos
+				return this.triangleData.itemRect.width - mouseX;
+			}
+
+			// triangle stops shrinking from 75% x pos
+			return Math.min(mouseX, this.triangleData.itemRect.width * 0.75);
+		};
+
+		const getTriangleTipY = (mouseY: number): number => {
```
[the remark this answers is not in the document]
bruno-sch: That's right! We should take advantage of classes here :)

packages/components/src/components/navigation-item/navigation-item.lite.tsx
bruno-sch: @nmerget what do you think?

nmerget (approved):

mfranzke:
@bruno-sch the way from the navigation item to a subitem now feels amazing, but it felt a little strange when going back from the subitem to the "parent" item, not being able to move to the next item below because of the triangle. So it actually does a great job when moving "down", but when moving back up one level again e.g. to select another one, we might have to think about whether we disable the triangle after the user has reached the subitems.

packages/components/src/components/navigation-item/model.ts
```diff
@@ -9,19 +9,14 @@ import {
...
 	 * Indicator for active navigation item (bold font).
 	 */
 	active?: boolean;
 
-	/**
-	 * If the attribute is set the item acts like a button with a sub-navigation
-	 */
-
-	areaPopup?: boolean;
-
```
mfranzke, line 24: removing this would even also need to end up within the migration guide.

bruno-sch: I have mentioned the removed prop in these docs:  `docs/migration/alpha-beta.mda`

nmerget (approved):

PR #2581, opened 2024-04-25
test: add a11y tests to showcases

nmerget:
## Proposed changes

closes https://github.com/db-ui/mono/issues/2571

## Types of changes


package.json
```diff
@@ -39,14 +39,17 @@
     "regenerate:screenshots:rebuild": "npm run rm:builds && node scripts/rebuild-e2e.js",
     "rm:builds": "npm run rm:builds --workspace=scripts",
     "start": "npm run start --workspace=patternhub",
+    "test:lighthouse": "npm run build && npm run build --workspace=react-showcase && lhci autorun",
```
mfranzke, line 42:
```suggestion
    "pretest:lighthouse": "npm run build && npm run build --workspace=react-showcase",
    "test:lighthouse": "lhci autorun",
```

this makes it a little more universal (I hope).

nmerget: was a leftover, removed it

__snapshots__/notification/showcase/webkit/regular/neutral/DBNotification-should-match-screenshot.png
mfranzke, line 1: the headlines font-sizes don't rely on density settings anymore, this doesn't look correct.

nmerget: fixed

packages/components/src/components/notification/notification.lite.tsx
```diff
@@ -33,7 +33,7 @@ export default function DBNotification(props: DBNotificationProps) {
 			data-link-variant={props.linkVariant}>
 			<Slot name="image" />
 			<Show when={props.headline}>
-				<h6>{props.headline}</h6>
+				<header>{props.headline}</header>
```
mfranzke, line 36: @nmerget `header` doesn't provide the same semantic meaning as headline. This is most likely  related to that we couldn't determine the correct headline level, is it? In this case we should use a prop to retrieve this information.

packages/components/src/components/tab-item/tab-item.lite.tsx
```diff
@@ -18,8 +19,22 @@ export default function DBTabItem(props: DBTabItemProps) {
...
 	// jscpd:ignore-start
 	const state = useStore<DBTabItemState>({
-		_id: DEFAULT_ID,
-		initialized: false
+		initialized: false,
+		_selected: false,
+		handleChange: (event: ChangeEvent<HTMLInputElement>) => {
+			if (props.onChange) {
+				props.onChange(event);
+			}
```
mfranzke, line 27:
```suggestion
			props.onChange?.(event);
```

packages/components/src/components/tab-item/tab-item.lite.tsx
```diff
@@ -18,8 +19,22 @@ export default function DBTabItem(props: DBTabItemProps) {
...
+
+			if (props.change) {
+				props.change(event);
+			}
```
mfranzke, line 31:
```suggestion
			props.change?.(event);
```

packages/components/src/components/tab-item/tab-item.lite.tsx
```diff
@@ -35,18 +50,23 @@ export default function DBTabItem(props: DBTabItemProps) {
...
 			<label
-				htmlFor={state._id}
+				htmlFor={props.id}
 				data-icon={props.icon}
 				data-icon-after={props.iconAfter}
 				data-no-text={props.noText}>
 				<input
 					disabled={props.disabled}
+					aria-selected={state._selected}
+					aria-controls={props.controls}
```
mfranzke, line 62: as this attribute would include the `id` value of the tab panel, why couldn't we determine it's value by ourselves?

mfranzke: we actually do, and we already transfer that value within the `tabs` component.

showcases/e2e/fixtures/hover.ts
```diff
@@ -0,0 +1,14 @@
+/* eslint-disable no-await-in-loop,unicorn/prefer-dom-node-dataset */
+import { type Page } from '@playwright/test';
+
+export const hoverPre = async (page: Page, selector: string) => {
+	const components = await page.locator('main').locator(selector).all();
+	for (const component of components) {
+		await component.evaluate((comp: HTMLElement) => {
+			comp.setAttribute('data-testing-hover', 'true');
```
mfranzke, line 8:
```suggestion
			comp.setAttribute('data-e2e-hover', 'true');
```

packages/components/src/components/popover/popover.scss
```diff
@@ -4,6 +4,7 @@
 .db-popover {
 	position: relative;
 
+	&[data-testing-hover="true"],
```
mfranzke, line 7:
```suggestion
	&[data-e2e-hover="true"],
```

packages/components/src/components/tooltip/tooltip.scss
```diff
@@ -21,6 +21,11 @@ $tooltip-arrow-shadow-size: variables.$db-sizing-2xs;
 
 	padding: variables.$db-spacing-fixed-2xs variables.$db-spacing-fixed-xs;
 
+	// Only for testing
+	&[data-testing-hover="true"] {
```
mfranzke, line 25:
```suggestion
	&[data-e2e-hover="true"] {
```

packages/components/src/components/tabs/tabs.spec.tsx
```diff
@@ -51,21 +51,23 @@ const testActions = () => {
...
-// TODO: AXE has problems with tabs...
-
-/*test.describe('DBTabs', () => {
-	test('should not have any A11y issues', async ({ page, mount }) => {
+const testA11y = () => {
+	test('should not have A11y issues', async ({ page, mount }) => {
 		await mount(comp);
 		const accessibilityScanResults = await new AxeBuilder({ page })
+			// TODO: There might be an issue in axe-core: https://github.com/dequelabs/axe-core/issues/4431
+			// So we disabled "aria-allowed-role"
```
mfranzke, line 59:
```suggestion
			// TODO: There might be an issue in our implementation of which elements get which roles
			// So we disabled "aria-allowed-role" for now
```

packages/components/src/components/notification/notification.lite.tsx
nmerget: The notification is an `<article>` and the `<header>` can stand alone without anything: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/header?retiredLocale=de#usage_notes

Moreover you can pass the correct heading `<h1-6>` into it as a slot.

packages/components/src/components/tab-item/tab-item.lite.tsx
nmerget: We had this already, it doesn't work for every mitosis output...

packages/components/src/components/tab-item/tab-item.lite.tsx
nmerget: https://github.com/db-ui/mono/pull/2581#discussion_r1597942940

packages/components/src/components/tab-item/tab-item.lite.tsx
mfranzke: oh no … :-(

packages/components/src/components/tab-item/tab-item.lite.tsx
```diff
@@ -35,18 +50,23 @@ export default function DBTabItem(props: DBTabItemProps) {
 	}, [ref, state.initialized]);
 
 	return (
-		<li className={cls('db-tab-item', props.className)} role="tab">
+		<li className={cls('db-tab-item', props.className)} role="none">
```
mfranzke, line 53: why does this need to get a `role="none"`, but `label` doesn't? Or in other words, do we really need the `role="none"` here?

showcases/e2e/home/showcase-home.spec.ts
```diff
@@ -1,9 +1,33 @@
...
 		await expect(page).toHaveTitle('Showcase');
 	});
-	// TODO write separate form functionality tests without screenshots
+
+	test('should not have any A11y issues', async ({ page }) => {
+		const isAngular = process.env.showcase.startsWith('angular');
+
+		if (isAngular) {
+			test.skip();
+		}
```
mfranzke, line 16: could you please add a comment on why we're skipping the Angular test?

showcases/e2e/home/showcase-home.spec.ts
```diff
@@ -1,9 +1,33 @@
...
+		await page.goto('./', {
+			waitUntil: 'domcontentloaded'
+		});
+
+		await waitForDBPage(page);
+		const accessibilityScanResults = await new AxeBuilder({
+			page
+		})
+			// TODO: Currently disable till we solved https://github.com/db-ui/mono/issues/2587
+			// TODO: There might be an issue in axe-core: https://github.com/dequelabs/axe-core/issues/4431
```
mfranzke, line 27:
```suggestion
			// TODO: There might be an issue our implementation of which elements get which roles
```

showcases/react-showcase/src/components/index.tsx
```diff
@@ -5,8 +5,12 @@ import type {
...
 } from '../../../shared/default-component-data';
 
-const VariantList = ({ examples }: DefaultComponentVariants) => (
-	<DBCard className="variants-card">
+const VariantList = ({ examples, color }: DefaultComponentVariants) => (
+	<DBCard
+		className="variants-card"
+		elevationLevel={
+			color?.includes('3') ? '3' : color?.includes('2') ? '2' : '1'
+		}>
```
mfranzke, line 13: you've extracted this for vue. Wouldn't this make sense here as well (or is that a React anti-pattern unknown to me)?

packages/components/src/components/tab-item/tab-item.lite.tsx
nmerget: In the accessibility tree the `label` isn't shown because it has a `for` connected with the radio button. We need to add `role="none"` to the `li` so the `tab`(radio button) is a direct child of `tablist`.

showcases/e2e/home/showcase-home.spec.ts
nmerget: Added a comment for this

PR #2611, opened 2024-05-06
fix: issues with header with wrong spacings and dividers based on slots

nmerget:
## Proposed changes

DBSWC-2706

## Types of changes


packages/components/src/components/header/model.ts
```diff
@@ -10,10 +10,15 @@ import {
...
-	callToAction?: unknown;
-	actionBar?: unknown;
+	primaryAction?: unknown;
+	secondaryAction?: unknown;
 	drawerOpen?: boolean;
 
+	/**
+	 * Set max width for header
+	 */
+	width?: 'full' | 'medium' | 'large';
```
bruno-sch, line 20: The value `full` can be set for this prop, but as I can see from the rest of the code, it has no effect. So it would be redundant to `undefined` - right?

see: **packages/components/src/components/header/header.scss** (line 22 - 30)

nmerget: Yes, it's the same like `undefined`. I added it because of the figma specs for designers

bruno-sch: Okay, I get it. We should discuss elsewhere how we want to deal with default values in principle.

packages/components/src/components/page/page.scss
```diff
@@ -18,15 +18,15 @@
 		display: flex;
 		flex-direction: column;
 
-		header,
-		footer {
+		header:first-of-type,
+		footer:first-of-type {
```
mfranzke, line 22:
```suggestion
		> :is(header, footer) {
```

I assume that you mean by this the first header or footer within the document or directly underneath the page. But any nested `header` or `footer` would also be targeted by this selector, as it mainly means "on the same level", compare to e.g. https://jsbin.com/xetuxovoru/edit?html,css,output

packages/components/src/components/page/page.scss
```diff
@@ -18,15 +18,15 @@
...
 			display: flex;
 			flex: 0 1 auto;
 			flex-grow: 0;
 			flex-shrink: 0;
 		}
 
-		main {
+		main:first-of-type {
```
mfranzke, line 29:
```suggestion
		> .db-main {
```

@nmerget I assume that you mean by this the first `main` within the document or directly underneath the page. But any nested `main` would also be targeted by this selector, as it mainly means "on the same level", compare to e.g. https://jsbin.com/xetuxovoru/edit?html,css,output

packages/components/src/components/page/page.scss
nmerget: fixed

packages/components/src/components/page/page.scss
nmerget: fixed

packages/components/src/components/header/model.ts
mfranzke: https://github.com/db-ux-design-system/core-web/issues/2626

packages/components/src/components/page/page.scss
nmerget: There is no header showcase for angular I need to test it as well...

mfranzke (dismissed):

packages/components/src/components/page/page.scss
```diff
@@ -13,20 +22,18 @@
...
 		flex-direction: column;
-
-		header,
-		footer {
-			display: flex;
-			flex: 0 1 auto;
-			flex-grow: 0;
-			flex-shrink: 0;
+		/* workaround for angular */
+		> :is(db-header, dbheader, dbfooter, db-footer) {
```
mfranzke, line 32:
```suggestion
		> :is(.db-header, .db-footer, db-header, dbheader, dbfooter, db-footer) {
```

why don't you just leave this in here, and add further selectors to the list?

mfranzke (approved):

packages/components/src/components/page/page.scss
nmerget: The mixin `@include header-footer-style();` has those selectors in it already, this would be only a duplicate

PR #2647, opened 2024-05-15
feat: enable color contrast and fix issues with components

### LINKED ISSUE #2628: switch: doesn't work with High Contrast mode

- [X] All
- [ ] HTML
- [ ] React
- [ ] Angular
- [ ] Vue
- [ ] Web components
- [ ] Power Apps

### Reproduction case

The handle doesn't get displayed in High Contrast Mode, it doesn't react on the [`forced-colors` setting](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/forced-colors).

https://db-ui.github.io/mono/version/v0.1.3/components/switch/overview

### Expected Behaviour

The handle would still get displayed in [High Contrast mode](https://www.smashingmagazine.com/2022/06/guide-windows-high-contrast-mode/).

### Screenshots

image

### Browser version

None

_No response_

nmerget:
## Proposed changes

closes #2628

- enabled high contrast for playwright
- fixed issues with high contrast

## Types of changes


nmerget:
There is an issue with axe-playwright, I reported it already [here](https://github.com/dequelabs/axe-core-npm/issues/1067). We should think about ignoring the rule for now :(

packages/components/src/styles/_form-components.scss
```diff
@@ -152,6 +152,7 @@ $check-border-size: min(#{variables.$db-border-height-2xs}, 2px);
 
 	#{$selector},
 	[id$="-placeholder"] {
+		--db-current-icon-color: var(--db-#{$variant}-on-bg-enabled);
```
mfranzke, line 155: this reminds me of the [`currentcolor keyword`](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value#currentcolor_keyword) – we might either use it here, or what about within the icon declarations themselves, and by that provide further resilience and reduce possible redundance, wouldn't we ?

packages/components/src/components/button/button.scss
```diff
@@ -8,6 +8,7 @@
 @use "../../styles/button-components";
 
 .db-button {
+	--db-current-icon-color: #{colors.$db-current-color-enabled};
```
mfranzke, line 11: why do you additionally set this color variable? The [fallback for that variable not existing would be `inherit`](https://github.com/db-ui/mono/blob/main/packages/foundations/scss/icons/_icon-helpers.scss#L121). And if there's a problem with that, which I couldn't come up with at the moment, this also reminds me of the [`currentcolor keyword`](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value#currentcolor_keyword) – we might either use it here, or what about within the icon declarations themselves, and by that provide further resilience and reduce possible redundance, wouldn't it ?

packages/components/src/styles/_form-components.scss
nmerget: I added it because of the screenshots. But the high-contrast with dark-mode is somehow buggy. I'll revert the changes...

packages/components/src/components/button/button.scss
nmerget: https://github.com/db-ui/mono/pull/2647#discussion_r1607744482

__snapshots__/switch/showcase/chromium-highContrast/regular/neutral-bg-lvl-1/DBSwitch-should-match-screenshot.png
bruno-sch, line 1: The layout looks broken to me. Can this be avoided?

packages/components/src/components/switch/switch.scss
```diff
@@ -72,10 +72,13 @@ $checked-active-transition-size: calc(
 
 		// thumb
 		&::before {
-			content: "";
+			content: "O";
```
bruno-sch, line 75: Why `content: "O"`?

nmerget: The color contrast mode remove the background color from the `::before`. Therefore, the users wouldn't see the status of the switch. The "O" is a workaround. It is invisible for the default user, but it works for contrast users this way.

image
It doesn't look perfect, but seems to be a easy solution for this edge case

__snapshots__/switch/showcase/chromium-highContrast/regular/neutral-bg-lvl-1/DBSwitch-should-match-screenshot.png
nmerget: It's because of https://github.com/db-ui/mono/pull/2647#discussion_r1658450811. Maybe you have a better solution for this issue?

__snapshots__/switch/showcase/chromium/regular/neutral-bg-lvl-1/DBSwitch-should-match-screenshot.png
mfranzke, line 1: The `O` character is actually visible currently.

https://github.com/db-ui/mono/pull/2647/files#r1658450811

mfranzke (approved):

__snapshots__/switch/showcase/chromium/regular/neutral-bg-lvl-1/DBSwitch-should-match-screenshot.png
nmerget: fixed

mfranzke (approved):

mfranzke (approved):

PR #2673, opened 2024-05-23
fix: issue for patternhub included inside iframe

nmerget:
## Proposed changes

[Marketingportal](https://marketingportal.extranet.deutschebahn.com/marketingportal/Design-Anwendungen/db-ux-design-system/version-3/components/action/button) cannot load properties page in `iframe`.

## Types of changes

### DIFF
--- a/packages/components/src/components/page/page.lite.tsx
+++ b/packages/components/src/components/page/page.lite.tsx
@@ -53,8 +53,8 @@ export default function DBPage(props: DBPageProps) {
 			document.documentElement.style.overflow === 'hidden'
 		) {
 			// remove document styles set by this
-			document.documentElement.style.blockSize = null;
-			document.documentElement.style.overflow = null;
+			document.documentElement.style.blockSize = '';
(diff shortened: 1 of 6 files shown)


packages/components/src/components/page/page.lite.tsx
```diff
@@ -53,8 +53,8 @@ export default function DBPage(props: DBPageProps) {
 			document.documentElement.style.overflow === 'hidden'
 		) {
 			// remove document styles set by this
-			document.documentElement.style.blockSize = null;
-			document.documentElement.style.overflow = null;
+			document.documentElement.style.blockSize = '';
+			document.documentElement.style.overflow = '';
```
mfranzke, line 57: I think it would be easier to just toggle a CSS class within this file (even also in the other places, like line 33 and 34) and define the relevant styling within the (S)CSS file.

nmerget: fixed

mfranzke (approved):

PR #2686, opened 2024-05-27
fix: tag wasn't able to use default html-tags like `button`, `a`, `input`

### LINKED ISSUE #2678: DBTag [removeable]: includes a lot of unused/not displayed DOM from DBCheckbox

- [ ] All
- [ ] HTML
- [X] React
- [ ] Angular
- [ ] Vue
- [ ] Web components
- [ ] Power Apps

### Reproduction case

Currently DB Tag with prop "removeable"  also includes a lot of not displayed HTML Code generated by checkboxes' validation (<DBInfotext>) – is this really necessary?

````
<div class="db-tag" data-semantic="successful" data-icon="plus">
     <div class="db-checkbox">
           <label for="checkbox">
                <input type="checkbox">
                Label-Text
           </label>
         <span id="checkbox-valid-message" class="db-infotext" data-semantic="successful" data-size="small">TODO: Add a validMessage</span>
         <span id="checkbox-invalid-message" class="db-infotext" data-semantic="critical" data-size="small">TODO: Add an invalidMessage</span>
    </div>
</div>
````

### Expected Behaviour

When using DB Tag with removeable option, a lighter version of DBCheckbox should be rendered

### Screenshots

_No response_

### Browser version

None

_No response_

### PR DESCRIPTION
**Pull request by nmerget**
## Proposed changes

closes #2678

## Types of changes

### CODE REVIEW

File: packages/components/src/components/tag/tag.lite.tsx [hunk 1]
```diff
@@ -41,16 +41,17 @@ export default function DBTag(props: DBTagProps) {
...
-					noText
-					variant="ghost"
+					data-icon="cross"
+					data-size="small"
+					data-no-text="true"
+					data-variant="ghost"
 					title={state.getRemoveButtonText()}>
 					{state.getRemoveButtonText()}
-				</DBButton>
+				</button>
```
mfranzke, line 54: we would need to add migration guide entries for changes like this.

nmerget: Why? This is only the internal component. For pure HTML users nothing would change, as well for framework users.

showcases/shared/tag.json
```diff
@@ -142,41 +144,38 @@
...
-					{
-						"name": "checkbox"
-					}
-				]
+				"code": {
+					"angular": "<db-tag><label><input type=\"checkbox\"/>Interactive Checkbox</label></db-tag>",
+					"html": "<div class=\"db-tag\"><label><input type=\"checkbox\"/>Interactive Checkbox</label></div>",
+					"react": "<DBTag><label><input type=\"checkbox\"/>Interactive Checkbox</label></DBTag>",
+					"vue": "<DBTag><label><input type=\"checkbox\"/>Interactive Checkbox</label></DBTag>"
+				}
```
mfranzke, line 152: @nmerget to my understanding, we could use children.name only to reference DB UI components or native HTML elements, but none of those native nested tags constructs like e.g. `<label><input type=\"checkbox\"/>Interactive Checkbox</label>`, could we?

mfranzke: actually, it is, via nested `children`

showcases/patternhub/scripts/utils.js
```diff
@@ -193,6 +193,10 @@ export const getCodeByFramework = (
 		}
 	}
 
+	if (native && name === 'input') {
```
mfranzke, line 196: why do you specifically select for `input` here?

nmerget: It has to be `<input/>` and nextjs throws and build error if we use `<input></input>`

packages/components/src/components/tag/tag.lite.tsx
```diff
@@ -25,6 +32,26 @@ export default function DBTag(props: DBTagProps) {
...
+					'button:not(.db-tab-remove-button)'
+				);
+				const input: HTMLInputElement = ref?.querySelector('input');
+				for (const element of [button, input]) {
+					if (element) {
+						element.disabled = props.disabled;
+					}
+				}
+			}
+		}
```
mfranzke, line 52:
```suggestion
		if (state.initialized && ref && props.disabled !== undefined) {
		  const button: HTMLButtonElement = ref?.querySelector(
			  'button:not(.db-tab-remove-button)'
		  );
		  const input: HTMLInputElement = ref?.querySelector('input');
		  for (const element of [button, input]) {
			  if (element) {
				  element.disabled = props.disabled;
			  }
		  }
	    }
```

mfranzke (approved):

mfranzke (approved):

PR #2775, opened 2024-06-21
fix: issue with tooltip colors changed by parent component like tag

### LINKED ISSUE #2774: DBTooltip style is overwritten by parent data-emphasis="strong"

- [X] All
- [ ] HTML
- [ ] React
- [ ] Angular
- [ ] Vue
- [ ] Web components
- [ ] Power Apps

### Reproduction case

When using DBTooltip with DBTag (emphasis strong) as parent element, the displayed tooltip inherits is text color from [data-amphasis="strong"]

```html
<DBTag describedbyid="tooltip-01" semantic="neutral" emphasis="strong">
KUZ
<DBTooltip id="tooltip-01" placement="right-end" variant="with arrow">Beschreibungstext</DBTooltip>
</DBTag>
```

### Expected Behaviour

DBTooltip component`s style should not be affected by its parent style.

### Screenshots

_No response_

### Browser version

None

_No response_

nmerget:
## Proposed changes

closes #2774

## Types of changes


showcases/react-showcase/src/components/form/index.tsx
```diff
@@ -369,6 +370,28 @@ const FormComponent = () => {
 					validMessage="😎"
 					minLength={10}
 				/>
+
+				<DBTag
+					describedbyid="tooltip-01"
+					semantic="neutral"
+					emphasis="strong">
+					KUZ
```
mfranzke, line 378: https://de.wikipedia.org/wiki/KUZ ?

mfranzke (approved):

showcases/react-showcase/src/components/form/index.tsx
nmerget: used the example from the issue, I don't know

PR #2798, opened 2024-06-27
fix: issue with removable tag closing button height

### LINKED ISSUE #2793: DBTag: removable button with a height of only 22px

- [X] All
- [ ] HTML
- [ ] React
- [ ] Angular
- [ ] Vue
- [ ] Web components
- [ ] Power Apps

### Reproduction case

The clickable area of the removable button is only 22px height, compare to https://db-ui.github.io/mono/review/main/components/data-display/tag/overview?page=behaviour

### Expected Behaviour

The clickable area of the removable button should be 24px height.

### Screenshots

_No response_

### Browser version

None

We might increase it e.g. by a pseudo content element that extends to 24px.

nmerget:
## Proposed changes

closes #2793

## Types of changes

### DIFF
--- a/packages/components/src/components/tag/tag.scss
+++ b/packages/components/src/components/tag/tag.scss
@@ -190,7 +190,8 @@
 			// overwrite to transparent
 			background-color: colors.$db-current-color-bg-transparent-full-enabled;
 			margin-inline-start: variables.$db-spacing-fixed-2xs;
-			block-size: 100%;
+			border: none;
+			aspect-ratio: 1 / 1;
 		}
 	}
 }


packages/components/src/components/tag/tag.scss
```diff
@@ -190,7 +190,8 @@
 			// overwrite to transparent
 			background-color: colors.$db-current-color-bg-transparent-full-enabled;
 			margin-inline-start: variables.$db-spacing-fixed-2xs;
-			block-size: 100%;
+			border: none;
+			aspect-ratio: 1 / 1;
```
mfranzke, line 194: ♥️

mfranzke (approved):

packages/components/src/components/tag/tag.scss
nmerget: learned from the best

PR #2828, opened 2024-07-05
fix: issue with popover visibility, replaced with display:none

### LINKED ISSUE #2804: DBPopover should use `display:none` instead of `visibility:hidden`?

- [X] All
- [ ] HTML
- [ ] React
- [ ] Angular
- [ ] Vue
- [ ] Web components
- [ ] Power Apps

### Reproduction case

Currently DBPopover's content (`<article>`) is hidden by `visibility: hidden`. Therefore all children should implement `visibility: inherit` to get hidden as well. 

### Expected Behaviour

Wouldn't `display: none` better in this case?

### Screenshots

_No response_

### Browser version

None

_No response_

nmerget:
## Proposed changes

closes #2804 

## Types of changes

### DIFF
--- a/packages/components/src/styles/_popover-component.scss
+++ b/packages/components/src/styles/_popover-component.scss
@@ -151,7 +151,7 @@ $popover-gap-inset: calc(-1 * #{$popover-gap-size});
 }
 
 %show-popover {
-	display: inherit;
+	display: revert;
 
 	&:not([data-placement]):not([data-outside-vy]),
 	&[data-placement^="bottom"]:not([data-outside-vy]),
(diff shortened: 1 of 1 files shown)


packages/components/src/styles/_popover-component.scss
```diff
@@ -151,7 +151,7 @@ $popover-gap-inset: calc(-1 * #{$popover-gap-size});
 }
 
 %show-popover {
-	visibility: visible;
+	display: inherit;
```
mfranzke, line 154:
```suggestion
	display: revert;
```

I think that we would like to revert to the value of the UA styling instead of inheriting a value from another wrapping tag here.

mfranzke (approved):

PR #2906, opened 2024-07-24
test: write test for required inputs to read messages

### LINKED ISSUE #2624: form elements: description optimization with `aria-live` annotations
as another optimization we should evaluate to only insert the texts into the related description HTML tags to enable status messages being articulated by screenreaders, with further `role=status` and `role=alert` added to those HTML tags.

_Originally posted by @mfranzke in https://github.com/db-ui/mono/issues/2599#issuecomment-2103907831_

### LINKED ISSUE #2625: form elements

- [X] All
- [ ] HTML
- [ ] React
- [ ] Angular
- [ ] Vue
- [ ] Web components
- [ ] Power Apps

### Reproduction case

When switching between the different states of formelements like e.g. a `required` textarea the text content of the invalid HTML element (`.db-textarea:has(textarea:required) .db-infotext`) changes in between todo and the validation message that gets retrieved by the element on invalid state.

### Expected Behaviour

There shouldn't be a reason for the copy to change.

### Screenshots

_No response_

### Browser version

None

Most likely the solution will be part of https://github.com/db-ui/mono/issues/2624, as we should provide the copy of the message dynamically, instead of switching the HTML elements.

nmerget:
## Proposed changes

closes #2624
closes #2625

Because of the changes in `aria-describedby` the screen reader already announces changes coming from the valid/invalid messages.

## Types of changes


showcases/screen-reader/translations.ts
```diff
@@ -0,0 +1,16 @@
...
+	blank: ['Leer'],
+	checked: ['aktiviert'],
+	' of ': [' von '],
+	clickable: ['anklickbar'],
+	'has auto complete': ['mit Auto Vervollständigung'],
+	required: ['erforderlich'],
+	'Please fill out this field..': ['Fülle dieses Feld aus..'],
+	unselected: ['nicht ausgewählt'],
+	selected: ['ausgewählt'],
+	'': ['. Nummernblock eingeschaltet']
```
mfranzke, line 15: why is this an empty key?

mfranzke (dismissed):

showcases/screen-reader/translations.ts
nmerget: I guess there is some extra setting for keyboards with a Num block, which isn't the case on the Runner

packages/components/src/styles/visually-hidden.scss
mfranzke, line 1: what's the difference to the contents out of https://github.com/db-ui/mono/blob/main/packages/foundations/scss/helpers/_a11y.scss#L1-L8 ?

nmerget: fixed

packages/foundations/scss/helpers/_a11y.scss
```diff
@@ -1,8 +1,17 @@
...
 	position: absolute !important;
-	white-space: nowrap;
-	inline-size: 1px;
+	inline-size: 1px !important;
+	border-width: 0 !important;
+	border-style: initial !important;
+	border-color: initial !important;
+	border-image: initial !important;
+	padding: 0 !important;
+	pointer-events: none !important;
```
mfranzke, line 16: @nmerget where have you gotten all of those new properties from? In general a smaller amount should even already do the trick, compare to: https://github.com/h5bp/html5-boilerplate/blob/main/dist/css/style.css#L100-L120

nmerget: I looked at https://component.gallery/components/visually-hidden/ and used everyhting from other design systems ^.^

mfranzke (approved):

mfranzke (approved):

mfranzke (approved):

PR #2968, opened 2024-08-03
refactor: replaced by simpler solution

mfranzke:
## Proposed changes

Resolves https://github.com/db-ui/mono/issues/2967

## Types of changes

### DIFF
--- a/packages/components/src/components/navigation/navigation.scss
+++ b/packages/components/src/components/navigation/navigation.scss
@@ -80,20 +80,11 @@
 					}
 				}
 			}
+		}
 
-			&:has([aria-current="page"]),
+		> .db-navigation-item {
 			&[aria-current="page"] {
 				@extend %show-db-puls-auto;
-
-				menu {
-					// hide puls for non navigation items
-					:has([aria-current="page"]),
-					[aria-current="page"] {
-						&::after {
-							display: none;
-						}
-					}
-				}
 			}
 		}
 	}


__snapshots__/navigation/showcase/chromium-highContrast/regular/neutral-bg-lvl-1/DBNavigation-should-match-screenshot.png
bruno-sch, line 1: Do you have any idea why the entire new screenshot is shifted down a little?

mfranzke: no, I even also wondered about this aspect. Even though it might be more correct than the previous solution, as we don't set an element that we hide afterwards anymore.

packages/components/src/components/navigation/navigation.scss
```diff
@@ -80,20 +80,11 @@
 					}
 				}
 			}
+		}
 
-			&:has([aria-current="page"]),
+		> .db-navigation-item {
 			&[aria-current="page"] {
 				@extend %show-db-puls-auto;
```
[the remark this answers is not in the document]
mfranzke: I think to remember that this wasn't even only about line 84 within the original file, but especially about the upfollowing 88 till 96, that even also provided a concatenated `:has` selector. I'll further investigate on this one.

bruno-sch (approved):

bruno-sch (dismissed):

nmerget (approved):

nmerget (approved):

PR #2995, opened 2024-08-06
chore(docs): added brackets

mfranzke:
## Types of changes


nmerget:
@mfranzke Why do we need this?

mfranzke:
> @mfranzke Why do we need this?

I think it would be easier to understand that this is a method you're refering to. Elsewhere "next" might be anything, from a product name, to a library, to whatever …

nmerget (approved):

PR #2997, opened 2024-08-06
test(guidepup): accordion

bruno-sch:
## Proposed changes

Adds guidepup test for accordion

## Types of changes


showcases/screen-reader/tests/accordion.spec.ts
mfranzke, line 1: This test always runs into a timeout and might as well test a few things, that we could easily test via the regular tests even only, like the different behaviour of single vs. multiple items being open. So we should most likely be able to shrink this implementation to navigating in between two items and ensure that the first items content doesn't get announced in the first place, move back to the first one, interact with it, move to it's (now visible and announceable) content and than move to the next item, which would ensure that we haven't crashed with the standard implementation of both visually and screenheader hidden content of those collapsing elements.

mfranzke: What about the following two lines in between 34 and 35, so that we could "prove" that the accordions contents are hidden in the first place, and than only opened after the interaction?

```javascript
await screenReader?.next(); // Focus: "item 2"
await screenReader?.previous(); // Focus: "item 1"
```

bruno-sch: Makes sense. I have added it.

showcases/screen-reader/__snapshots__/macos/webkit/DBAccordion-default-1.txt
```diff
@@ -0,0 +1 @@
+["Item 1 collapsed summary","Item 2 collapsed summary","Item 1 collapsed summary","","Content 1","Item 2 collapsed summary"]
```
bruno-sch, line 1: @mfranzke The change you suggested causes a behavior that I cannot explain:

With `act()` on “item 1” only an **empty string** is output by voiceover, instead of `"Item 1 chevron_down expanded summary"` before.

mfranzke: > @mfranzke The change you suggested causes a behavior that I cannot explain:
> 
> With `act()` "on item 1" only an **empty string** is output by voiceover, instead of `"Item 1 chevron_down expanded summary"` before.

are you sure, that this behaviour of `act()` is related to my suggested navigation one item forward and one back? previously to doing the `act()`?

bruno-sch: I have created a new issue for this matter: https://github.com/db-ui/mono/issues/3133

Can you please approve the PR now? :)

nmerget (approved):

PR #3005, opened 2024-08-07
test(guidepup): checkbox

bruno-sch:
## Proposed changes

adds guidepup test for checkbox

## Types of changes


mfranzke (approved):

showcases/screen-reader/__snapshots__/windows/chromium/DBCheckbox-default-1.txt
```diff
@@ -0,0 +1 @@
+["check box, half checked, Required Indeterminate","check box, half checked, Required Indeterminate","check box, half checked, Required Indeterminate","star","check box, not checked, required, Required star","TODO: Add a valid Message. checked","Required","star","TODO: Add a valid Message","star","Required","check box, checked, required, Required star","Please check this box if you want to proceed.. not checked","Required","star","Please check this box if you want to proceed."]
```
nmerget, line 1: We should add an alternative text empty for the "*" like we did it with the icons: https://github.com/db-ui/mono/blob/main/packages/foundations/scss/icons/_icon-helpers.scss#L47-L50

packages/components/src/styles/_form-components.scss
```diff
@@ -265,7 +265,7 @@ $input-valid-types: "color", "date", "datetime-local", "email", "file", "hidden"
 		&:is(label),
 		label {
 			&::after {
-				content: "*";
+				content: "*" / "";
```
nmerget, line 268: @bruno-sch you need to add the `supports` stuff as well: https://developer.mozilla.org/en-US/docs/Web/CSS/@supports

showcases/screen-reader/tests/checkbox.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+				await voiceOver?.next(); // Focus checkbox 2 message
+				await voiceOver?.next(); // Focus checkbox 3
+				await voiceOver?.previous(); // Focus checkbox 2 message
+				await voiceOver?.previous(); // Focus checkbox 2 label
+				await voiceOver?.previous(); // Focus checkbox 2
+				await voiceOver?.act(); // Tick checkbox 2
+				await voiceOver?.next(); // Focus checkbox 2 label
+				await voiceOver?.next(); // Focus checkbox 2 message
+				await voiceOver?.next(); // Focus checkbox 3
+			}
```
mfranzke, line 43: it looks like that the tests of both screenreaders are set up totally differently, probably to check for screenreader specific aspects that we've analyzed during testing. Is there a reason for this? Than we should at least document it with a simple comment.

bruno-sch: @nmerget can you explain why your test for nvda is so much more compact?

mfranzke: From what we discussed last week, one of the reasons is that it takes less commands to go over the elements and trigger the same actions. But I would additionally need to go through it whether it actually does the same things/takes the same steps (because it seems to be different).

mfranzke (dismissed):

mfranzke (dismissed):

showcases/screen-reader/tests/checkbox.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+			'should tick and untick checkbox, feedback messages must appear',
+		url: './#/03/checkbox?page=requirement',
+		async testFn(voiceOver, nvda) {
+			if (nvda) {
+				await nvda?.next(); // Focus checkbox 2
+				await nvda?.act(); // Tick checkbox 2
+				await nvda?.next(); // Focus checkbox 2 message
+				await nvda?.previous(); // Focus checkbox 2
+				await nvda?.act(); // Tick checkbox 2
+				await nvda?.next(); // Focus checkbox 2 label
```
mfranzke, line 19: at least the comments (and that for the expected output) of line 16 and 19 doesn't seem to be in sync. on line 16 it's being mentioned that after navigating away from the checkbox to the next element we would reach the checkboxes message (without navigating onto the `label` first), while instead on line 19 we would actually reach that `label` on the same way of navigation.

nmerget (approved):

PR #3007, opened 2024-08-07
test(guidepup): brand

bruno-sch:
## Proposed changes

test: adds guidepup test for `brand` component

## Types of changes


showcases/screen-reader/__snapshots__/windows/chromium/DBBrand-default-1.txt
mfranzke, line 1: we need to identify how we could prevent an false positive like this.

mfranzke: It seems like that NVDA does provide that further text in case that there's some very small amount of words within the alternative-text.

mfranzke: The minimum to fulfill this aspect seems to be two phrases (our minimum phrase that we found out to be working was "Logo 1").

mfranzke: final: The minimum to fulfill this aspect seems to be at least 6 characters (our minimum phrase that we found out to be working were "Logoo1" or "Logo 1").

nmerget (approved):

PR #3009, opened 2024-08-07
refactor(guidepup): simplification

mfranzke:
## Proposed changes

Let's further simplify a central conditional check.

## Types of changes


showcases/screen-reader/default.ts
```diff
@@ -34,15 +34,15 @@ const cleanSpeakInstructions = (phraseLog: string[]): string[] =>
...
-					!(
-						sPhrase.includes('You are currently') ||
-						sPhrase.includes('To enter') ||
-						sPhrase.includes('To exit') ||
-						sPhrase.includes('To click') ||
-						sPhrase.includes('To select') ||
-						sPhrase.includes('To interact') ||
-						sPhrase.includes('Press Control')
-					)
+					![
```
nmerget, line 37: Thanks :)

We should move the array outside of the functions as well

nmerget (approved):

PR #3033, opened 2024-08-13
feat: add correct margins for divider

### LINKED ISSUE #2981: DBDivider "margin" prop not clear

- [ ] All
- [ ] HTML
- [X] React
- [ ] Angular
- [ ] Vue
- [ ] Web components
- [ ] Power Apps

### Reproduction case

```jsx
<DBDivider variant="vertical" margin="some_value" />
```

### Expected Behaviour

I'm able to control margin of the component.

### Screenshots

_No response_

### Browser version

None

In general, I'm not sure which values are allowed for the "margin" prop of the DBDivider component. I've tried different spacing values and other variables mentioned here https://db-ui.github.io/mono/version/v0.2.5/foundations/readme and here https://db-ui.github.io/mono/version/v0.2.5/foundations/variables/readme, but nothing worked so far.
Also, typescript allows only two values for that prop: "none" and "_".

Can you please provide an example which values can be used here?

nmerget:
## Proposed changes

closes #2981

## Types of changes

### DIFF
--- a/packages/components/src/shared/model.ts
+++ b/packages/components/src/shared/model.ts
@@ -74,6 +74,15 @@ export type SpacingProps = {
 	 */
 	spacing?: SpacingType;
 };
+export const MarginList = ['medium', 'small', 'large', 'none'] as const;
+export type MarginType = (typeof MarginList)[number];
+
+export type MarginProps = {
+	/**
+	 * The margin attribute changes the margin of the component.
+	 */
+	margin?: MarginType;
+};
 
 export const PlacementList = [
 	'left',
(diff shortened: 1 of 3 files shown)


packages/components/src/shared/model.ts
```diff
@@ -74,6 +74,15 @@ export type SpacingProps = {
...
 };
+export const MarginList = ['medium', 'small', 'large', 'none'] as const;
+export type MarginType = (typeof MarginList)[number];
+
+export type MarginProps = {
+	/**
+	 * The margin attribute changes the margin of the component.
+	 */
+	margin?: MarginType;
+};
```
bruno-sch, line 85: Why so complicated? And not simply MarginType as a union type from the strings?

nmerget: We changed every literal type like this. There was a request from the community that they need the list for iteration.

mfranzke (approved):

PR #3053, opened 2024-08-15
test(guidepup): navigation

bruno-sch:
## Proposed changes

test: adds guidepup test for navigation

## Types of changes


nmerget (dismissed):

showcases/screen-reader/tests/navigation.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+	testDefault({
+		test,
+		title: 'default',
+		description:
+			'should have texts inline or as data-label attributes (next())',
+		url: './#/05/navigation?page=density',
+		async testFn(voiceOver, nvda) {
+			if (nvda) {
+				await nvda?.next(); // Navigation Item 1
+				await nvda?.act(); // Open menu
```
mfranzke, line 15:
```suggestion
				await nvda?.next(); // Navigation Item 2
				await nvda?.previous(); // Navigation Item 1
				await nvda?.act(); // Open menu
```

even though that NVDA and VoiceOver wouldn't be feature equal in this case, we would check whether the navigation is actually hidden from NVDA screenreader in the first place before interacting with the button.

nmerget: Why do we need this? NVDA tells us with `..., menu button, collapsed,...` that the sub-navigation is hidden isn't it?

mfranzke: It tells us that we've set an attribute that would announce `collapsed`, but it doesn't test for whether the navigation is actually visually hidden.

nmerget (dismissed):

mfranzke (approved):

__snapshots__/tab-item/patternhub/tab-item-properties-should-match-screenshot.png
mfranzke, line 1: I think we do have a general problem here.

mfranzke: https://github.com/db-ui/mono/issues/3170

mfranzke (approved):

PR #3082, opened 2024-08-22
test: adds guidepup test for section

bruno-sch:
## Proposed changes

test: adds guidepup test for section

## Types of changes


showcases/screen-reader/__snapshots__/macos/webkit/DBSection-default-1.txt
```diff
@@ -0,0 +1 @@
+["Medium","Large","Large"]
```
bruno-sch, line 1: @mxmehl 

As expected (?), only the inline texts from the sections are output.

I found the following quote: [scroll to "SECTION element"](https://www.accessibilityoz.com/2020/02/html5-sectioning-elements-and-screen-readers/#SECTION-element):
"__Using Section with Aria-LabelledBy, Aria-Label or Title makes screen readers treat the Section element as a landmark,...__"

Or is this more on the side of the consumers? What do you think?

mxmehl: @bruno-sch, did you tag me by mistake? I don't really understand the context and what I could contribute to your question :)

mfranzke: @mxmehl I think so, yes. But you're always welcome and always an enrichment by any contribution ;-)

mfranzke: > @mxmehl
> 
> As expected (?), only the inline texts from the sections are output.
> 
> I found the following quote: [scroll to "SECTION element"](https://www.accessibilityoz.com/2020/02/html5-sectioning-elements-and-screen-readers/#SECTION-element): "**Using Section with Aria-LabelledBy, Aria-Label or Title makes screen readers treat the Section element as a landmark,...**"
> 
> Or is this more on the side of the consumers? What do you think?

@bruno-sch I think that this is perfectly fine. From my understanding, `section` itself is kind of similar to a `div` element, it doesn't provide any specific meaning. Whereas much more semantic elements like e.g. `nav` actually do provide a meaning as a landmark, and due to that fact actually need to get enriched by alternative text to be distinguishable.

So in this case it's perfectly fine that the `section` itself isn't announced at all. Even though we most likely could keep it at two items to test for instead of three.

bruno-sch: @mxmehl opsi :D

showcases/screen-reader/tests/section.spec.ts
```diff
@@ -0,0 +1,23 @@
...
+		async testFn(voiceOver, nvda) {
+			if (nvda) {
+				await nvda?.next();
+			}
+
+			const screenReader = voiceOver ?? nvda;
+			await screenReader?.clearSpokenPhraseLog();
+			await screenReader?.next(); // Section 1
+			await screenReader?.next(); // Section 2
+			await screenReader?.next(); // Section 3
```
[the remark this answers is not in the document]
bruno-sch: removed :)

showcases/screen-reader/__snapshots__/windows/chromium/DBSection-default-1.txt
mfranzke, line 1: even though that the tests are running through smoothly, this doesn't look correct. It should at least provide the texts out of these elements.

bruno-sch: Change

nmerget (approved):

PR #3198, opened 2024-09-16
chore: try to harden screen reader tests to avoid duplication in macOS

nmerget:
## Proposed changes

Sometimes we get something like this as an output for screen reader tests for macOS: 
` [...,"required checked Required TODO: Add a validMessage checkbox. TODO: Add a validMessage",...]`

This PR tries to get rid of the duplicate `TODO: Add a validMessage `

## Types of changes

### DIFF
--- a/showcases/screen-reader/default.ts
+++ b/showcases/screen-reader/default.ts
@@ -53,7 +53,7 @@ const cleanSpeakInstructions = (phraseLog: string[]): string[] =>
 					for (const pPart of phraseParts) {
 						const pPartIndex = phraseParts.indexOf(pPart);
 						// We only use parts with multiple words
-						if (pPart.includes(' ') && part.includes(pPart)) {
+						if (pPart !== part && part.includes(pPart)) {
 							lastFoundIndex = pPartIndex;
 						}
 					}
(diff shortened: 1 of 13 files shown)


mfranzke (dismissed):

showcases/screen-reader/default.ts
```diff
@@ -40,12 +40,32 @@ const flakyExpressions: Record<string, string> = {
...
 				(sPhrase) =>
 					!standardPhrases.some((string) => sPhrase.includes(string))
 			)
+			.map((part, index) => {
+				// There is an issue with macOS duplicating some parts, we remove the duplicates here
+				if (!isWin()) {
+					let lastFoundIndex = -1;
+					for (const pPart of phraseParts) {
+						const pPartIndex = phraseParts.indexOf(pPart);
+						// We only use parts with multiple words
```
mfranzke, line 55: I don't understand this comment. We won't ensure somehow that a part includes multiple words, but only ensure that we take it as a whole instead of splitting it up like we did within one of the previous commits.

mfranzke (dismissed):

mfranzke (approved):

PR #3205, opened 2024-09-17
chore: improve playwright tests by reducing amount of running projects

nmerget:
## Proposed changes

Currently we are testing a11y with every project (chrome, webkit,...) but this isn't required for every project with every variantion.

## Types of changes


showcases/e2e/default.ts
```diff
@@ -19,17 +20,24 @@ export type DefaultSnapshotTestType = {
...
+};
+
 export const waitForDBPage = async (page: Page) => {
 	const dbPage = page.locator('.db-page');
 	// We wait till db-page fully loaded
 	await dbPage.evaluate((element) => {
 		element.style.transition = 'none';
 	});
-	await expect(dbPage).toHaveAttribute('data-fonts-loaded', 'true');
+	await expect(dbPage).not.toHaveAttribute('data-fonts-loaded', 'false');
```
mfranzke, line 40: I think that `await expect(dbPage).toHaveAttribute('data-fonts-loaded', 'true');` is slightly better, because it actually tests for an attribute being set via JS, instead of just it's absence (which might be caused by several different aspects, like e.g. an incorrect implementation / bug of that part).

mfranzke (approved):

mfranzke (approved):

PR #3218, opened 2024-09-22
refactor(accordion): adapt native standard for `behaviour`

mfranzke:
## Proposed changes

Resolves https://github.com/db-ui/mono/issues/1630

## Types of changes


packages/components/src/components/accordion/accordion.lite.tsx
```diff
@@ -88,7 +88,7 @@ export default function DBAccordion(props: DBAccordionProps) {
 
 	onUpdate(() => {
 		if (ref && state._initOpenIndexDone) {
-			if (props.initOpenIndex?.length > 0) {
+			if (props?.initOpenIndex && props.initOpenIndex?.length > 0) {
```
mfranzke, line 91:
```suggestion
			if (props?.initOpenIndex?.length > 0) {
```

shouldn't this this be the equivalent to a chaining like this ?

nmerget: Nope. Some typescript configs will mark this as an error

mfranzke: > Nope. Some typescript configs will mark this as an error

can we remove that "typescript" ?

packages/components/scripts/post-build/angular.ts
```diff
@@ -249,6 +249,10 @@ export default (tmp?: boolean) => {
 				to:
 					'@ViewChild("ref") ref!: ElementRef | undefined;' +
 					getAttributePassing(component.name)
+			},
+			{
+				from: /this.ref.nativeElement/g,
+				to: 'this.ref?.nativeElement'
```
mfranzke, line 255: couldn't we write it like this in the regular code even already, does that conflict with some other framework / library ?

packages/components/src/components/accordion-item/accordion-item.lite.tsx
```diff
@@ -41,23 +41,26 @@ export default function DBAccordionItem(props: DBAccordionItemProps) {
 	// jscpd:ignore-end
 
 	return (
-		<details
-			ref={ref}
+		<li
```
mfranzke, line 44: this would prevent using a simple, standalone `accordion-item`.

packages/components/src/components/accordion-item/accordion-item.lite.tsx
```diff
@@ -41,23 +41,26 @@ export default function DBAccordionItem(props: DBAccordionItemProps) {
...
-				<Show when={!props.headlinePlain}>
-					<Slot name="headline" />
-				</Show>
-			</summary>
-			<div>
-				<Show when={props.content}>{props.content}</Show>
-				<Show when={!props.content}>{props.children}</Show>
-			</div>
-		</details>
+			aria-disabled={props.disabled}>
```
mfranzke, line 47: according to https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-disabled#associated_roles this attribute isn't supposed to get set on a list-item.

packages/components/src/components/accordion-item/accordion-item.lite.tsx
nmerget: Yes, based on our logic `*-item` is always a `<li>` you shouldn't use `accordion-item` without `accordion`

packages/components/src/components/accordion-item/accordion-item.lite.tsx
nmerget: fixed

packages/components/src/components/accordion-item/accordion-item.lite.tsx
```diff
@@ -39,23 +39,24 @@ export default function DBAccordionItem(props: DBAccordionItemProps) {
...
+					<Show when={props.headlinePlain}>
+						{props.headlinePlain}
+					</Show>
+					<Show when={!props.headlinePlain}>
+						<Slot name="headline" />
+					</Show>
+				</summary>
+				<div>
+					<Show when={props.content}>{props.content}</Show>
+					<Show when={!props.content}>{props.children}</Show>
```
mfranzke, line 57:
```suggestion
					<Show when={props.content} else={props.children}>{props.content}</Show>
```

nmerget: fixed

packages/components/src/components/accordion/accordion.lite.tsx
```diff
@@ -35,89 +37,82 @@ export default function DBAccordion(props: DBAccordionProps) {
...
-				Array.from<HTMLDetailsElement>(childDetails).forEach(
-					(details: HTMLDetailsElement) => {
-						details.open = state.openItems.includes(details.id);
+				for (const details of Array.from<HTMLDetailsElement>(
+					childDetails
+				)) {
+					if (state._name === '') {
+						if (details.hasAttribute('name')) {
+							details.removeAttribute('name');
+						}
```
mfranzke, line 80:
```suggestion
						details.removeAttribute('name');
```

mfranzke:
@nmerget ready for review.

nmerget (approved):

PR #3232, opened 2024-09-24
fix: issue with long inline links not breaking

### LINKED ISSUE #3110: Inline links should wrap
Links used inside a paragraph should wrap.

In case of `data-variant="inline"` the  property should change to `white-space: wrap;` and the display should change to `   display: inline;`

https://github.com/db-ui/mono/blob/cc84e5d56beaab44a5fa7a3fc5fd144e8b843769/packages/components/src/styles/_link-components.scss#L43

nmerget:
## Proposed changes

closes #3110

## Types of changes

### DIFF
--- a/packages/components/src/components/link/link.scss
+++ b/packages/components/src/components/link/link.scss
@@ -17,6 +17,11 @@
 		@extend %db-link-height;
 	}
 
+	&[data-variant="inline"] {
+		white-space: inherit;
+		display: inline;
+	}
+
 	&:not([data-variant="inline"]) {
 		&[data-content="internal"] {
 			@include icons.set-icon("arrow_right", "after");
--- a/packages/components/src/components/link/link.scss
+++ b/packages/components/src/components/link/link.scss
@@ -17,11 +17,6 @@
 		@extend %db-link-height;
 	}
 
-	&[data-variant="inline"] {
-		white-space: inherit;
-		display: inline;
-	}
-
 	&:not([data-variant="inline"]) {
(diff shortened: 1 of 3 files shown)


packages/components/src/components/link/link.scss
```diff
@@ -17,6 +17,11 @@
 		@extend %db-link-height;
 	}
 
+	&[data-variant="inline"] {
+		white-space: inherit;
+		display: inline;
+	}
+
```
mfranzke, line 24: What do you think about wrapping the following line instead by a selector that makes this more specific w/o the need to rewrite it afterwards?
https://github.com/db-ui/mono/blob/main/packages/components/src/styles/_link-components.scss#L43

Wrapping it with:
``` CSS
&:not([data-variant="inline"]) {
```

nmerget: done

showcases/react-showcase/src/components/form/index.tsx
mfranzke, line 1: isn't it that we would need to include this components variant on our regular documentation page as well?

mfranzke: https://github.com/db-ui/mono/issues/3237

mfranzke (approved):

showcases/react-showcase/src/components/form/index.tsx
nmerget: It's not in the design specs. We need to discuss this

PR #3251, opened 2024-09-27
chore: pin eslint version

nmerget:
## Proposed changes

Pin `eslint` dependency for dependabot.

`eslint` v9 introduced breaking changes -> most eslint plugins won't work, which we require for our project

## Types of changes


mfranzke:
@nmerget could you please add q quick comment on why we would need the pinning to the description ?

mfranzke (approved):

PR #3400, opened 2024-10-31
docs: enable correct types for properties page in patternhub

nmerget:
## Proposed changes

Add correct types for properties

## Types of changes


packages/components/src/components/link/model.ts
```diff
@@ -16,9 +16,17 @@ export const LinkContentList = ['external', 'internal'] as const;
 export type LinkContentType = (typeof LinkContentList)[number];
 
 export type DBLinkDefaultProps = {
+	/**
+	 * Adds a different arrow to after the link to show external or internal link
```
mfranzke, line 20:
```suggestion
	 * Adds a different arrow after the link to show external or internal link
```

packages/components/src/components/input/model.ts
```diff
@@ -56,6 +62,9 @@ export type DBInputDefaultProps = {
 	 * 	Type of form control
 	 */
 	type?: InputTypeType | string;
+	/**
+	 * Sets step based in [this](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/step).
```
mfranzke, line 66:
```suggestion
	 * Sets [step value](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/step).
```

packages/components/src/components/input/model.ts
```diff
@@ -37,7 +37,13 @@ export const InputTypeList = [
 export type InputTypeType = (typeof InputTypeList)[number];
 
 export type DBInputDefaultProps = {
+	/**
+	 * Set a data list via attributes instead of child.
+	 */
 	dataList?: string[] | ValueLabelType[];
+	/**
+	 * Add a custom id to data list if your using `dataList` attribute.
```
mfranzke, line 45:
```suggestion
	 * Add a custom id to data list if you're using `dataList` attribute.
```

packages/components/src/components/input/model.ts
```diff
@@ -37,7 +37,13 @@ export const InputTypeList = [
 export type InputTypeType = (typeof InputTypeList)[number];
 
 export type DBInputDefaultProps = {
+	/**
+	 * Set a data list via attributes instead of child.
```
mfranzke, line 41:
```suggestion
	 * Set a [data list](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/datalist) via attributes instead of child.
```

packages/components/src/components/header/model.ts
```diff
@@ -9,10 +9,32 @@ import {
...
+	 */
 	metaNavigation?: unknown;
+	/**
+	 * Slot to pass one or more elements like DBButton (e.g. search) as primary action.
+	 * Desktop: Shown next to the main-navigation
+	 * Mobile: Shown next to the brand
+	 */
 	primaryAction?: unknown;
+	/**
+	 * Slot to pass one or more element like DBButton (e.g. profile, language, etc.) as secondary action.
```
mfranzke, line 29:
```suggestion
	 * Slot to pass one or more elements like DBButton (e.g. profile, language, etc.) as secondary action.
```

packages/components/src/components/header/model.ts
```diff
@@ -9,10 +9,32 @@ import {
...
+	 * Desktop: Shown seperated by divider at the end of the header
+	 * Mobile: Shown inside the drawer at the bottom.
+	 */
 	secondaryAction?: unknown;
+
+	/**
+	 * Open/closes the drawer on for mobile header or if `forceMobile` is true;
```
mfranzke, line 36:
```suggestion
	 * Open/closes the drawer for mobile header or if `forceMobile` is true.
```

packages/components/src/components/card/model.ts
```diff
@@ -18,6 +18,9 @@ export type DBCardDefaultProps = {
 	 */
 	behaviour?: CardBehaviourType;
 
+	/**
+	 * Changes the elevation of the card which is equal to basic-background-level
```
mfranzke, line 22:
```suggestion
	 * Changes the elevation of the card which is equal to `basic-background-level`
```

packages/components/package.json
```diff
@@ -17,8 +17,8 @@
   "scripts": {
     "build": "npm-run-all -p build-components build-assets -s build-style:*",
     "build-assets": "cpr src build -o -f \"(.ts|.tsx|.md|.html)$\"",
-    "build-components": "npm-run-all build:mitosis -p build-components:*",
-    "build-components:docs": "react-docgen ../../output/react/src/components/**/*.tsx -o ../../output/docs.json -i *.spec.tsx",
+    "build-components": "npm-run-all build:mitosis build-components:post build-components:docs",
```
mfranzke, line 20: do all of those need to run sequentially, or could some run in parallel ?

output/stencil/scripts/packageLinkPhase.js
```diff
@@ -46,11 +46,11 @@ const getFilteredContextData = (data) => {
 /**
  * Resolves props by data
  * @param resolvedData {object}
- * @param value {object}
+ * @param value {object|string}
  * @return {{resolvedType, name, type}}
  */
 const resolveProp = (resolvedData, value) => {
-	const { type } = value;
+	let type = value.type ? value.type : value;
```
mfranzke, line 53:
```suggestion
	let type = value.type || value;
```

packages/components/src/shared/model.ts
```diff
@@ -204,7 +207,7 @@ export const CustomValidityList = [
 export type CustomValidityType = (typeof CustomValidityList)[number];
 export type FormProps = {
 	/**
-	 * Marks an input element as invalid (red) | valid(green) | no-validation(grey). Overwrites the :user-valid selector.
+	 * Marks an input element as invalid (red) / valid(green) / no-validation(grey). Overwrites the :user-valid selector.
```
mfranzke, line 210:
```suggestion
	 * Marks an input element as invalid (red) / valid (green) / no-validation (grey). Overwrites the :user-valid selector.
```

packages/components/src/shared/model.ts
```diff
@@ -405,16 +408,49 @@ export const LinkReferrerPolicyList = [
...
+	/**
+	 * Sets aria attribute based on [`aria-current`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-current).
+	 */
 	current?: boolean | LinkCurrentType;
+	/**
+	 * Disables the link.
+	 */
 	disabled?: boolean;
+	/**
+	 * The URL that the hyperlink points to. See [this](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#href).
```
mfranzke, line 420:
```suggestion
	 * The URL that the hyperlink points to. See [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#href).
```

packages/components/src/shared/model.ts
```diff
@@ -405,16 +408,49 @@ export const LinkReferrerPolicyList = [
...
+	 */
 	href?: string;
+	/**
+	 * Hints at the human language of the linked URL. See [this](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#hreflang).
```
mfranzke, line 424:
```suggestion
	 * Hints at the human [language of the linked URL](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#hreflang).
```

packages/components/src/shared/model.ts
```diff
@@ -405,16 +408,49 @@ export const LinkReferrerPolicyList = [
...
+	 */
 	hreflang?: string;
+	/**
+	 * Sets aria attribute based on [`aria-label`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label).
+	 */
 	label?: string;
+	/**
+	 * Where to display the linked URL, as the name for a browsing context. See [this](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#target).
```
mfranzke, line 432:
```suggestion
	 * Where to open the linked URL, as the name for a [browsing context](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#target).
```

showcases/patternhub/components/old-routing-fallback/index.tsx
```diff
@@ -32,7 +32,8 @@ const OldRoutingFallback = () => {
 				}
 
 				// This is for the old implementation to work with iframes
-				const foundRoute = getAllNavigationItems().find((item) =>
+				const allNavigationItems = getAllNavigationItems();
+				const foundRoute = allNavigationItems.find((item) =>
```
mfranzke, line 36: what's the reason for this change ?

nmerget: just for debugging, to set a breakpoint on line 36

packages/components/package.json
nmerget: Yes the docs may be affected by the post-build script. The post-build takes only some seconds so this shouldn't be a gib performance improvement

__snapshots__/brand/patternhub/brand-properties-should-match-screenshot.png
mfranzke, line 1: why is describedbyid removed? Did it become a global attribute ?

output/stencil/scripts/packageLinkPhase.js
nmerget: fixed

__snapshots__/brand/patternhub/brand-properties-should-match-screenshot.png
nmerget: I generate the `properties.mdx` out from `web-types.json` instead of `react-docgen`. 

The web-component anaylzer uses only props which are included inside the component. While some props e.g. `describedbyid` are inside the `GlobalProps`, but not used in the component `DBBrand`.

mfranzke (approved):

mfranzke (approved):

PR #3465, opened 2024-11-14
feat: Figma properties alignment breaking

nmerget:
## Types of changes


__snapshots__/header/patternhub/header-overview-should-match-screenshot.png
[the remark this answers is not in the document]
nmerget: screenshots weren't updated

README.md
```diff
@@ -100,6 +100,8 @@ Check our migration docs for breaking changes:
 1. [Alpha ➡ Beta](https://github.com/db-ui/mono/tree/main/docs/migration/alpha-beta.md) (0.0.x➡0.x.x)
 2. [v0.2 ➡ v0.3](https://github.com/db-ui/mono/tree/main/docs/migration/v0.2.x-to-v0.3.x.md)
 3. [v0.3 ➡ v0.4](https://github.com/db-ui/mono/tree/main/docs/migration/v0.3.x-to-v0.4.x.md)
+4. [v0.4 ➡ v0.5](https://github.com/db-ui/mono/tree/main/docs/migration/v0.4.x-to-v0.5.x.md)
```
mfranzke, line 103:
```suggestion
4. v0.4 ➡ v0.5: no migration needed, no breaking changes
```

docs/migration/v0.4.x-to-v0.5.x.md
mfranzke, line 1: let's delete this file

__snapshots__/accordion-item/patternhub/accordion-item-properties-should-match-screenshot.png
mfranzke, line 1: @nmerget this screenshot now includes both a "TODO" as well as a "No description" entry within the description column. This is most likely not intended, is it ?

packages/components/src/components/select/select.scss
```diff
@@ -43,7 +43,7 @@ $has-before-padding: calc(
 		color: colors.$db-adaptive-on-bg-basic-emphasis-100-default;
 	}
 
-	&[data-variant="hidden"] {
+	&[data-hide-label="true"] {
```
mfranzke, line 46: @nmerget isn't it, that we even also need to mention this aspect within the https://github.com/db-ui/mono/blob/3e026a9d28075bb1ba634f6c8d53acc1540e4bca/docs/migration/v0.5.x-to-v0.6.x.md file ?

nmerget: It is named `showLabel="true/false"` in design. But the label is the default so I inverted the internal `data-` attribute. The property will be `showLabel` as we see it in the migration guide

mfranzke: but than you could still make this `data-show-label="false"`.

mfranzke (approved):

PR #3476, opened 2024-11-18
refactor: using local font references again

mfranzke:
## Proposed changes

Let's activate the local references again as soon as the local fonts are repaired.

## Types of changes


mfranzke:
the local files are still not correct.

PR #3477, opened 2024-11-18
refactor: removed inner div inside `section.lite.tsx`

nmerget:
## Proposed changes

Removed inner `<div>` from `DBSection` to handle proper layouting.

## Types of changes


mfranzke (dismissed):

packages/components/src/utils/index.ts
```diff
@@ -224,6 +224,21 @@ export const enableCustomElementAttributePassing = (
...
+				const isWebComponent = attr.value.includes('hydrated');
+				const value = attr.value.replace('hydrated', '').trim();
+				const currentClass = element.getAttribute('class');
+				element.setAttribute(
+					attr.name,
+					`${currentClass ? `${currentClass} ` : ''}${value}`
+				);
+				if (isWebComponent) {
+					// Stencil is using this class for lazy loading component
+					parent.setAttribute('class', 'hydrated');
```
mfranzke, line 237:
```suggestion
					parent.classList.add('hydrated');
```

nmerget: I need to replace everything inide `class` at this point

mfranzke: > I need to replace everything inide `class` at this point

I even already wondered about this aspect, but as you explained it, that totally makes sense to me.

mfranzke (approved):

PR #3500, opened 2024-11-21
feat: add headline sizes as css class to overwrite headlines on purpose

nmerget:
## Proposed changes

We were missing a way to overwrite the default headline font sizes. Needed for layouts with a semantic e.g. `h2` but with a smaller font-size/line-height.

## Types of changes


mfranzke:
what about introducing the headline component instead ?

nmerget:
> @nmerget what about introducing the headline component instead ?

This would be something in cooperation with design. - I guess I wront an internal ticket at some point
We need some solution for now :)

mfranzke (approved):

PR #3527, opened 2024-11-28
docs: disable auto-dark mode for iframes

nmerget:
## Proposed changes

Disable auto dark-mode for Marketingportal iframes

## Types of changes

### DIFF
--- a/showcases/patternhub/styles/globals.scss
+++ b/showcases/patternhub/styles/globals.scss
@@ -7,6 +7,10 @@
 @forward "@db-ui/foundations/build/css/colors/classes/all";
 @forward "@db-ui/foundations/build/css/fonts/classes/all";
 
+:root:not(:has(.db-page)) {
+	color-scheme: light;
+}
+
 body:not(:has(.db-page)) {
 	padding-block-end: variables.$db-spacing-fixed-lg;
 }


showcases/patternhub/styles/globals.scss
```diff
@@ -7,6 +7,10 @@
 @forward "@db-ui/foundations/build/css/colors/classes/all";
 @forward "@db-ui/foundations/build/css/fonts/classes/all";
 
+:root:not(:has(.db-page)) {
```
mfranzke, line 10: The problem with this selector: not all of our consuming developers (99%) would use DBPage component in use (especially as it's undocumented).

nmerget: I don't understand this comment.

This is only for patternhub. Every site we deliver as iframe doesn't have .db-page -> this forces light-mode.
The consuming developers aren't affected by this change. If they head over to https://db-ui.github.io/mono/review/main/ they would use their prefered scheme

mfranzke: sorry, I haven't been looking at the filename. Pardon moi.

mfranzke (approved):

PR #3529, opened 2024-11-28
fix: replaced JavaScript selector

mfranzke:
## Proposed changes

Resolves https://github.com/db-ui/mono/issues/3528

Whereas `&` would be perfectly fine at this place to be used within CSS, when executing `.querySelector` or `.querySelectorAll` in JavaScript we would need to use [`:scope`](https://drafts.csswg.org/selectors-4/#the-scope-pseudo) instead, that gets it's [scoping root](https://drafts.csswg.org/selectors-4/#scoping-root) by the documents part on which `.querySelector` or `.querySelectorAll` are based on (the `ref` in our case).

## Types of changes

### DIFF
--- a/packages/components/src/components/tabs/tabs.lite.tsx
+++ b/packages/components/src/components/tabs/tabs.lite.tsx
@@ -119,7 +119,7 @@ export default function DBTabs(props: DBTabsProps) {
 
 				const tabPanels = Array.from<Element>(
 					ref.querySelectorAll(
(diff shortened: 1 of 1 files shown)


mfranzke:
@biancavo could you please try out whether prerelease version _0.6.0-0_ of the `@db-ui/v-components` packages solves the problem in your Vitest setup ?

satistl:
@mfranzke I just tested the fix and it solves the problem with the vitest setup 👍

nmerget (approved):

PR #3530, opened 2024-11-28
refactor(tabs): ensure valid tab `id` values

mfranzke:
## Proposed changes

We need to ensure that tabs only include a valid tab value, whereas the current implementation might start with e.g. a number.

image

## Types of changes


nmerget:
Shouldn't we do it for tab-panels as well?

mfranzke:
> Shouldn't we do it for tab-panels as well?

of course. My world is just very small, or in other words, my blinder very big.

nmerget (approved):

PR #3563, opened 2024-12-06
refactor: remove brand fonts

mfranzke:
## Types of changes


mfranzke:
I do see three different options of how to proceed with this:
- Use DB Non Latin
- Use any general "safe" font definition like e.g. _Helvetica_.
- Use Google Noto.

To my understanding we've agree on the first option, the second would be make sense as well strategic-wise. But we shouldn't proceed with the third, as especially the first one would be a simpler one, and second more accurate to the expected result in companies context usage.

PR #3617, opened 2024-12-19
fix: minor issue with checkbox layout

nmerget:
## Proposed changes

Align layout with design

## Types of changes


packages/components/src/components/checkbox/checkbox.scss
```diff
@@ -32,7 +32,7 @@
 	input {
 		display: flex;
 		border-radius: variables.$db-border-radius-2xs;
-		margin-block: auto;
+		margin-block: 0 auto;
```
mfranzke, line 35:
```suggestion
```

What for exactly do we have this whole declaration anyhow?

mfranzke (approved):

PR #3647, opened 2025-01-07
feat: noto sans eu

nmerget:
## Proposed changes

replaced current `Noto-Sans` with optimized font for EU

## Types of changes


packages/foundations/assets/fonts/NotoSans-Black.ttf
mfranzke, line 1: @nmerget do we actually need those TTF files in here ?

packages/foundations/assets/fonts/README.md
```diff
@@ -0,0 +1,23 @@
+# Generate fonts
+
+To generate optimal fonts we use [fonttools](https://github.com/fonttools/fonttools). To use the tools you need python installed:
+
+1. Install [python](https://www.python.org/downloads/)
```
mfranzke, line 5:
```suggestion
1. Install [python](https://docs.python-guide.org/starting/installation/#installation)
```

I'm thinking about providing a link to a more generic description by not directing the users to (even though official) download links, but a description, that would even also mention "the right way"™️ by using homebrew, etc.

packages/foundations/assets/fonts/README.md
```diff
@@ -0,0 +1,23 @@
...
+2. Install fonttools: `pip3 install fonttools`
+3. Install brotli: `pip3 install brotli`
+
+Afterward, you can generate a font by running the following command:
+
+```shell
+pyftsubset NotoSans-Regular.ttf --layout-features=* --flavor=woff2 --unicodes-file=unicode-eu.txt --output-file=NotoSans-EU-Regular.woff2
```
mfranzke, line 12:
```suggestion
pyftsubset NotoSans-Regular.ttf --layout-features=* --flavor=woff2 --unicodes-file=unicode-eu.txt --output-file=NotoSans-Regular-EU.woff2
```

packages/foundations/assets/fonts/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+## Local development
+
+You can use `packages/foundations/assets/fonts/generate-eu-fonts.ts` to generate all fonts inside this directory. To do so, run the following command:
+
+```shell
+node packages/foundations/assets/fonts/generate-eu-fonts.ts
+```
+
+To check if everything works fine you can check all glyphs [here](https://opentype.js.org/glyph-inspector.html).
```
mfranzke, line 23:
```suggestion
To check if everything works fine you can check all glyphs with an [opentype.js.org glyph inspector tool](https://opentype.js.org/glyph-inspector.html).
```

packages/foundations/assets/fonts/generate-eu-fonts.ts
```diff
@@ -0,0 +1,40 @@
+import { glob } from 'glob';
+import { exec } from 'child_process';
```
mfranzke, line 2:
```suggestion
import { exec } from 'node:child_process';
import { glob } from 'glob';
```

packages/foundations/assets/fonts/generate-eu-fonts.ts
```diff
@@ -0,0 +1,40 @@
...
+import { promisify } from 'util';
```
mfranzke, line 3:
```suggestion
import { exec } from 'node:child_process';
import { promisify } from 'node:util';
import { glob } from 'glob';
```

packages/foundations/assets/fonts/NotoSans-Black.ttf
mfranzke: yes, for regeneration of the fonts

mfranzke: @nmerget but couldn't we even also use woff2 as an input instead of ttf files, so that we won't need to provide those?

mfranzke: > pyftsubset is an OpenType font subsetter and optimizer, based on fontTools. It accepts any TT- or CFF-flavored OpenType (.otf or .ttf) or WOFF (.woff) font file. The subsetted glyph set is based on the specified glyphs or characters, and specified OpenType layout features.

source: https://fonttools.readthedocs.io/en/stable/subset/index.html#module-fontTools.subset:~:text=pyftsubset%20is%20an%20OpenType%20font%20subsetter%20and%20optimizer%2C%20based%20on%20fontTools.%20It%20accepts%20any%20TT%2D%20or%20CFF%2Dflavored%20OpenType%20(.otf%20or%20.ttf)%20or%20WOFF%20(.woff)%20font%20file.%20The%20subsetted%20glyph%20set%20is%20based%20on%20the%20specified%20glyphs%20or%20characters%2C%20and%20specified%20OpenType%20layout%20features.

mfranzke: in this case, instead of using (and providing) the `ttf` files within this repo, by using the `woff` files we could instead use and provide those, which would be beneficial for users that still want to include those in their codebase for downwards compatibility.

nmerget: I included them, because they are the base for the fonts.

If users need some other subset, they could use them as a reference. It doesn't matter if it is `.ttf` or `.woff` because they wouldn't use them inside their appication anyways.

mfranzke (approved):

Issue #3669, opened 2025-01-14
v3 textarea readonly + disabled: border is cut off


leape:
### Which generators are impacted?

- [X] All
- [ ] HTML
- [ ] React
- [ ] Angular
- [ ] Vue
- [ ] Web components
- [ ] Power Apps

### Reproduction case

right top border is cut off for readonly = true and disabled = true
https://db-ui.github.io/mono/review/main/components/data-input/textarea/overview

### Expected Behaviour

border is displayed correctly

### Screenshots

![image](https://github.com/user-attachments/assets/d0467e3f-22b6-4dcd-b532-a5664380772a)

### Browser version

Chrome

PR #3670, opened 2025-01-14
fix(textarea): scope overflow declaration

mfranzke:
## Proposed changes

Resolves https://github.com/db-ux-design-system/core-web/issues/3669

## Types of changes


packages/components/src/components/textarea/textarea.scss
```diff
@@ -43,8 +43,10 @@
 		@extend %resize-textarea;
 		@extend %scrollbar;
 
-		cursor: auto;
-		overflow-y: scroll;
+
+		&:not(:read-only, :disabled) {
```
nmerget, line 47: image

That's not the correct solution, you should be able to scroll

mfranzke: you're right.

nmerget:
https://css-tip.com/overflow-detection/ <- we could check out this as well, but it is nice to have

mfranzke:
> https://css-tip.com/overflow-detection/ <- we could check out this as well, but it is nice to have

extracted this one into https://github.com/db-ux-design-system/core-web/issues/3922

PR #3671, opened 2025-01-14
chore: replace mitosis generated vue output (options-api) with composition-api

nmerget:
## Proposed changes

Moved from `Vue` option-api to composition-api

## Types of changes


packages/components/src/components/drawer/drawer.scss
mfranzke, line 1: why do you nest `prefers-reduced-motion` @media queries into each other ?

packages/components/src/components/tooltip/tooltip.lite.tsx
```diff
@@ -47,12 +49,12 @@ export default function DBTooltip(props: DBTooltipProps) {
 					);
 				});
 				parent.setAttribute('data-has-tooltip', 'true');
-				parent.setAttribute('aria-describedby', state._id);
+				parent.setAttribute('aria-describedby', state._id || '');
```
mfranzke, line 52: this doesn't take into account an existing value, does it ?

packages/components/src/components/drawer/drawer.scss
nmerget: I guess this is some auto `--fix` from `stylelint`

packages/components/src/components/tooltip/tooltip.lite.tsx
nmerget: fixed

mfranzke (approved):

PR #3715, opened 2025-01-21
Fix: Fixed broken hrefs urls

leonmarkacz:
## Proposed changes

Updated the URLs in the documentation files to replace /blob/ with /tree/ for consistency and to prevent potential navigation issues. These changes ensure the links point to the correct paths for better user experience.

## Types of changes

- [x] Documentation Update (if none of the other choices apply)

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,13 +155,13 @@ If you're working as a developer on the DB UX Design System, please have a look 
 
 ## Contributions
 
-Contributions are very welcome, please refer to the [contribution guide](https://github.com/db-ui/mono/blob/main/CONTRIBUTING.md).
+Contributions are very welcome, please refer to the [contribution guide](https://github.com/db-ui/mono/tree/main/CONTRIBUTING.md).
 
 ## Code of conduct
 
 We as members, contributors, and leaders pledge to make participation in our
 community a harassment-free experience for everyone – have a look at
-our [Contributor Covenant Code of Conduct](https://github.com/db-ui/mono/blob/main/CODE-OF-CONDUCT.md).
+our [Contributor Covenant Code of Conduct](https://github.com/db-ui/mono/tree/main/CODE-OF-CONDUCT.md).
 
 ## License
 
--- a/docs/migration/alpha-beta.md
+++ b/docs/migration/alpha-beta.md
@@ -76,4 +76,4 @@ This is related to the following properties:
 We add some more information about our styling and try to generate classes and data-attributes to use in the project, based on user-preferences.
 Moreover, we add all optional styles to `db-ui-42` which may increase the size, but reduces the complexity for using the Design-System with all features.
 
-If you encounter performance issues try to reduce your loaded styles with this [documentation](https://github.com/db-ui/mono/blob/main/packages/components/README.md#optimize-dependencies) or by using a tool like [purgecss](https://purgecss.com/).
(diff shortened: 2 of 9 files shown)


mfranzke:
@leonmarkacz thanks a lot for your contribution !

I wasn't aware of the exact differences and needed to ask ChatGPT – could you please elaborate a little but further why you suggest using `tree` over `blob`?

## ChatGPTs answer:

When referencing URLs on GitHub, the parts `/tree/main/` and `/blob/main/` serve different purposes, and whether one is better or more stable depends on your use case. Here's a breakdown of both:

### `/tree/main/`
- **Purpose**: Refers to a directory structure in the repository.
- **Example**: `https://github.com/username/repository/tree/main/`
- **Use case**: When you're pointing to the root or any folder in the repository.
- **Stability**: It's stable in the sense that the directory structure itself doesn’t change unless the repository is altered. However, the contents of the directory (i.e., which files it contains) can change depending on commits and merges.
- **Why use it**: This URL is useful when you're referencing a directory, such as pointing users to a folder where certain files are located.

### `/blob/main/`
- **Purpose**: Refers to a specific file in the repository at a given commit.
- **Example**: `https://github.com/username/repository/blob/main/path/to/file`
- **Use case**: When you're referencing a particular file and its contents at a specific point in the repository (i.e., a snapshot of the file in a particular commit on the `main` branch).
- **Stability**: It's stable as long as the file exists in the `main` branch. However, if the file is renamed, moved, or deleted, this URL will break. It also points to a specific version of the file based on the commit you're referring to.
- **Why use it**: This URL is ideal when you need to directly reference or share a file in the repository.

### Which one to choose?
- If you want to **reference a file's contents**, use the `/blob/main/` URL. It's perfect for sharing a file directly.
- If you want to **reference a directory or structure**, use `/tree/main/`. It's great for referencing or browsing a folder in the repository.

### Stability considerations:
- Neither URL is *completely stable* across all versions of the repository. If files or directories change, the URLs can break.
- To make links more stable, it's generally better to **link to a specific commit hash** (e.g., `/commit/{commit-hash}/`), especially in cases where stability across versions is critical. This way, even if the branch name changes (e.g., `main` to `master`), the URL will still refer to that exact state of the repository.

### Example:
- **`/tree/main/`**: Link to the directory: `https://github.com/username/repository/tree/main/src`
- **`/blob/main/`**: Link to the file: `https://github.com/username/repository/blob/main/src/index.js`

In summary:
- Use `/tree/main/` for directories.
- Use `/blob/main/` for files.
- If stability is important, consider linking to a specific commit, e.g., `/commit/{commit-hash}/`.

leonmarkacz:
Thank you for taking the time to review my PR. I now realize that I made a mistake - blob is indeed the correct prefix for files, while tree is used for directories. I apologize for the oversight and have closed the PR accordingly.

Thank you for making this UI library publicly available. It’s a great project, and I look forward to seeing how it grows

mfranzke:
> Thank you for taking the time to review my PR. I now realize that I made a mistake - blob is indeed the correct prefix for files, while tree is used for directories. I apologize for the oversight and have closed the PR accordingly.
> 
> Thank you for making this UI library publicly available. It’s a great project, and I look forward to seeing how it grows

all good, thanks for your support and your feedback !

PR #3819, opened 2025-02-14
fix: issues for new sass versions

nmerget:
## Types of changes


.stylelintrc.json
```diff
@@ -32,6 +32,7 @@
 		"a11y/selector-pseudo-class-focus": true,
 		"csstools/use-logical": ["always", { "except": ["float"] }],
 		"max-nesting-depth": null,
+		"scss/dollar-variable-colon-space-after": null,
```
mfranzke, line 35: why do we need to deactivate a code style in total ?

mfranzke: https://github.com/db-ux-design-system/core-web/issues/3820

nmerget: prettier rules over stylelint

mfranzke: Oh no !

PR #3880, opened 2025-02-28
fix: issue for angular and vue triggered "checked" instead of "value"

nmerget:
## Proposed changes

Currently the `ngModel` informs about `checked` for radio which makes no sense. It should inform the user with the `value` 

## Types of changes


mfranzke:
@nmerget wouldn't this be a BREAKING CHANGE ?

mfranzke (approved):

PR #3889, opened 2025-03-03
docs(develoment): added relevant aspect

mfranzke:
## Proposed changes

we actually need to mention the important environmental variable setup.

## Types of changes

- [x] Documentation Update (if none of the other choices apply)


docs/development.md
```diff
@@ -2,6 +2,10 @@
 
 ### Start developing
 
+You'll need to insert the environmental variables as described within the package [`@db-ux/db-theme`](https://www.npmjs.com/package/@db-ux/db-theme) first of all.
+
+Afterwards run the following commands:
+
```
mfranzke, line 8: @nmerget is this how you would expect it to be the best solution, have I gotten this right ?

PR #3929, opened 2025-03-11
refactor: we could just remove those depreacted rules

mfranzke:
## Proposed changes

"removing the deprecated rules from your configuration object", source "removing the deprecated rules from your configuration object", source https://github.com/stylelint/stylelint/blob/main/docs/migration-guide/to-15.md#deprecated-stylistic-rules

## Types of changes

### DIFF
--- a/.stylelintrc.json
+++ b/.stylelintrc.json
@@ -49,24 +49,6 @@
 				"ignoreAnnotations": ["default", "global"]
 			}
 		],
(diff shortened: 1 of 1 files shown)


nmerget:
We cannot remove this because the `@stylistic` rules comming from this https://github.com/bjankord/stylelint-config-sass-guidelines/blob/main/index.js#L61 config

PR #3939, opened 2025-03-13
fix: issues with form controls not showing proper valid state

### LINKED ISSUE #3096: Option to disable validation message

- [X] All
- [ ] HTML
- [ ] React
- [ ] Angular
- [ ] Vue
- [ ] Web components
- [ ] Power Apps

### Reproduction case

Use an input field with certain attributes set that enable validation, e.g. a numeric field with min/max set.

### Expected Behaviour

It would be nice to have an option to disable the validation message entirely (incl. the icon), since displaying it increases the height of the input component and in multi-line forms, this will shift the layout around.

Validation could still take place, and the input field could be coloured according to the result, but there would be no message box underneath (and thus, no change in component height). The validation message could e.g. instead be show as a tooltip, maybe?

### Screenshots

image

### Browser version

None

_No response_

nmerget:
## Proposed changes

closes #3096

## Types of changes


packages/components/src/styles/internal/_form-components.scss
```diff
@@ -288,7 +284,9 @@ $input-valid-types:
 	@include set-required-label($selector);
 
 	@include get-validity($selector) {
-		@include get-validity-color($selector, "valid");
+		&:has(.db-infotext[data-semantic="successful"]) {
+			@include get-validity-color($selector, "valid");
+		}
```
mfranzke, line 289: why can't we still color the field in the valid color even if not showing a visual textual "valid" message ?

packages/components/src/styles/internal/_form-components.scss
```diff
@@ -445,7 +445,14 @@ $input-valid-types:
 	@include set-required-label(input);
 
 	@include get-validity($selector) {
-		@include get-validity-color-check("valid");
+		@if ($selector == check) {
+			&:has(.db-infotext[data-semantic="successful"]) {
```
mfranzke, line 449: why can't we still color the fields border in the valid color even if not showing a visual textual "valid" message ?

packages/components/src/styles/internal/_form-components.scss
nmerget: We shouldn't do it. The border as a signal alone wouldn't be accessible for all users with color issues (contrast mode etc.). That's the reason why we need to show a success icon/message as well.

packages/components/src/styles/internal/_form-components.scss
nmerget: https://github.com/db-ux-design-system/core-web/pull/3939#discussion_r1994960255

packages/components/src/styles/internal/_form-components.scss
mfranzke: > And it would be inconsitent between design and dev

we wouldn't color the border in design w/o showing a success message ?

nmerget: yes

packages/components/src/components/select/select.lite.tsx
```diff
@@ -262,20 +277,20 @@ export default function DBSelect(props: DBSelectProps) {
...
-				size="small"
-				semantic="successful">
-				{props.validMessage ?? DEFAULT_VALID_MESSAGE}
-			</DBInfotext>
+			<Show when={state.hasValidState()}>
+				<DBInfotext
+					id={state._validMessageId}
+					size="small"
+					semantic="successful">
+					{props.validMessage || DEFAULT_VALID_MESSAGE}
```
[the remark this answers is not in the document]
nmerget: We want to overwrite an empty string in this case

packages/components/src/components/checkbox/checkbox.lite.tsx
```diff
@@ -211,21 +220,20 @@ export default function DBCheckbox(props: DBCheckboxProps) {
...
-				size="small"
-				semantic="successful">
-				{props.validMessage ?? DEFAULT_VALID_MESSAGE}
-			</DBInfotext>
+			<Show when={state.hasValidState()}>
+				<DBInfotext
+					id={state._validMessageId}
+					size="small"
+					semantic="successful">
+					{props.validMessage || DEFAULT_VALID_MESSAGE}
```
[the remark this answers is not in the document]
nmerget: We want to overwrite an empty string in this case

packages/components/src/components/textarea/textarea.lite.tsx
```diff
@@ -198,21 +211,20 @@ export default function DBTextarea(props: DBTextareaProps) {
...
-				size="small"
-				semantic="successful">
-				{props.validMessage ?? DEFAULT_VALID_MESSAGE}
-			</DBInfotext>
+			<Show when={state.hasValidState()}>
+				<DBInfotext
+					id={state._validMessageId}
+					size="small"
+					semantic="successful">
+					{props.validMessage || DEFAULT_VALID_MESSAGE}
```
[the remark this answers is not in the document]
nmerget: We want to overwrite an empty string in this case

mfranzke (approved):

PR #3940, opened 2025-03-13
fix: issues with boolean & numbers as strings for web-component frameworks (Stencil, Angular)

nmerget:
## Types of changes


mfranzke:
@nmerget is there an issue that this PR resolves ?

packages/components/src/components/input/input.lite.tsx
```diff
@@ -193,22 +195,25 @@ export default function DBInput(props: DBInputProps) {
...
-				max={props.max}
-				min={props.min}
-				readOnly={props.readOnly}
+				maxLength={getNumber(props.maxLength, props.maxlength)}
+				minLength={getNumber(props.minLength, props.minlength)}
+				max={getNumber(props.max)}
+				min={getNumber(props.min)}
+				readOnly={
+					getBooleanAsString(props.readOnly) ||
+					getBooleanAsString(props.readonly)
```
mfranzke, line 208: I don't get the reason why to include an expression like this, line 207 and 208 are the same method call, aren't they ?

packages/components/src/components/textarea/textarea.lite.tsx
```diff
@@ -161,12 +163,15 @@ export default function DBTextarea(props: DBTextareaProps) {
...
 				id={state._id}
 				data-resize={props.resize}
-				disabled={props.disabled}
-				required={props.required}
-				readOnly={props.readOnly}
+				disabled={getBooleanAsString(props.disabled)}
+				required={getBooleanAsString(props.required)}
+				readOnly={
+					getBooleanAsString(props.readOnly) ||
+					getBooleanAsString(props.readonly)
```
mfranzke, line 170: I don't get the reason why to include an expression like this, line 169 and 170 are the same method call, aren't they ?

packages/components/src/shared/model.ts
```diff
@@ -285,15 +285,27 @@ export type FormTextProps = {
...
+	 */
+	maxlength?: number | string;
+	/**
+	 * Minimum length (number of characters) of value
+	 */
+	minlength?: number | string;
+	/**
+	 * The disabled attribute can be set to keep a user from edit on the form element
+	 */
+	readOnly?: boolean | string;
```
mfranzke, line 304: Isn't it, that those are included redundant ?

mfranzke, line 304: Isn't it, that those are included redundant ? Maybe there is a good reason that I don't get at the moment.

showcases/react-showcase/src/components/switch/index.tsx
```diff
@@ -20,7 +24,7 @@ const getSwitch = ({
 	<>
 		<DBSwitch
 			visualAid={visualAid}
-			defaultChecked={checked}
+			defaultChecked={getBooleanAsString(checked)}
```
mfranzke, line 27: wouldn't you need `getBooleanAsString` on the upfollowing `disabled` and `required` attributes as well?

showcases/react-showcase/src/components/radio/index.tsx
```diff
@@ -22,7 +26,7 @@ const getRadio = ({
 			label={label}
 			size={size}
 			name={name}
-			defaultChecked={checked}
+			defaultChecked={getBooleanAsString(checked)}
```
mfranzke, line 29: wouldn't you need `getBooleanAsString` on the upfollowing `disabled` and `required` attributes as well?

showcases/react-showcase/src/components/checkbox/index.tsx
```diff
@@ -25,7 +29,7 @@ const getCheckbox = ({
 			label={label}
 			size={size}
 			name={name}
-			defaultChecked={checked}
+			defaultChecked={getBooleanAsString(checked)}
```
mfranzke, line 32: wouldn't you need `getBooleanAsString` on the upfollowing `disabled`, `required` and `indeterminate` attributes as well?

nmerget:
> @nmerget is there an issue that this PR resolves ? (there doesn't have to be one, but in case it does, we should link it)

No there is no issue, was a finding from a conversation

packages/components/src/components/input/input.lite.tsx
nmerget: We have different ways between the frameworks to pass Properties.

While JSX uses camelCase there are frameworks like Angular which uses everything in lowercase. The problem is that the TS Suggestions still show the lowercase property besides the other one for Angular projects. This leads to errors because you can pass any property, but we wouldn't consume it.

The duplication is a quick and easy way to provide the correct syntax for every framework

packages/components/src/components/textarea/textarea.lite.tsx
nmerget: https://github.com/db-ux-design-system/core-web/pull/3940#discussion_r1994964883

packages/components/src/shared/model.ts
nmerget: https://github.com/db-ux-design-system/core-web/pull/3940#discussion_r1994964883

showcases/react-showcase/src/components/switch/index.tsx
nmerget: No `disabled `and `required` have the correct type because they are 1:1 our type. The `defaultChecked` is a React property which is defined as boolean | undefined

showcases/react-showcase/src/components/radio/index.tsx
nmerget: https://github.com/db-ux-design-system/core-web/pull/3940#discussion_r1994966860

showcases/react-showcase/src/components/checkbox/index.tsx
nmerget: https://github.com/db-ux-design-system/core-web/pull/3940#discussion_r1994966860

packages/components/src/components/input/input.lite.tsx
mfranzke: Got it. I overlooked the tiny difference in the parameter.

packages/components/src/components/checkbox/checkbox.lite.tsx
```diff
@@ -144,14 +146,14 @@ export default function DBCheckbox(props: DBCheckboxProps) {
 					) {
 						// When indeterminate is set, the value of the checked prop only impacts the form submitted values.
 						// It has no accessibility or UX implications. (https://mui.com/material-ui/react-checkbox/)
-						_ref.indeterminate = props.indeterminate;
+						_ref.indeterminate = !!getBoolean(props.indeterminate);
```
mfranzke, line 149: I don't get the reason for this notation, `getBoolean` even already returns a Boolean. Or is it related to some framework specifics or build-script ?

packages/components/src/components/header/header.lite.tsx
```diff
@@ -27,7 +33,7 @@ export default function DBHeader(props: DBHeaderProps) {
 		forcedToMobile: false,
 		toggle: () => {
 			if (props.onToggle) {
-				props.onToggle(!props.drawerOpen);
+				props.onToggle(!Boolean(props.drawerOpen));
```
mfranzke, line 36: are you not using `getBoolean` on purpose here ?

packages/components/src/components/input/input.lite.tsx
```diff
@@ -193,22 +196,25 @@ export default function DBInput(props: DBInputProps) {
...
-				max={props.max}
-				min={props.min}
-				readOnly={props.readOnly}
+				maxLength={getNumber(props.maxLength, props.maxlength)}
+				minLength={getNumber(props.minLength, props.minlength)}
+				max={getNumber(props.max)}
+				min={getNumber(props.min)}
+				readOnly={
+					getBoolean(props.readOnly, 'readOnly') ||
+					getBoolean(props.readonly, 'readonly')
```
mfranzke, line 209: based on this, we maybe would want to compare against both lowercase attribute value and propertyName within `getBoolean` method.

packages/components/src/components/navigation-item/navigation-item.lite.tsx
```diff
@@ -86,7 +93,9 @@ export default function DBNavigationItem(props: DBNavigationItemProps) {
 
 	onUpdate(() => {
 		if (props.subNavigationExpanded !== undefined) {
-			state.isSubNavigationExpanded = !!props.subNavigationExpanded;
+			state.isSubNavigationExpanded = Boolean(
```
mfranzke, line 96: are you not using `getBoolean` here on purpose ?

packages/components/src/components/checkbox/checkbox.lite.tsx
nmerget: `getBoolean` returns a valid value for HTML which can be boolean or undefined.

The JS `inderterminate` doesn't allow undefined...

packages/components/src/components/input/input.lite.tsx
nmerget: image

You would write it like this in JSX, so it's correct here

packages/components/src/components/navigation-item/navigation-item.lite.tsx
nmerget: fixed

packages/components/src/components/input/input.lite.tsx
mfranzke: Damn, I didn't recognize the letters upper- and lowercase party. Again.

mfranzke (approved):

PR #4011, opened 2025-04-03
feat: add min-inline-size to input, select and textarea

michaelmkraus:
## Proposed changes

This pull request adjusts the min-inline-size property for input, select, and textarea elements to improve layout consistency. The changes include:

* Baseline Adjustment: Sets a base min-inline-size for elements without icons.

* Icon Support: Increases the min-inline-size for elements that include icons or specific input types that imply icon usage.

* Enhanced Icon Handling: Applies an even larger size when both leading and trailing icons are present.

## Types of changes


mfranzke:
@michaelmkraus I'm partly struggling whether this might be a release 2.0.0 topic, as it's "breaking" expectations of min-width (or do you expect those to only get sized to the absolute bare minimum so it's anyhow always only a benefit without any (negative) "side effects"?

packages/components/src/components/input/input.scss
```diff
@@ -15,6 +15,28 @@ $icon-padding: calc(
 .db-input {
 	@include form-components.set-default-form-component(input);
 
+	// base inline size for inputs without icons
+	min-inline-size: variables.$db-sizing-lg;
+
+	// increased inline size for elements with icons or specific input types that have icons
+	&[data-icon]:not([data-hide-icon="true"]),
+	&[data-icon-after]:not([data-hide-icon="true"]),
```
mfranzke, line 23:
```suggestion
	&:is([data-icon], [data-icon-after]):not([data-hide-icon="true"]),
```

mfranzke (approved):

packages/components/src/components/input/input.scss
```diff
@@ -15,6 +15,28 @@ $icon-padding: calc(
 .db-input {
 	@include form-components.set-default-form-component(input);
 
+	// base inline size for inputs without icons
+	min-inline-size: variables.$db-sizing-lg;
```
nmerget, line 19: We can move this directly into `form-components.set-default-form-component` this applies to all 3 form-components right?

packages/components/src/components/textarea/textarea.scss
```diff
@@ -39,12 +39,17 @@
...
+	min-inline-size: variables.$db-sizing-lg;
+
 	textarea {
 		@extend %resize-textarea;
 		@extend %scrollbar;
 
 		overflow-y: scroll;
 
+		// assign min-inline-size also to textarea itself to ensure min-inline-size even when resizing the textarea
+		min-inline-size: variables.$db-sizing-lg;
```
nmerget, line 51: Maybe we can move this into `form-components.set-default-form-component` as well. It won't hurt the select and input and may be useful for the custom-select to share this size.

You should use the css variable I suggested [here](https://github.com/db-ux-design-system/core-web/pull/4011/files#r2028144657) to keep it the same every time :)

packages/components/src/styles/internal/_form-components.scss
```diff
@@ -305,6 +313,10 @@ $input-valid-types:
 		appearance: none;
 		max-inline-size: 100%;
 		inline-size: 100%;
+		min-inline-size: var(
+			--db-form-component-min-inline-size,
+			#{variables.$db-sizing-lg}
+		);
```
nmerget, line 319:
```suggestion
		min-inline-size: $db-min-inline-size;
```

packages/components/src/components/textarea/textarea.scss
```diff
@@ -45,6 +45,9 @@
...
```
nmerget, line 49:
```suggestion
```

packages/components/src/components/input/input.scss
```diff
@@ -15,6 +15,24 @@ $icon-padding: calc(
...
+	&:is([data-icon], [data-icon-after]):not([data-hide-icon="true"]),
+	&:has(
+			input[type="search"],
+			input[type="date"],
+			input[type="month"],
+			input[type="week"],
+			input[type="datetime-local"],
+			input[type="time"]
+		) {
+		min-inline-size: variables.$db-sizing-xl;
```
nmerget, line 28:
```suggestion
		--db-form-component-min-inline-size: #{variables.$db-sizing-xl};
```

packages/components/src/components/input/input.scss
```diff
@@ -15,6 +15,24 @@ $icon-padding: calc(
...
+	}
+
+	// even larger size when both - leading and trailing icons - are present
+	&[data-icon][data-icon-after] {
+		min-inline-size: form-components.$db-min-inline-size;
```
nmerget, line 33:
```suggestion
		--db-form-component-min-inline-size: #{variables.$db-sizing-2xl};
```

We need to use 2xl here right?

packages/components/src/components/select/select.scss
```diff
@@ -23,16 +23,18 @@ $has-before-padding: calc(
...
-	min-inline-size: calc(
-		4 * #{variables.$db-sizing-md} + var(--db-form-has-before) *
-			#{variables.$db-sizing-md}
-	);
-
 	@include form-components.set-default-form-component(select);
 
+	// increased inline size for elements with icons
+	&[data-icon]:not([data-hide-icon="true"]) {
+		min-inline-size: form-components.$db-min-inline-size;
```
nmerget, line 35:
```suggestion
		--db-form-component-min-inline-size: #{variables.$db-sizing-2xl};
```

packages/components/src/components/textarea/textarea.scss
mfranzke: @nmerget what about also deleting the comment on top of it and the empty line as well ?

nmerget, line 49:
```suggestion
```

packages/components/src/components/textarea/textarea.scss
```diff
@@ -45,6 +45,7 @@
 
 		overflow-y: scroll;
 
+
```
mfranzke, line 48:
```suggestion
```

packages/components/src/components/select/select.scss
```diff
@@ -23,16 +23,18 @@ $has-before-padding: calc(
 	--db-form-component-padding-inline-end: calc(
 		#{form-components.$font-size-height} + #{variables.$db-spacing-fixed-sm}
 	);
+	--db-form-component-min-inline-size: variables.$db-sizing-xl;
```
mfranzke, line 26:
```suggestion
	--db-form-component-min-inline-size: ${variables.$db-sizing-xl};
```

mfranzke (approved):

mfranzke (approved):

PR #4012, opened 2025-04-03
feat: adapt placeholder for input, textarea and select

michaelmkraus:
## Proposed changes

This pull request adapts the styling for placeholder text for input, textarea and select. It adapts the following styles:

* Uses the sans-serif font family as defined by var(--db-font-family-sans)
* Applies an italic font style
* Sets the opacity to 0.75
* Fixed incorrectly inserting the label as inline-text-content in case that someone uses the `options` property instead of composition

## Types of changes


packages/components/src/styles/internal/_form-components.scss
```diff
@@ -509,3 +509,9 @@ $input-valid-types:
 		}
 	}
 }
+
+@mixin placeholder-style {
```
nmerget, line 513: There is already a implementation for placeholders: 

https://github.com/db-ux-design-system/core-web/pull/4012/files#diff-fc8ed644a470e8653bcb58e284747ce63f7ab258a112d921ab5605159aa41a8cR320

Furthermore, we need to add the 0.75 opacity to the label as well if we have a the floating variant. There is a `component.$placeholder-disabled` token which we need to update, which will apply to the label and the placeholder.
Check the `@include placeholder-content()` to find every place inside the `_form-components.scss` and add the `font-style: italic` there. After this you should be able to remove the mixins from input, select and textarea.

packages/components/src/styles/internal/_component.scss
```diff
@@ -9,7 +9,7 @@ $min-mobile-header-height: calc(
 
 $default-disabled: 0.4;
 // 0.68 would be fine as well for a11y, but we are big Bryan Adams fans :)
```
nmerget, line 11:
```suggestion
```

This isn't correct anymore 😢

mfranzke: oh no … what a pity …

packages/components/src/styles/internal/_form-components.scss
```diff
@@ -127,6 +127,8 @@ $check-border-size: min(#{variables.$db-border-width-2xs}, 2px);
 			#{$selector} {
 				@include placeholder-content() {
 					opacity: component.$placeholder-disabled;
+					font-family: var(--db-font-family-sans);
+					font-style: italic;
```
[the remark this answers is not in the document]
michaelmkraus: When we use it, the block inside the @include is passed into the mixin and inserted into the @content. So, the opacity, font-family and font-style are applied to the ::placeholder and the sibling element with an id ending in -placeholder.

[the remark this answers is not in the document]
mfranzke: sorry, I misread that declaration.

Issue #4019, opened 2025-04-04
DBSelect dropdown checkmark on first element


flipswitchingmonkey:
### Which generators are impacted?

- [x] All
- [ ] HTML
- [ ] React
- [ ] Angular
- [ ] Vue
- [ ] Web components
- [ ] Power Apps

### Reproduction case

Opening a dropdown when no value is set, the first item in the dropdown list is always decorated with a checkmark. Even if no entry is selected, at next opening it is still marked as checked, even though no value is selected.

### Expected Behaviour

No checkmark when no value is selected.

### Screenshots

![Image](https://github.com/user-attachments/assets/bb4bed59-ceff-45da-bd0b-353cc640ba0e)

### Browser version

None

PR #4020, opened 2025-04-04
refactor: let's not hide the first element in the select anymore

mfranzke:
## Proposed changes

- it's inconsistent in between browsers (Safari would still display that `option[hidden]`)
- it's inconsistent and incorrect to regular `select`-HTML-elements, as those would actually mark the first `option` as checked as well, whereas this is actually the selected one (gets transmitted in form submits etc.)
- In case that somebody would like to use floating labels or `placeholder`, there should be an `option` that the users could set to "unselect" all selectable options, or in other words "reset" any other selection. We wouldn't provide this possibility at the moment anyhow.

Resolves https://github.com/db-ux-design-system/core-web/issues/4019

## Types of changes

### DIFF
--- a/packages/components/src/components/select/select.lite.tsx
+++ b/packages/components/src/components/select/select.lite.tsx
@@ -224,7 +224,9 @@ export default function DBSelect(props: DBSelectProps) {
 				}
 				aria-describedby={state._descByIds}>
 				{/* Empty option for floating label */}
-				<option hidden></option>
+				<Show when={props.variant === 'floating' || props.placeholder}>
+					<option class="placeholder"></option>
+				</Show>
 				<Show when={props.options}>
 					<For each={props.options}>
 						{(option: DBSelectOptionType) => (
--- a/packages/components/src/components/select/select.scss
+++ b/packages/components/src/components/select/select.scss
@@ -76,7 +76,7 @@ $has-before-padding: calc(
 		);
 	}
 
-	&:has(> select option:checked:not([hidden])) {
+	&:has(> select option:checked:not(.placeholder)) {
 		[id$="-placeholder"] {
 			display: none;
 		}
--- a/packages/components/src/styles/internal/_form-components.scss
+++ b/packages/components/src/styles/internal/_form-components.scss
@@ -115,7 +115,7 @@ $check-border-size: min(#{variables.$db-border-height-2xs}, 2px);
 		&:has(
 				#{$selector}:focus-within,
 				#{$selector}:is(input, textarea):not(:placeholder-shown),
-				> select option:checked:not([hidden])
+				> select option:checked:not(.placeholder)
 			) {
 			label {
 				@extend %db-overwrite-font-size-2xs;


packages/components/src/components/select/select.lite.tsx
```diff
@@ -229,7 +229,9 @@ export default function DBSelect(props: DBSelectProps) {
 				}
 				aria-describedby={props.ariaDescribedBy ?? state._descByIds}>
 				{/* Empty option for floating label */}
-				<option hidden></option>
+				<Show when={props.variant === 'floating' || props.placeholder}>
+					<option class="placeholder" value=""></option>
```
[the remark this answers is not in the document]
mfranzke: we're even already connecting the visual placeholder via `aria-describedby`-HTML-attribute to the `select`-HTML-element (which is even better to have it as a `description` than a `label`).

packages/components/src/components/select/select.lite.tsx
```diff
@@ -233,7 +233,9 @@ export default function DBSelect(props: DBSelectProps) {
...
```
michaelmkraus, line 241:
```suggestion
					<option class="placeholder" value="" disabled></option>
```
If it is not wanted to select the empty value, we can set 'disabled', so the value can't be selected and submitted.

mfranzke: Valid point. As the initial state would need to be this initial option selected, so that we could display the placeholder or floating label, it might the best and easiest for users to be able to get back to this state again even after doing another selection; this would be similar to input elements, and any other behaviour most likely an edge case. If somebody would still need this different behaviour, they could still request it.

michaelmkraus (dismissed):

packages/components/src/components/select/select.lite.tsx
```diff
@@ -94,7 +94,7 @@ export default function DBSelect(props: DBSelectProps) {
 				}
 			} else if (stringPropVisible(props.message, props.showMessage)) {
 				state._descByIds = state._messageId;
-			} else {
+			} else if (props.placeholder) {
 				state._descByIds = state._placeholderId;
 			}
```
mfranzke:
```suggestion
			} else if (props.placeholder) {
				state._descByIds = state._placeholderId;
			} else {
				state._descByIds = undefined;
			}
```

We would still need a default case.

packages/components/src/components/select/select.lite.tsx
```diff
@@ -182,7 +182,7 @@ export default function DBSelect(props: DBSelectProps) {
 
 			if (stringPropVisible(props.message, props.showMessage)) {
 				state._descByIds = messageId;
-			} else {
+			} else if (props.placeholder) {
 				state._descByIds = placeholderId;
 			}
```
mfranzke:
```suggestion
			} else if (props.placeholder) {
				state._descByIds = placeholderId;
			} else {
				state._descByIds = undefined;
			}
```

we would still need a default case.

PR #4077, opened 2025-04-17
feat: change color of label on hovered and active state

michaelmkraus:
## Proposed changes

* change color of label on hovered and active state

## Types of changes

### DIFF
--- a/packages/components/src/styles/internal/_form-components.scss
+++ b/packages/components/src/styles/internal/_form-components.scss
@@ -428,6 +428,21 @@ $input-valid-types:
 	}
 }
 
+@mixin get-validity-color-check-label-color($state, $key: "valid") {
+	$variant: successful;
+
+	@if ($key != "valid") {
+		$variant: critical;
+	}
+
+	&:has(input:not([role="switch"])) {
+		&:is(label),
+		> label {
+			color: var(--db-#{$variant}-on-bg-basic-emphasis-100-#{$state});
+		}
+	}
+}
+
 // CHECKBOX & RADIO
 
 @mixin get-validity-color-check($key: "valid") {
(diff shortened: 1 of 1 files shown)


mfranzke (dismissed):

__snapshots__/tag/component/chromium/DBTag-should-be-a-radio-checked.png
mfranzke, line 1: these most likely result out of that within DB Tag the DB Radio component is being used:
https://github.com/db-ux-design-system/core-web/blob/feat-optimize-hover-data-inputs/packages/components/src/components/tag/tag.spec.tsx#L86-L105

mfranzke: https://github.com/db-ux-design-system/core-web/issues/4095

nmerget, line 1: This snapshot looks wrong ^.^

mfranzke: indeed.

mfranzke: > This snapshot looks wrong ^.^

-> https://github.com/db-ux-design-system/core-web/issues/4095

nmerget:
There are some additonal things: 

- You need to check the [disabled](https://github.com/db-ux-design-system/core-web/pull/4077/files) state the label is interactive
- The radio[checked] shouldn't be interactive, the label shouldn't be interactive as well for this

packages/components/src/styles/internal/_form-components.scss
```diff
@@ -428,6 +428,21 @@ $input-valid-types:
 	}
 }
 
+@mixin get-validity-color-check-label-color($state, $key: "valid") {
```
nmerget, line 431: Can we merge this with `get-validity-color-check`? There is a similar [selector](https://github.com/db-ux-design-system/core-web/pull/4077/files#diff-fc8ed644a470e8653bcb58e284747ce63f7ab258a112d921ab5605159aa41a8cR485-R490). Maybe we should change the complete `@mixin set-default-check-element` to use the `$selector` instead of `&:has(input:not([role="switch"]))`.

There is another `hover` & `pressed` function for the [input](https://github.com/db-ux-design-system/core-web/pull/4077/files#diff-fc8ed644a470e8653bcb58e284747ce63f7ab258a112d921ab5605159aa41a8cR581-R587) and for the disabled state. By rewriting the complete `set-default-check-element` we could save a lot of complexity. Just use hover pressed if not `$selector=switch` or `radio:checked` or `input:disabled`

mfranzke:
blocked by https://github.com/db-ux-design-system/core-web/pull/4096, which needs to get further developed first.

.prettierrc.json
mfranzke, line 1: what does this change have to do with the topic of this ticket ?

nmerget: https://github.com/db-ux-design-system/core-web/pull/4155

nmerget (approved):

packages/components/src/styles/internal/_form-components.scss
```diff
@@ -428,73 +428,55 @@ $input-valid-types:
...
+		);
+
 		&:is(label),
 		> label {
-			color: var(--db-#{$variant}-on-bg-basic-emphasis-80-default);
+			cursor: helpers.$cursor-pointer;
+		}
+
+		input {
+			cursor: helpers.$cursor-pointer;
```
mfranzke, line 466: @michaelmkraus how is this related to changing the color ?

Issue #4075, opened 2025-04-17
All labels are displayed some pixels off


mfranzke:
### Which generators are impacted?

- [x] All
- [ ] HTML
- [ ] React
- [ ] Angular
- [ ] Vue
- [ ] Web components
- [ ] Power Apps

### Reproduction case

All labels are displayed a little off, moved some pixels to the top. It's most likely related to the new font-family.

### Expected Behaviour

All labels would still get displayed vertically centered.

### Screenshots

![Image](https://github.com/user-attachments/assets/d694fe96-251b-4621-b986-74fe2c13c3a6)

![Image](https://github.com/user-attachments/assets/6fc2dd3d-7d85-4105-bb1c-fb9616a08131)

### Browser version

None

### Add any other context about the problem here.

~probably related to `-webkit-font-smoothing: antialiased;`~

PR #4078, opened 2025-04-17
fix(DB Neo): removed local font references

mfranzke:
## Proposed changes

The new DB Neo Screen font seems to get rendered differently when referenced from locally installed files vs. the web fonts.

Resolves https://github.com/db-ux-design-system/core-web/issues/4075

## Types of changes

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,7 @@
     "@axe-core/playwright": "^4.10.1",
     "@commitlint/cli": "19.8.0",
     "@commitlint/config-conventional": "19.8.0",
-    "@db-ux/db-theme": "2.0.1",
+    "@db-ux/db-theme": "2.0.2",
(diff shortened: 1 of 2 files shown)


mfranzke:
looks much better now:
https://design-system.deutschebahn.com/core-web/review/4075-all-labels-are-displayed-some-pixels-off/components/data-input/custom-select/overview

package.json
```diff
@@ -110,6 +110,7 @@
     "zone.js": "~0.15.0"
   },
   "overrides": {
+    "@db-ux/db-theme-fonts": "2.0.1",
```
nmerget, line 113: Can you update `@db-ux/db-theme` instead of an override...

mfranzke: that was only test wise, but yes, sure.

PR #4096, opened 2025-04-22
refactor: adapted native elements syntax from documentation

mfranzke:
## Proposed changes

documentation: https://design-system.deutschebahn.com/core-web/version/v2.0.2/components/data-display/tag/overview#

Resolves https://github.com/db-ux-design-system/core-web/issues/4095

## Types of changes


nmerget:
Both variants should work the same way. Based on the changes of the snapshots, we need to fix the distance between the check icon. Furthermore, we should test both variants at least for the component tests

mfranzke:
> Both variants should work the same way. Based on the changes of the snapshots, we need to fix the distance between the check icon. Furthermore, we should test both variants at least for the component tests

#discuss

PR #4098, opened 2025-04-23
fix: sass mixed-decls deprecation

nmerget:
## Proposed changes

- Add `--fatal-deprecation=mixed-decls` to force an error for sass, this allows the pipeline to fail if we use a deprecated feature
- fixed issues from `mixed-decls`

## Types of changes


mfranzke:
> Add `--fatal-deprecation=mixed-decls` to prevent warnings in sass

to my understanding, `--fatal-deprecation=mixed-decls` isn't about preventing/silencing warnings (`--silence-deprecation=mixed-decls` would be doing this, but we would not want to do this at all), but increase the output level from warning to error. So I see a mismatch in between the description and solution of this PR.

nmerget:
> > Add `--fatal-deprecation=mixed-decls` to prevent warnings in sass
> 
> @nmerget to my understanding, `--fatal-deprecation=mixed-decls` isn't about preventing/silencing warnings (`--silence-deprecation=mixed-decls` would be doing this, but we would not want to do this at all), but increase the output level from warning to error. So I see a mismatch in between the description and solution of this PR.

I updated the description

packages/foundations/scss/icons/_icon-helpers.scss
```diff
@@ -122,10 +122,6 @@ $default-icon-font-size: var(--db-icon-font-size, #{$default-icon-size-rem});
 }
 
 %icon {
-	@include icon-content($default-icon-content); // TODO: ONLY FOR data-icon
-	// Hiding icon from screenreaders
-	-webkit-alt: "";
-	alt: "";
```
mfranzke, line 128: why have these declarations been removed as well besides moving the other `@include` declaration?

nmerget: Why do we need them? I couldn't find any information about those on mdn

mfranzke: those are included for backwards-compatibility. In case that we don't want to use them anymore, let's remove them in a separate PR so that we could make this decision transparent, as it's not clear under the circumstances of this PR.

mfranzke: https://github.com/db-ux-design-system/core-web/commit/56d7892b9bf13ee56e705784e25e72ab7e516b8c

PR #4136, opened 2025-04-30
fix: issue with select for angular not using initial value

### LINKED ISSUE #4076: db-select initial value not set (DB UX 2.0.0)

- [ ] All
- [ ] HTML
- [ ] React
- [x] Angular
- [ ] Vue
- [ ] Web components
- [ ] Power Apps

### Reproduction case

With the new release of DB UX 2.0.0 the db-select component no longer displays the initial value and remains empty after the site has finished loading.

Image

If the value is set after the view is initialised (for example with `AfterViewInit`) the initial value is displayed correctly.

### Expected Behaviour

The initial value of the db-select component should be displayd without explicitly setting the value after the view is initialized.

### Screenshots

_No response_

### Browser version

Chrome

_No response_

nmerget:
## Proposed changes

closes #4076

## Types of changes

### DIFF
--- a/packages/components/scripts/post-build/components.ts
+++ b/packages/components/scripts/post-build/components.ts
@@ -222,11 +222,19 @@ export const getComponents = (): Component[] => [
 			}
 		}
 	},
-
 	{
 		name: 'select',
 		overwrites: {
-			angular: [{ from: '<HTMLElement>', to: '<HTMLSelectElement>' }],
+			angular: [
+				{ from: '<HTMLElement>', to: '<HTMLSelectElement>' },
+				// We can move this to onMount with useTarget after https://github.com/BuilderIO/mitosis/pull/1750 is merged
+				{
+					from: 'ngAfterViewInit() {',
+					to:
+						'ngAfterViewInit() {\n' +
+						'\t  this.writeValue(this.value());'
+				}
+			],
 			react: [
 				// React not allowing selected for options
 				{ from: 'selected={option.selected}', to: '' },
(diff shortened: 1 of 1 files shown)


packages/components/scripts/post-build/components.ts
```diff
@@ -222,11 +222,19 @@ export const getComponents = (): Component[] => [
...
 		}
 	},
-
 	{
 		name: 'select',
 		overwrites: {
-			angular: [{ from: '<HTMLElement>', to: '<HTMLSelectElement>' }],
+			angular: [
+				{ from: '<HTMLElement>', to: '<HTMLSelectElement>' },
+				// TODO: We can move this to onMount with useTarget after https://github.com/BuilderIO/mitosis/pull/1750 is merged
```
mfranzke, line 230: We can move this to onMount with useTarget after https://github.com/BuilderIO/mitosis/pull/1750 is merged

packages/components/scripts/post-build/components.ts
```diff
@@ -222,11 +222,19 @@ export const getComponents = (): Component[] => [
...
+				{
+					from: 'ngAfterViewInit() {',
+					to:
+						'ngAfterViewInit() {\n' +
+						'\t  this.writeValue(this.value());'
+				}
```
mfranzke, line 236: We can move this to onMount with useTarget after https://github.com/BuilderIO/mitosis/pull/1750 is merged

mfranzke (approved):

Issue #4143, opened 2025-04-30
DBInput: Min and max props with type string results in NaN


biancavo:
### Which generators are impacted?

- [ ] All
- [ ] HTML
- [ ] React
- [ ] Angular
- [x] Vue
- [ ] Web components
- [ ] Power Apps

### Reproduction case

After updating from version 1.2.1 to 2.0.5 the min an max props with type string in DBInputs don't work anymore. DBInputs with min and max props with type number still work normally.

**Example**
Code: 
`<DBInput type="date" min="2024-12-15" max="2025-12-13" />`
Browser:
`<input type="date" min="NaN" max="NaN">`

### Expected Behaviour

The min an max props with type string should work normally as in the previous version.

**Example**
Code: 
`<DBInput type="date" min="2024-12-15" max="2025-12-13" />`
Browser:
`<input type="date" min="2024-12-15" max="2025-12-13">`

### Browser version

None

PR #4144, opened 2025-04-30
refactor(input): differentiating expected type for min and max

mfranzke:
## Proposed changes

types: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/min#syntax

Resolves https://github.com/db-ux-design-system/core-web/issues/4143

## Types of changes


packages/components/src/components/input/input.lite.tsx
```diff
@@ -221,8 +221,16 @@ export default function DBInput(props: DBInputProps) {
 				value={props.value ?? state._value}
 				maxLength={getNumber(props.maxLength, props.maxlength)}
 				minLength={getNumber(props.minLength, props.minlength)}
-				max={getNumber(props.max)}
-				min={getNumber(props.min)}
+				max={
+					props.type && ['number', 'range'].includes(props.type)
```
nmerget, line 225: You should move this to utils.ts to use it for `min` and `max`

nmerget (approved):

PR #4155, opened 2025-05-06
chore: remove prettier (`prettier-plugin-sh`) plugin

nmerget:
## Types of changes

### DIFF
--- a/.prettierrc.json
+++ b/.prettierrc.json
@@ -7,14 +7,8 @@
 			"options": {
 				"bracketSameLine": true
 			}
-		},
-		{
-			"files": [".env"],
-			"options": {
-				"parser": "sh"
-			}
 		}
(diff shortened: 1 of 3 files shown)


michaelmkraus (approved):

mfranzke:
@nmerget what's the reason to remove this plugin? The last publish has been one month old, and it supports the current version of `prettier`.

.prettierrc.json
mfranzke, line 1: @nmerget what's the reason to remove this plugin? The last publish has been one month old, and it supports the current version of prettier.

nmerget: I checked it again. The problem here is that the plugin uses a feature from prettier v3. Mitosis has currently a dependency which requires prettier < v3...

PR #4202, opened 2025-05-14
fix: focus outline for input calendar

### LINKED ISSUE #4134: DBInput: Focus in Calender-Icon is postponed

- [ ] All
- [ ] HTML
- [x] React
- [ ] Angular
- [ ] Vue
- [ ] Web components
- [ ] Power Apps

### Reproduction case

When DBInput is in use with a type like date, datetime, week, month and so on the focus is postponed. 

Image

This happens on Chrome and Edge. 

### Expected Behaviour

The focus should outline the icon right, with the icon being in the center of the focus-outline

### Screenshots

_No response_

### Browser version

Chrome

_No response_

### LINKED ISSUE #4228: DBInput: Type "datetime" and "datetime-local" show two icons on FireFox

- [ ] All
- [x] HTML
- [x] React
- [ ] Angular
- [ ] Vue
- [x] Web components
- [ ] Power Apps

### Reproduction case

In the current version, the DBInput component has a problem displaying the datepicker types "date" and "datetime-local" on FireFox. With these types two different "calendar" icons are displayed at the end of the field. The error only occurs on FireFox and can be seen on the Repo [DemoPage](https://design-system.deutschebahn.com/core-web/review/main/components/data-input/input/overview?page=example+-+types+-+floating+label)

Tested on:
MacOs 15.4.1
FireFox 130.0 (64-Bit)

### Expected Behaviour

The Component should only display one icon, as it does on Chrome and Safari.

### Screenshots

Image

### Browser version

other

_No response_

nmerget:
## Proposed changes

closes #4134
closes #4228

## Types of changes


packages/components/src/components/input/input.scss
```diff
@@ -105,25 +107,48 @@ $icon-padding: calc(
...
-		&::-webkit-search-cancel-button {
-			background-image: none;
-			inline-size: icons.$default-icon-font-size;
-			block-size: icons.$default-icon-font-size;
-			padding: 0;
-			cursor: pointer;
-		}
+		@supports selector(::-webkit-calendar-picker-indicator) {
+			&::-webkit-calendar-picker-indicator,
+			&::-webkit-search-cancel-button {
```
mfranzke, line 127: @nmerget could you please add a comment within the source code on why you've introduced the `@supports` wrapper? Regularly I would've expected that the browsers would based on the forgiving approach just ignore declarations that they don't support.

packages/components/src/components/input/input.scss
```diff
@@ -152,8 +177,10 @@ $icon-padding: calc(
 		@extend %dropdown-icon;
 
 		input {
-			&::-webkit-calendar-picker-indicator {
-				display: none !important;
+			@supports selector(::-webkit-calendar-picker-indicator) {
+				&::-webkit-calendar-picker-indicator {
```
mfranzke, line 181: @nmerget could you please add a comment within the source code on why you've introduced the `@supports` wrapper? Regularly I would've expected that the browsers would – based on the forgiving approach –just ignore declarations that they don't support.

__snapshots__/input/showcase/mobile-safari/should-have-same-aria-snapshot/DBInput-should-have-same-aria-snapshot.yaml
mfranzke, line 1: does this mean, that the calendar control is now unlabeled?

__snapshots__/input/showcase/firefox/DBInput-should-match-screenshot-1/DBInput-should-match-screenshot.png
mfranzke, line 1: several types input fields icons have disappeared.

nmerget: firefox doesn't support sime input types like "week", and `-webkit-calendar-picker-indicator` isn't supported as well. I removed the icons for firefox

packages/components/src/components/input/input.scss
nmerget: removed it

packages/components/src/components/input/input.scss
nmerget: removed it

__snapshots__/input/showcase/firefox/DBInput-should-match-screenshot-1/DBInput-should-match-screenshot.png
mfranzke: @nmerget but in these cases, it's not just the `input[type="week"]` and `input[type="month"]` icons that have been removed.

nmerget: yes all "time" type icons are removed, because we cannot hide the default calendar icon for firefox

mfranzke (dismissed):

mfranzke (approved):

PR #4213, opened 2025-05-15
fix: missing ariaDescribedBy property for form-components

### LINKED ISSUE #4212: Attribute `aria-describedby` is not considered for multiple form elements

- [x] All
- [ ] HTML
- [ ] React
- [ ] Angular
- [ ] Vue
- [ ] Web components
- [ ] Power Apps

### Reproduction case

Compared to `DBInput`, it is not possible to add an `aria-describedBy` attribute to other form elements.
Passing this information is ignored and but it shouldn't. 
Similiar to `DBInput` where the property is checked and used if passed, it should also be possible to pass a property for other form elements like `DBCheckbox` or `DBSelect`. It should also be documented.

**DBInput:**

```jsx
aria-describedby={props.ariaDescribedBy ?? state._descByIds}
```

See: https://github.com/db-ux-design-system/core-web/blob/main/packages/components/src/components/input/input.lite.tsx#L250

**DBSelect:**

```jsx
aria-describedby={state._descByIds}
```

See: https://github.com/db-ux-design-system/core-web/blob/main/packages/components/src/components/select/select.lite.tsx#L233
Or Checkbox - see: https://github.com/db-ux-design-system/core-web/blob/main/packages/components/src/components/checkbox/checkbox.lite.tsx#L207

### Expected Behaviour

All form elements should allow to pass a `aria-describedBy` prop (or `ariaDescribedBy`)

### Screenshots

_No response_

### Browser version

None

_No response_

nmerget:
## Proposed changes

closes #4212

## Types of changes


packages/components/src/components/checkbox/checkbox.lite.tsx
```diff
@@ -204,7 +204,7 @@ export default function DBCheckbox(props: DBCheckboxProps) {
 					onFocus={(event: InteractionEvent<HTMLInputElement>) =>
 						state.handleFocus(event)
 					}
-					aria-describedby={state._descByIds}
+					aria-describedby={props.ariaDescribedBy ?? state._descByIds}
```
mfranzke, line 207: @nmerget `aria-describedby` values could get chained, separated by spaces, to define more than one elements containing text describing an element. Do we want to replace our internal relationships or shouldn't these be additive?

mfranzke (approved):

packages/components/src/components/checkbox/checkbox.lite.tsx
nmerget: I think it should be a replacement

PR #4238, opened 2025-05-20
fix: additional custom select issues

### LINKED ISSUE #4236: DBCustomSelect property `tagWrapping` does not exist

- [ ] All
- [ ] HTML
- [x] React
- [ ] Angular
- [ ] Vue
- [ ] Web components
- [ ] Power Apps

### Reproduction case

The documentation doe the `DBCustomSelect` shows in the code examples the property `tagWrapping="grow"`. This seems not to exist (tested in React).

Image

Image

### Expected Behaviour

The property should probably exist or the documentation should be changed

### Screenshots

_No response_

### Browser version

Edge

_No response_

### LINKED ISSUE #4237: DBCustomSelect in DBAccordion shows multiple carets

- [x] All
- [ ] HTML
- [x] React
- [ ] Angular
- [ ] Vue
- [ ] Web components
- [ ] Power Apps

### Reproduction case

Using a `DBCustomSelect` within a `DBAccordion` will have side effect of the styles for both components.
In the result, the `DBCustomSelect` show two caret icons when the `DBAccordion` is open. @nmerget  we already had a look at it and I think you've already fixed it 😄 

Image

### Expected Behaviour

Only one caret is enough =)

### Screenshots

_No response_

### Browser version

None

_No response_

nmerget:
## Proposed changes

closes #4237
closes #4236

## Types of changes

### DIFF
--- a/packages/components/src/components/accordion-item/accordion-item.scss
+++ b/packages/components/src/components/accordion-item/accordion-item.scss
@@ -13,13 +13,13 @@
 
 	> details {
 		&[open] {
-			summary + div {
+			> summary + div {
 				block-size: auto;
 				/* stylelint-disable-next-line declaration-property-value-no-unknown */
 				block-size: calc-size(auto, size);
 			}
 
-			summary::after {
+			> summary::after {
 				transform: form-components.$dropdown-icon-transform;
 			}
 		}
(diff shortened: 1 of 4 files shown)


packages/components/src/components/accordion-item/accordion-item.scss
```diff
@@ -28,56 +28,52 @@
 			pointer-events: none;
 			opacity: component.$default-disabled;
 		}
-	}
 
-	@media screen and (prefers-reduced-motion: reduce) {
-		summary + div {
+		@media screen and (prefers-reduced-motion: no-preference) {
```
michaelmkraus, line 32: Changed from prefers-reduced-motion: reduce to prefers-reduced-motion: no-preference. Was this switch intentional?

nmerget: Yes I reverted the selector below to reduce the amount of css

michaelmkraus (approved):

PR #4253, opened 2025-05-22
feat: add new colors light-green and burgundy

michaelmkraus:
## Proposed changes

* add new color burgundy and light-green

## Types of changes

### DIFF
--- a/packages/foundations/scss/defaults/_default-variables.scss
+++ b/packages/foundations/scss/defaults/_default-variables.scss
@@ -2,7 +2,7 @@
  * Do not edit directly, this file was auto-generated.
  */
 
-:is(:root, :host) {
+:root {
 	--db-base-icon-weight-regular-desktop-headline-3xl: 96;
 	--db-base-icon-weight-regular-desktop-headline-2xl: 80;
 	--db-base-icon-weight-regular-desktop-headline-xl: 64;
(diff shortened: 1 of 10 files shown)


packages/foundations/scss/defaults/_default-variables.scss
```diff
@@ -2,7 +2,7 @@
  * Do not edit directly, this file was auto-generated.
  */
 
-:is(:root, :host) {
+:root {
```
nmerget, line 5: We need to update the output from theme-builder to add `:is(:root, :host)`.

You can revert this here otherwise it wouldn't work for shadow root components

mfranzke:
@michaelmkraus @nmerget these are most likely planned for release 2.2.0, not 2.1.0

nmerget:
If @leape is ready, we could release it in 2.1.0 as well

mfranzke:
> If @leape is ready, we could release it in 2.1.0 as well

exactly, that's what we need to find out later. But as it's part of the current sprint, some issues are currently still in progress, https://github.com/db-ux-design-system/core-team/issues/254 is defined for milestone 2.2.0 and we haven't talked about this within our final release planning meeting yesterday, I would doubt that we could make it for 2.1.0. But I will align on this.

PR #4254, opened 2025-05-22
feat: add field-sizing for input and textarea

### LINKED ISSUE #2497: DBTextarea multiline
Would it be possible to add multiline support to the existing DBTextarea component? We have a use case where the textarea element should start with 1 row, but be automatically adjusted each time user presses Enter or when the content is long and it breaks into a new row.

The end result would be something very similar/the same as here: https://mui.com/base-ui/react-textarea-autosize/.

nmerget:
## Proposed changes

closes #2497

## Types of changes

### DIFF
--- a/packages/components/src/styles/internal/_scrollbar.scss
+++ b/packages/components/src/styles/internal/_scrollbar.scss
@@ -7,6 +7,12 @@
 $scrollbar-width: helpers.px-to-rem(8);
 
 %scrollbar {
+	&[data-field-sizing="content"] {
+		@supports (field-sizing: content) {
+			scrollbar-width: none;
+		}
+	}
+
 	&::-webkit-scrollbar {
 		@extend %default-transition;
 
--- a/packages/components/src/components/input/input.lite.tsx
+++ b/packages/components/src/components/input/input.lite.tsx
@@ -211,6 +211,7 @@ export default function DBInput(props: DBInputProps) {
 			<input
 				aria-invalid={props.validation === 'invalid'}
 				data-custom-validity={props.validation}
+				data-field-sizing={props.fieldSizing}
 				ref={_ref}
 				id={state._id}
 				name={props.name}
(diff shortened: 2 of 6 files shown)


packages/components/src/components/textarea/model.ts
```diff
@@ -34,6 +34,12 @@ export type DBTextareaDefaultProps = {
 	 * In most browsers, textareas are resizable — you'll notice the drag handle in the right-hand corner, you can control it with this
 	 */
 	resize?: TextareaResizeType;
+
+	/**
+	 * Show/Hides drag handle in the right-hand corner - default: true
+	 */
+	showResizer?: boolean | string;
+
```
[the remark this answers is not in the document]
nmerget: We need it because it is the property from design. We keep `resize="none"` otherwise we have an breaking change

packages/components/src/styles/internal/_scrollbar.scss
```diff
@@ -7,6 +7,14 @@
 $scrollbar-width: helpers.px-to-rem(8);
 
 %scrollbar {
+	&[data-field-sizing="content"] {
+		&:is([data-hide-resizer="true"], [data-resize="none"]) {
+			@supports (field-sizing: content) {
+				scrollbar-width: none;
```
mfranzke, line 14: @nmerget what for do we need this? Could you please add a comment?

mfranzke (dismissed):

packages/components/src/components/textarea/textarea.scss
```diff
@@ -8,20 +8,26 @@
...
 %resize-textarea {
-	&[data-resize="none"] {
+	&[data-hide-resizer="true"] {
 		resize: none;
 	}
 
-	&[data-resize="both"] {
-		resize: both;
-	}
+	&:not([data-hide-resizer="true"]) {
```
mfranzke, line 16:
```suggestion
	// TODO: Consolidate data-hide-resizer and data-resize attributes in the future
	&:not([data-hide-resizer="true"]) {
```

packages/components/src/styles/internal/_scrollbar.scss
nmerget: done

mfranzke (approved):

PR #4258, opened 2025-05-23
feat: set border- and label-color for in/valid state of textarea, custom select and input

michaelmkraus:
## Proposed changes

* set border of textarea to corresponding color when in/valid state is set
* set floating label color of textarea, input and custom select to corresponding color when in/valid state is set

## Types of changes

### DIFF
--- a/packages/components/src/styles/internal/_form-components.scss
+++ b/packages/components/src/styles/internal/_form-components.scss
@@ -153,6 +153,24 @@ $db-min-inline-size: var(
 				}
 			}
 		}
+
+		@each $state, $variant in (valid: successful, invalid: critical) {
+			&[data-custom-validity="#{$state}"] > label,
+			&:has(> #{$selector}[data-custom-validity="#{$state}"]) > label {
+				color: var(--db-#{$variant}-on-bg-basic-emphasis-70-default);
+			}
+
+			@if ($selector == textarea) {
+				&:has(> #{$selector}[data-custom-validity="#{$state}"]) {
+					&::before,
+					&::after {
+						// necessary to force this override over the placeholder-content mixin's ::before rule
+						// to avoid the label to influence the border of the textarea
+						content: none !important;
+					}
+				}
+			}
+		}
 	}
 }
 
@@ -368,6 +386,16 @@ $input-valid-types:
 				--db-textarea-read-only,
 				#{colors.$db-adaptive-bg-basic-level-1-default}
 			) !important;
+		}
+
+		&:is(textarea) {
+			@each $state, $variant in (valid: successful, invalid: critical) {
+				&[data-custom-validity="#{$state}"] {
+					border: var(--db-border-width-3xs)
+						solid
+						var(--db-#{$variant}-on-bg-basic-emphasis-70-default);
+				}
+			}
 		}
 	}
 


packages/components/src/styles/internal/_form-components.scss
```diff
@@ -369,6 +387,16 @@ $input-valid-types:
...
+
+		&:is(textarea) {
+			@each $state, $variant in (valid: successful, invalid: critical) {
+				&[data-custom-validity="#{$state}"] {
+					border: var(--db-border-width-3xs)
+						solid
+						var(--db-#{$variant}-on-bg-basic-emphasis-70-default);
+				}
+			}
+		}
```
nmerget, line 429:
```suggestion
```

We don't need this here replace https://github.com/db-ux-design-system/core-web/pull/4258/files#diff-fc8ed644a470e8653bcb58e284747ce63f7ab258a112d921ab5605159aa41a8cR204-R228 with:

`	& {
		@if ($selector == textarea) {
			--db-textarea-read-only: var(
				--db-#{$variant}-bg-basic-level-1-default
			);
		}

		@extend %db-#{$variant}-variables;
	}`
	
I don't know why there was a workaround for textarea but with this we set all adaptive colors like we should.

packages/components/src/styles/internal/_form-components.scss
```diff
@@ -153,6 +153,24 @@ $db-min-inline-size: var(
...
+				&:has(> #{$selector}[data-custom-validity="#{$state}"]) {
+					&::before,
+					&::after {
+						// necessary to force this override over the placeholder-content mixin's ::before rule
+						// to avoid the label to influence the border of the textarea
+						content: none !important;
+					}
+				}
+			}
+		}
```
nmerget:
```suggestion
```

You cannot remove the content on `::before` or `::after` it is a workaround to ensure the floating label stays above the text. You see the problem, when you add a lot of text into the textarea.

michaelmkraus: I see the problem with long texts. We could handle it like that

```suggestion
		@each $state, $variant in (valid: successful, invalid: critical) {
			&[data-custom-validity="#{$state}"] > label,
			&:has(> #{$selector}[data-custom-validity="#{$state}"]) > label {
				color: var(--db-#{$variant}-on-bg-basic-emphasis-70-default);
			}
			@if ($selector == textarea) {
				&:has(> #{$selector}[data-custom-validity="#{$state}"]) {
					&::before,
					&::after {
						// necessary to force this override over the placeholder-content mixin's ::before rule
						// to avoid the label to influence the border of the textarea
						border-color: transparent;
						background-color: inherit !important;
					}
				}
			}
		}
```

This way, we would keep the workaround and ensure that the label does not influence the border of the textarea.

packages/components/src/styles/internal/_form-components.scss
```diff
@@ -220,6 +224,12 @@ $db-min-inline-size: var(
 		caret-color: var(--db-#{$variant}-on-bg-basic-emphasis-100-default);
 	}
 
+	@if $selector == textarea {
+		#{$selector} {
+			border-color: var(--db-#{$variant}-on-bg-basic-emphasis-90-default);
+		}
+	}
```
mfranzke: @michaelmkraus I couldn't have found the declarations for input quickly, but as textarea and input should be styled equally, I'm wondering why we don't extend the existing input declarations regarding this aspect and instead add this textarea specific condition here.

__snapshots__/textarea/showcase/firefox/DBTextarea-should-match-screenshot-1/DBTextarea-should-match-screenshot.png
mfranzke, line 1: @michaelmkraus the textarea borders are now red and green for the different validation states. But those colors are different comparing input and textarea:
- https://design-system.deutschebahn.com/core-web/review/fix-valid-invalid-data-inputs-border-label-color/components/data-input/textarea/overview?page=validation
- https://design-system.deutschebahn.com/core-web/review/fix-valid-invalid-data-inputs-border-label-color/components/data-input/input/overview?page=validation

If I'm not mistaken those should actually be equal, shouldn't they ?

packages/components/src/styles/internal/_form-components.scss
michaelmkraus: I replaced the textarea-specific block with the already existing variable that holds both, input and textarea.

__snapshots__/textarea/showcase/firefox/DBTextarea-should-match-screenshot-1/DBTextarea-should-match-screenshot.png
michaelmkraus: You are right. I checked this and adapted the color.

packages/components/src/styles/internal/_form-components.scss
```diff
@@ -182,6 +182,10 @@ $db-min-inline-size: var(
 		$variant: "critical";
 	}
 
+	@if $selector == textarea {
+		--db-textarea-read-only: var(--db-#{$variant}-bg-basic-level-1-default);
+	}
+
```
mfranzke, line 213: @michaelmkraus isn't this redundant to line 194 ?

michaelmkraus: You are right!

.changeset/sour-brooms-tie.md
mfranzke, line 1: this shouldn't have been deleted.

.changeset/heavy-days-call.md
mfranzke, line 1: this shouldn't have been deleted.

.changeset/chatty-jobs-tell.md
mfranzke, line 1: this shouldn't have been added, it was even already merged and published.

.config/.jscpd.json
mfranzke, line 1: these shouldn't have been removed.

.github/actions/cancel-workflow/action.yml
mfranzke, line 1: this shouldn't have been added.

__snapshots__/textarea/showcase/chromium/DBTextarea-should-match-screenshot-1/DBTextarea-should-match-screenshot.png
nmerget, line 1: @michaelmkraus right now, the label is always colored. But it should only be colored for the floating label

michaelmkraus: Now only the floating label is colored.

packages/components/src/styles/internal/_form-components.scss
```diff
@@ -1,6 +1,7 @@
 @use "sass:map";
 @use "sass:string";
 @use "sass:list";
+@use "sass:selector";
```
mfranzke: @michaelmkraus are you actually using this `selector.` tool? I can't find it in this file.

michaelmkraus: Yes, it is no more used. I removed it.

Issue #4246, opened 2025-05-20
DBTabItem: generated `aria-controls` value does not refer to an existing element with the `id`


d-koppenhagen:
### Which generators are impacted?

- [x] All
- [ ] HTML
- [x] React
- [ ] Angular
- [ ] Vue
- [ ] Web components
- [ ] Power Apps

### Reproduction case

The generated value for `aria-controls` does not point to an existing element with its corresponding `id`.

Example code:

```jsx
<DBTabs>
    <DBTabList>
        <DBTabItem>Einfache Fahrt</DBTabItem>
        <DBTabItem>Hin- und Rückfahrt</DBTabItem>
    </DBTabList>
</DBTabs>
```

example generated / rendered:

```
aria-controls="tabs-bb11d662-1521-4d2c-8f6c-bb9ebfaabc0d-tab-panel-0"
```

In this case there is no element with the id `tabs-bb11d662-1521-4d2c-8f6c-bb9ebfaabc0d-tab-panel-0`.

### Expected Behaviour

`aria-controls` should refer to the matching element or probably be removed when not explicitly defined.

### Screenshots

<img width="1524" alt="Image" src="https://github.com/user-attachments/assets/acd65d55-9be8-4b2b-bb5e-7d5a39033ea2" />

### Browser version

Edge

PR #4270, opened 2025-05-26
fix(tabs): only set aria-controls attribute when tab panels exist

michaelmkraus:
## Proposed changes

* Add `aria-controls` attribute only when panels exist

closes https://github.com/db-ux-design-system/core-web/issues/4246

## Types of changes


packages/components/src/components/tabs/tabs.lite.tsx
```diff
@@ -97,12 +102,14 @@ export default function DBTabs(props: DBTabsProps) {
 						if (!input.id) {
 							const tabId = `${state._name}-tab-${index}`;
 							label.setAttribute('for', tabId);
-							input.setAttribute(
-								'aria-controls',
-								`${state._name}-tab-panel-${index}`
-							);
 							input.id = tabId;
 							input.setAttribute('name', state._name);
+							if (tabPanels.length > 0) {
```
nmerget, line 107: You should check if the same index exists as well.

If a user has 2 items and only 1 panel it would be an issue as well

PR #4283, opened 2025-05-27
fix(infotext): preserve spaces around inline tag values

### LINKED ISSUE #2714: DBInfotext is missing spaces around tags like bold, code etc

- [X] All
- [ ] HTML
- [ ] React
- [ ] Angular
- [ ] Vue
- [ ] Web components
- [ ] Power Apps

### Reproduction case

Add a text with tags such as `b`, `strong`, `u` or `code` to **c**. As **DBInfotext** uses `display: flex`, spaces around these tags are not displayed.

### Expected Behaviour

Spaces around tags are also displayed within DBInfotext.

### Screenshots

Bildschirmfoto 2024-06-04 um 11 54 38

### Browser version

None

_No response_

michaelmkraus:
## Proposed changes

This PR fixes a visual bug in `DBInfotext` where spaces surrounding inline elements (like `<b>`, `<strong>`, `<code>`, `<a>`) were not displayed correctly.

The `DBInfotext` component uses `display: flex` for alignment. In a flex container, direct text nodes are treated as anonymous flex items. This causes the browser to handle whitespace collapsing differently, resulting in spaces around tags.

We use a `<span>` to maintain valid HTML, as `DBInfotext` itself renders as a `span`. The wrapper is styled with `display: block` to create a new block formatting context. This isolates the text content from the parent's flex layout, restoring standard browser whitespace behavior.

closes https://github.com/db-ux-design-system/core-web/issues/2714

## Types of changes

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/fix-preserve-whitespaces-infotext>


nmerget (dismissed):

__snapshots__/textarea/showcase/chromium-highContrast/should-have-same-aria-snapshot/DBTextarea-should-have-same-aria-snapshot.yaml
mfranzke, line 1: why are those snapshot removals part of a styled related change ?

__snapshots__/badge/component/chromium/DBBadge-should-match-screenshot.png
mfranzke, line 1: this tag is changing it's color – I wouldn't expect a side effect like this from a white-space CSS change.

__snapshots__/badge/showcase/mobile-safari/DBBadge-should-match-screenshot-1/DBBadge-should-match-screenshot.png
mfranzke, line 1: there are several huge unexpected changes included in this screenshot.

packages/components/src/components/infotext/infotext.lite.tsx
```diff
@@ -28,9 +27,11 @@ export default function DBInfotext(props: DBInfotextProps) {
...
 			data-size={props.size}
 			data-show-icon-leading={getBooleanAsString(props.showIcon ?? true)}>
-			<Show when={props.text} else={props.children}>
-				{props.text}
-			</Show>
+			<span class="db-infotext-content">
+				<Show when={props.text} else={props.children}>
+					{props.text}
+				</Show>
+			</span>
```
nmerget: This doesn't fix the problem or am I doing something wrong?

Image

michaelmkraus: image

It looks fine for me.

nmerget: I'm not a big fan of wrapping a `<span>` inside a `<span>`, what do you think @mfranzke .

Maybe a solution could be using `gap: calc(1ch / 2)`? This should be exactly the spacing of ` `. We just need to remove this gap from the `--db-icon-margin-end` to align the icon correctly. Then we could keep the `display: flex` and would remove the second wrapper.

michaelmkraus: The wrapper-span is necessary to isolate the flex-layout from the text-layout. Inside the wrapper, the standard inline formatting applies, which preserves the whitespace around inline elements. Since the component might be used inside a `<p>`, using a `<span>` (instead of a `<div>`) is - as far as I know - semantically the safest choice to avoid HTML validation errors.

nmerget: The infotext should be a helper component and shouldn't be used inside a `<p>`, which suggests that it is a block text. It should stand alone by itself

Maybe the `<span>` was wrong in the first place as wrapping element. But `<span>` inside `<span>` looks wrong to me.

Right now we use `flex` to align the icon with the text. Maybe the solution is to use `<p>` as wrapper with `<span>` inside?

michaelmkraus: Since the root element currently is a `<span>`, changing it to a `<p>` or `<div>` right now would be a breaking change. If a consumer currently uses `<DBInfotext>` inside a paragraph, the browser would [automatically close](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/p#technical_summary) their outer `<p>` (paragraph inside paragraph is not valid), breaking their layout.

mfranzke: > Maybe a solution could be using `gap: calc(1ch / 2)`? This should be exactly the spacing of ` `. We just need to remove this gap from the `--db-icon-margin-end` to align the icon correctly. Then we could keep the `display: flex` and would remove the second wrapper.

I like this solution.

packages/components/src/components/infotext/infotext.scss
```diff
@@ -6,11 +6,15 @@
 @use "@db-ux/core-foundations/build/styles/icons";
 
 .db-infotext {
-	--db-icon-margin-end: #{variables.$db-spacing-fixed-2xs};
+	--db-icon-margin-end: calc(
+		#{variables.$db-spacing-fixed-2xs} - calc(1ch / 2)
+	);
```
nmerget:
```suggestion
	--db-icon-margin-end: 0;
```

This has to be `0` right? Otherwise we have the gap twice

michaelmkraus: Right. Fixed it.

nmerget (approved):

mfranzke:
@codex please review

packages/components/src/components/infotext/infotext.scss
```diff
@@ -6,11 +6,13 @@
 @use "@db-ux/core-foundations/build/styles/icons";
 
 .db-infotext {
-	--db-icon-margin-end: #{variables.$db-spacing-fixed-2xs};
+	--db-icon-margin-end: 0;
 
 	@extend %db-overwrite-font-size-sm;
 	@include helpers.display(flex);
 
+	gap: calc(1ch / 2);
```
mfranzke, line 14:
```suggestion
	gap: .5ch;
```

@michaelmkraus wouldn't this be the same as `.5ch`?

.changeset/tender-impalas-work.md
```diff
@@ -0,0 +1,5 @@
+---
+"@db-ux/core-components": patch
```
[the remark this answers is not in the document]
mfranzke: added with https://github.com/db-ux-design-system/core-web/commit/1bee561de6984dd8d27570e7f219c8f5f3365d96

PR #4291, opened 2025-05-28
Fix patternhub warnings

michaelmkraus:
## Proposed changes

* Boolean-as-content warnings
Changed content: true flags in tag variants to a string ("true"), that <db-tag> no longer receives raw booleans as children.

* Controlled vs. uncontrolled inputs
Replaced checked: true with defaultChecked: true to remove the “missing onChange handler” warning.

* Label for → htmlFor
Updated <label> prop from "for": "…” to "htmlFor": "…” so that React creates the correct htmlFor attribute instead of the invalid for.

* List key warnings
Added key props to <option> elements rendered inside <select> because of React’s requirement for unique keys.

## Types of changes

### DIFF
--- a/packages/components/src/components/select/select.lite.tsx
+++ b/packages/components/src/components/select/select.lite.tsx
@@ -236,6 +236,7 @@ export default function DBSelect(props: DBSelectProps) {
 									when={option.options}
 									else={
 										<option
+											key={option.value.toString()}
 											value={option.value}
 											disabled={option.disabled}
 											selected={option.selected}>
--- a/showcases/shared/tag.json
+++ b/showcases/shared/tag.json
@@ -189,7 +189,7 @@
 						"native": true,
 						"content": "Interactive Checkbox",
 						"props": {
-							"for": "checkbox01"
+							"htmlFor": "checkbox01"
 						},
 						"children": [
 							{
@@ -216,7 +216,7 @@
 						"native": true,
 						"content": "Interactive Radio 1",
 						"props": {
-							"for": "radio01"
+							"htmlFor": "radio01"
 						},
 						"children": [
 							{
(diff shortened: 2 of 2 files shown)


packages/components/src/components/select/select.lite.tsx
```diff
@@ -236,6 +236,7 @@ export default function DBSelect(props: DBSelectProps) {
 									when={option.options}
 									else={
 										<option
+											key={option.value.toString()}
```
nmerget, line 239: We should do this the same way like the [custom-select](https://github.com/db-ux-design-system/core-web/blob/main/packages/components/src/components/custom-select/custom-select.lite.tsx#L776-L782). Only Angular and React needs a key. And we should use a function to obtain it. Maybe we can use the same function as we used it in custom-select by moving it to `utils.ts`?

showcases/shared/tag.json
```diff
@@ -189,7 +189,7 @@
 						"native": true,
 						"content": "Interactive Checkbox",
 						"props": {
-							"for": "checkbox01"
+							"htmlFor": "checkbox01"
```
nmerget: Those will be rendered inside the `showCode` section. We should change the generation here instead of using `htmlFor` because it is only used in React

showcases/shared/tag.json
```diff
@@ -296,22 +296,22 @@
 				"name": "True",
 				"props": {
 					"component": "checkbox",
-					"checked": true
+					"defaultChecked": true
```
nmerget, line 299: We need to handle this inside the react component, otherwise it would break the showcases for the other frameworks

packages/components/src/components/select/select.lite.tsx
michaelmkraus: Moved it into utils and used it in select and custom-select

mfranzke (approved):

mfranzke (dismissed):

nmerget (approved):

PR #4292, opened 2025-05-28
feat: self healing dependabot updates

mfranzke:
## Proposed changes

On `prettier` and `stylelint` dependency updates the codestyles might have changed so that the pipelines would initially fail and someone would need to run those linter/code formatters to transform the codebase to the new codestyles. Instead of doing this manually it would be nice to have a "self-healing" pipeline.

The only remaining step at the moment would be to retrigger the pipeline again, as GitHub wouldn't allow a pipeline run to get triggered after the pipeline has commited changed code (like in this case).<details><summary>ChatGPTs answer on why the pipeline wouldn't run after a commit by the pipeline itself</summary>

You're running into a known and nuanced behavior of GitHub Actions.

### ✅ Summary of the Issue

When you **push a commit to a Pull Request branch from within a GitHub Actions workflow**, GitHub **cancels the current workflow** (as expected) and **starts a new one**, but sometimes **the new one does not run any jobs** and just says:

> "There are no checks for this commit"

This happens especially when:

* The workflow is modifying and committing to the same branch it is running from.
* The commit is made with a bot token (e.g., `GITHUB_TOKEN`), not a real user PAT (Personal Access Token).
* The workflow uses default `on: pull_request` or `on: push` triggers.

---

### 🔍 Why It Happens

1. **Workflow-trigger loop prevention**
   GitHub has **safety logic** to prevent infinite CI loops. If a workflow pushes a commit back to the same branch (or PR) it's running from, **GitHub may choose not to trigger a new workflow**, especially if the commit came from a GitHub Action itself.

2. **`GITHUB_TOKEN` behavior**

   * Commits made using `GITHUB_TOKEN` don't trigger workflows on `push` or `pull_request` events unless explicitly configured.
   * This token is designed to avoid self-triggering behavior.

3. **Triggering logic specifics**

   * If the new commit **doesn’t change any files that affect workflow conditions**, or if the workflow file was modified but is now identical, GitHub may skip re-running the pipeline.
   * Also, if there is **no `on: push` event configured**, and you're committing directly to a branch, nothing may run.

---

#### ✅ Option 1: Use a Personal Access Token (PAT)

Replace `GITHUB_TOKEN` with a PAT from a GitHub user with permission. This allows the commit to trigger workflows as if it were from a real developer.

```yaml
- name: Commit and push changes
  run: |
    git config user.name "your-username"
    git config user.email "your-email@example.com"
    git add .
    git commit -m "Automated update"
    git push https://<username>:${{ secrets.PAT }}@github.com/owner/repo.git HEAD:branch
```

> ⚠️ **Do not hardcode the token** — use `secrets.PAT`.

#### ✅ Option 2: Use `workflow_run` to chain jobs

If you want to chain workflows after a commit, use `workflow_run` to explicitly start a second workflow after the first completes.

#### ✅ Option 3: Modify `on:` conditions

Ensure your workflow is triggered by both `push` **and** `pull_request`:

```yaml
on:
  push:
    branches:
      - '**'
  pull_request:
    types: [opened, synchronize, reopened]
```

---

### 🔍 Optional: Debug with `actions/github-script`

Use `actions/github-script` to log what's happening with events and commits programmatically if needed.

---

### ✅ TL;DR

* **Yes,** GitHub Actions **can** make commits that trigger further workflows.
* **But** if you use `GITHUB_TOKEN`, GitHub suppresses those triggers **by default** to avoid loops.
* Use a **Personal Access Token** if you want those commits to start new workflows.
* Also ensure your workflows are listening for the right `on:` events.

Let me know if you want a code example or help setting up a PAT securely.
</details>

Additionally some problems wouldn't be auto-fixable (by `prettier --write` or `stylelint --fix`), those would still need to be done manually obviously.

I've successfully tested it with https://github.com/db-ui/base/pull/665/commits/e3efcf0c11d314f840310bb6e8494144bba82ec5

## Types of changes


.github/workflows/99-self-healing-dependabot-updates.yml
```diff
@@ -0,0 +1,60 @@
+name: Auto-Format with Stylelint and Prettier on PR for "self-healing" PRs
+
+on:
+  workflow_call:
+
+jobs:
+  format:
+    runs-on: ubuntu-latest
```
nmerget, line 8: Add `if: ${{ github.actor == 'dependabot[bot]' }}` to avoid running this in a normal PR

.github/workflows/99-self-healing-dependabot-updates.yml
```diff
@@ -0,0 +1,60 @@
...
+            echo "Stylelint update detected."
+            echo "stylelint_update=true" >> $GITHUB_ENV
+          elif [[ "${{ github.event.pull_request.title }}" =~ "bump prettier from" ]]; then
+            echo "Prettier update detected."
+            echo "prettier_update=true" >> $GITHUB_ENV
+          else
+            echo "No Stylelint or prettier updates detected."
+          fi
+
+      - name: Set up Node.js
```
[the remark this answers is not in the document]
mfranzke: noooooo

.github/workflows/99-self-healing-dependabot-updates.yml
```diff
@@ -0,0 +1,60 @@
...
+
+      - name: Install dependencies
+        if: env.stylelint_update == 'true' || env.prettier_update == 'true'
+        run: |
+          npm ci
+
+      - name: Run Stylelint to format the code
+        if: env.stylelint_update == 'true'
+        run: |
+          npx --no stylelint "**/*.{css,scss}" --fix
```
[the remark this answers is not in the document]
mfranzke: #discuss

.github/workflows/99-self-healing-dependabot-updates.yml
```diff
@@ -0,0 +1,60 @@
...
+            echo "stylelint_update=true" >> $GITHUB_ENV
+          elif [[ "${{ github.event.pull_request.title }}" =~ "bump prettier from" ]]; then
+            echo "Prettier update detected."
+            echo "prettier_update=true" >> $GITHUB_ENV
+          else
+            echo "No Stylelint or prettier updates detected."
+          fi
+
+      - name: Set up Node.js
+        if: env.stylelint_update == 'true' || env.prettier_update == 'true'
```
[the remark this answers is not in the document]
mfranzke: good question, I even also thought about using a general env variable. But I would need the tool specific one for differentiating the runs on https://github.com/db-ux-design-system/core-web/pull/4292/files#diff-32d2647f0f2aac6258db5b1714a1ca87ddd57204462755e6d9f72ca8c17743efR42-R50

.github/workflows/99-self-healing-dependabot-updates.yml
```diff
@@ -0,0 +1,91 @@
...
+
+      - name: Install dependencies
+        if: env.stylelint_update == 'true' || env.prettier_update == 'true'
+        run: |
+          npm ci
+
+      - name: Run Stylelint to format the code
+        if: env.stylelint_update == 'true'
+        run: |
+          npx --no stylelint "**/*.{css,scss}" --fix
```
mfranzke:
```suggestion
          npm run lint:stylelint --fix
```

.github/workflows/99-self-healing-dependabot-updates.yml
```diff
@@ -0,0 +1,91 @@
...
+        if: env.stylelint_update == 'true'
+        run: |
+          npm run lint:stylelint --fix
+
+      - name: Run Prettier to format the code
+        if: env.prettier_update == 'true'
+        run: |
+          npx --no prettier . --write
+
+      # https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/making-authenticated-api-requests-with-a-github-app-in-a-github-actions-workflow
```
nmerget: I think it makes sense to move the part below into a custom action to use it for every auto-commit in our repo.

It would be an additional advantage because you could use the `if: env.stylelint_update == 'true' || env.prettier_update == 'true'` once. Would you like to do it or should I help you?

mfranzke: @nmerget I like this idea, and the latter, please.

.github/workflows/00-init.yml
```diff
@@ -17,10 +17,10 @@ jobs:
           ASSET_INIT_VECTOR: ${{ secrets.ASSET_INIT_VECTOR }}
           ASSET_PASSWORD: ${{ secrets.ASSET_PASSWORD }}
 
-#      - name: 🚮 Dump GitHub context for debugging
-#        env:
-#          GITHUB_CONTEXT: ${{ toJson(github) }}
-#        run: echo "$GITHUB_CONTEXT"
+      - name: 🚮 Dump GitHub context for debugging
```
nmerget: @mfranzke I would keep this always inside the pipeline. We don't like to uncommet/comment it every time we debug.

mfranzke: It was a comment by GitHub Copilot that sharing the full content of this variable might leak internal information. As the logs are public, this might be a valid aspect. Is there a non-public-space we could pass the output to ?

nmerget (approved):

nmerget (approved):

package.json
```diff
@@ -30,11 +30,12 @@
     "clean": "git clean -dfx --exclude=.env",
     "commit:updated-snapshots": "git diff --name-only --diff-filter=M | xargs git add && git commit -m 'test: updated snapshots'",
     "dev": "npm run dev --workspace=scripts",
+    "fmt": "prettier . --write",
```
mfranzke, line 33:
```suggestion
    "format": "prettier . --write",
```

.github/workflows/99-self-healing-dependabot-updates.yml
```diff
@@ -0,0 +1,61 @@
...
+
+      - name: 🏃 Run Stylelint to format the code
+        if: env.stylelint_update == 'true'
+        run: |
+          npm run lint:stylelint --fix
+
+      - name: 🏃 Run Prettier to format the code
+        if: env.prettier_update == 'true'
+        run: |
+          npm run fmt
```
mfranzke, line 53:
```suggestion
          npm run format
```

package.json
nmerget: [fmt ](https://docs.deno.com/runtime/reference/cli/fmt/)is a common abbreviation

.github/workflows/99-self-healing-dependabot-updates.yml
nmerget: https://github.com/db-ux-design-system/core-web/pull/4292#discussion_r2166345231

package.json
mfranzke: In the tradeoff in between reading code and interacting with our software vs. clarity, there's no benefit in a short name.

mfranzke: https://github.com/db-ux-design-system/core-web/issues/4411

PR #4309, opened 2025-05-30
fix: issue with `package.json` overrides and `vite` dependency

nmerget:
## Proposed changes

the `overrides` part in `package.json` lead to a broken `package-lock.json`

## Types of changes

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,7 +69,7 @@
     "@guidepup/guidepup": "0.24.0",
     "@guidepup/playwright": "^0.14.2",
     "@guidepup/record": "^0.1.0",
-    "@playwright/test": "1.52.0",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -103,14 +103,14 @@
...
-    "vite": "^6.3.5",
+    "vite": "6.3.5",
     "vitest": "^3.1.4",
     "xo": "^0.60.0",
     "zone.js": "~0.15.1"
   },
   "overrides": {
     "cross-spawn": "~7.0.5",
-    "vite": "$vite"
+    "vite": "6.3.5"
```
mfranzke, line 113: @nmerget dependabot most likely wouldn't update this entry, would it? So we would have some manual process.

PR #4366, opened 2025-06-11
feat: add property `requiredAnnotation` to disable visual indicator (`*`) for semantically marked required form elements

### LINKED ISSUE #4197: Allow to not add "*" for form control labels when marked as required
Currently `DBSelect`, `DBInput` etc add a small asterisk in the `::after` Pseudo Element of their label when the control is marked as required. This should be optional or skippable.

Let's say we have a simple form with a few `<input>` and `<select>` fields and we have a form-wide description that all fields are required or we simply saying everything is required unless a field is labelled as "optional".

So in my case I assume that I don't need to visually tell users, for every field that it is required since the every information is required and probably I've also written it before. In such case I think the asterisk "*" coming from `input[required=true]` or `select[required=true]` is may visually distracting. Nonetheless, I want to semantically mark the fields as `required` sind when going through the form with a screenreader I don't have this visual overview an probably I jump directly into the form without reading the hint before that every field is required. In this cases, I want to clearly know that the input is required.

Long story short: It would be nice to have an option to disable / hide the Asterisk ("*") within the label even if a control has been marked as required.

For example:

```jsx
<DBInput label="firstname" required ={true} requiredAnnotation={false} />
```

In such case the rendered HTML should contain `required="true"` but the CSS applied to the label should not have the "*" in the `::after()` element.

d-koppenhagen:
## Proposed changes

add property `requiredAnnotation` to disable visual indicator (`*`) for semantically marked required form elements

## Types of changes

## Further comments

closes #4197 / https://github.com/db-ux-design-system/core-team/issues/580

### DIFF
--- a/packages/components/src/components/checkbox/checkbox.lite.tsx
+++ b/packages/components/src/components/checkbox/checkbox.lite.tsx
@@ -8,7 +8,7 @@ import {
 	useStore,
 	useTarget
 } from '@builder.io/mitosis';
-import { DBCheckboxProps, DBCheckboxState } from './model';
+
 import {
 	DEFAULT_INVALID_MESSAGE,
 	DEFAULT_INVALID_MESSAGE_ID_SUFFIX,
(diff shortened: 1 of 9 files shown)


packages/components/src/components/checkbox/checkbox.lite.tsx
mfranzke, line 1: @d-koppenhagen do you follow a specific pattern with the reordering until line 35 ?

packages/components/src/components/checkbox/checkbox.lite.tsx
```diff
@@ -192,6 +194,10 @@ export default function DBCheckbox(props: DBCheckboxProps) {
 					disabled={getBoolean(props.disabled, 'disabled')}
 					value={props.value}
 					required={getBoolean(props.required, 'required')}
+					{...(getBooleanAsString(props.requiredAnnotation) ===
+					'false'
+						? { 'data-required-icon': 'false' }
+						: {})}
```
nmerget:
```suggestion
data-hide-required-asterisk={getHideProp(props.showRequiredAsterisk)}
```

We have already something `data-hide-icon={getHideProp(props.showIcon)}` for many components. I would like to stick with this format for every prop which is `true` at the beginning. 
Furthermore, I think we should name the prop `showRequiredAsterisk` what do you think?

packages/components/src/components/checkbox/checkbox.lite.tsx
d-koppenhagen: Not really, I think in my IDE ordered imports are active. Should I revert this part?

PR #4407, opened 2025-06-24
feat(datalist): possibility to provide a hidden value

mfranzke:
## Types of changes


nmerget:
[`value`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/datalist) is required why do we need this PR?

mfranzke:
let's close it for the moment, as we most likely won't need it soon.

Issue #2515, opened 2024-04-15
card: evaluate simplifications regarding `interactive` behaviour


mfranzke:
why don't you just make it a `button`?

_Originally posted by @mfranzke in https://github.com/db-ui/mono/pull/1897#discussion_r1565210409_

nmerget:
It shouldn't be a button, look at this example: https://github.com/db-ui/mono/blob/main/showcases/patternhub/components/default-page.tsx#L176

nmerget:
related https://github.com/db-ux-design-system/core-web/issues/212

Issue #212, opened 2023-01-18
Card: provide `button` or `a` tag within the `card` component


mfranzke:
We should either use a `button` or `a` HTML tag within the `card` component to enable keyboard and screenreader compatibility.

_Originally posted by @mfranzke in https://github.com/db-ui/mono/pull/211#discussion_r1073673263_

nmerget:
We should use `<button>` for interactive cards and then there should be a `disabled` prop for UX and tech

PR #4506, opened 2025-07-09
feat: align how to use a DBCard as button or anchor

nmerget:
## Proposed changes

solves #2515
solves #212

## Types of changes


docs/migration/v2.x.x-to-v3.0.0.md
```diff
@@ -16,6 +16,30 @@ Please provide the correct `type` property in your code anyhow as a best practis
...
+## db-card/DBCard
+
+### `behavior='interactive'` change
+
+We changed the `behavior='interactive'` property not applying `role='button'` and `tabIndex` anymore.
+If you want to use an interactive card, you should wrap the card with the correct HTML element, like a `button` or `a` tag:
+
+```html
+<!--Angular-->
+<button>
```
mfranzke:
```suggestion
<button type="button" aria-label="button copy">
```

docs/migration/v2.x.x-to-v3.0.0.md
```diff
@@ -16,6 +16,30 @@ Please provide the correct `type` property in your code anyhow as a best practis
...
+	<db-card behavior="interactive">
+		<!-- card content -->
+	</db-card>
+</button>
+
+<!--React/Vue-->
+<button>
```
mfranzke:
```suggestion
<button type="button" aria-label="button copy">
```

We should both advocate for developers setting the type explicitly. And in most cases when adding a wrapping interactive element not the whole contents of a card should be the accessible name of that interactive element. That for we should also encourage the developers to add this attribute.

showcases/vue-showcase/src/components/card/Card.vue
```diff
@@ -9,7 +9,17 @@ import DefaultComponent from "../DefaultComponent.vue";
 		<template
 			#example="{ exampleIndex, variantIndex, exampleName, exampleProps }"
 		>
+			<button v-if="exampleProps?.behavior === 'interactive'">
```
mfranzke:
```suggestion
			<button type="button" aria-label="button copy" v-if="exampleProps?.behavior === 'interactive'">
```

showcases/react-showcase/src/components/card/index.tsx
```diff
@@ -10,14 +10,18 @@ const getCard = ({
...
+	const card = (
+		<DBCard
+			behavior={behavior}
+			spacing={spacing}
+			elevationLevel={elevationLevel}>
+			<strong>{children}</strong>
+		</DBCard>
+	);
+
+	return behavior === 'interactive' ? <button>{card}</button> : card;
```
mfranzke:
```suggestion
	return behavior === 'interactive' ? <button type="button" aria-label="button copy">{card}</button> : card;
```

showcases/angular-showcase/src/app/components/card/card.component.html
```diff
@@ -10,12 +10,24 @@
...
 	>
-		<db-card
-			[behavior]="exampleProps?.behavior"
-			[elevationLevel]="exampleProps?.elevationLevel"
-			[spacing]="exampleProps?.spacing"
-		>
-			<strong>{{ exampleName }}</strong>
-		</db-card>
+		@if (exampleProps?.behavior === "interactive") {
+			<button>
```
mfranzke:
```suggestion
			<button type="button" aria-label="button copy">
```

docs/migration/v2.x.x-to-v3.0.0.md
```diff
@@ -16,6 +16,30 @@ Please provide the correct `type` property in your code anyhow as a best practis
 We removed the `state` property from the `db-button`/`DBButton` component for now.
 It wasn't implemented in any framework causing some confusion, and we will reintroduce it in a future version.
 
+
+## db-card/DBCard
+
+### `behavior='interactive'` change
```
mfranzke, line 22: If we would ask the developers to wrap this component on their own, what for do we still need the `behavior="interactive"` property? This would be equivalent to any `button db-card` or `button db-link` structure.

nmerget: You are right, we could add this as well. But I would keep `behavior='interactive'` to stay aligned with design. Maybe there is another use-case with another HTML Tag, like an <input type="submit"> etc. so we don't remove this functionality

packages/foundations/dev/icons.html
mfranzke, line 1: @nmerget we need to investigate why this reformatting is happening from time to time. Are you committing with `--no-verify`? Elsewhere a file like this might have been regenerated, but should have been formatted via `prettier`.

packages/components/src/components/card/card.scss
```diff
@@ -3,58 +3,57 @@
...
-:is(a, button):has(.db-card) {
-	all: unset;
+:is(a, button) {
+	&:has(.db-card) {
+		all: unset;
+	}
+
+	@include helpers.hover {
+		> .db-card,
+		> db-card > .db-card {
```
mfranzke, line 13:
```
		.db-card {
```

would we have nested cards, or couldn't we use a simple selector as shown above ?

nmerget: This is the angular/stencil syntax because of the custom elements

mfranzke: I know, but if we wouldn't nest cards and/or wouldn't include several elements within a link or a button (in general a bad idea because of a huge amount of content / accessible name for a link/button, than writing the selector like the following would work for both cases (stencil/angular and the others), wouldn't it ?

```scss
:is(a, button) {
	@include helpers.hover {
		.db-card {
```

which would result in
```css
:is(a, button) .db-card
```

docs/migration/v2.x.x-to-v3.0.0.md
```diff
@@ -16,6 +16,30 @@ Please provide the correct `type` property in your code anyhow as a best practis
...
+
+We changed the `behavior='interactive'` property not applying `role='button'` and `tabIndex` anymore.
+If you want to use an interactive card, you should wrap the card with the correct HTML element, like a `button` or `a` tag:
+
+```html
+<!--Angular-->
+<button type="button" aria-label="button copy">
```
mfranzke:
```suggestion
<button type="button">
```

Probably this aspect is too complex to explain, so we should remove this part.

docs/migration/v2.x.x-to-v3.0.0.md
```diff
@@ -16,6 +16,30 @@ Please provide the correct `type` property in your code anyhow as a best practis
...
+	<db-card behavior="interactive">
+		<!-- card content -->
+	</db-card>
+</button>
+
+<!--React/Vue-->
+<button type="button" aria-label="button copy">
```
mfranzke:
```suggestion
<button type="button">
```

Probably this aspect is too complex to explain, so we should remove this part.

showcases/angular-showcase/src/app/components/card/card.component.html
```diff
@@ -10,12 +10,24 @@
...
 	>
-		<db-card
-			[behavior]="exampleProps?.behavior"
-			[elevationLevel]="exampleProps?.elevationLevel"
-			[spacing]="exampleProps?.spacing"
-		>
-			<strong>{{ exampleName }}</strong>
-		</db-card>
+		@if (exampleProps?.behavior === "interactive") {
+			<button type="button" aria-label="button copy">
```
mfranzke:
```suggestion
			<button type="button">
```

Probably this aspect is too complex to explain, so we should remove this part.

showcases/react-showcase/src/components/card/index.tsx
```diff
@@ -10,25 +10,33 @@ const getCard = ({
...
-	return (
-		<DefaultComponent
-			title={'DBCard'}
-			variants={getVariants(
-				defaultComponentVariants,
-				getCard,
-				props.slotCode
-			)}></DefaultComponent>
+	return behavior === 'interactive' ? (
+		<button type="button" aria-label="button copy">
```
mfranzke:
```suggestion
		<button type="button">
```

Probably this aspect is too complex to explain, so we should remove this part.

packages/components/src/components/card/card.scss
mfranzke: @leape is one allowed to nest cards into each other ?

mfranzke: @leape says: definitely maybe.

docs/migration/v2.x.x-to-v3.0.0.md
mfranzke: https://github.com/db-ux-design-system/core-web/issues/4577

PR #4522, opened 2025-07-10
fix: auto close for sub-navigation click

### LINKED ISSUE #4189: DBNavigation sub-navigation not closing after navigation is performed

- [ ] All
- [ ] HTML
- [ ] React
- [x] Angular
- [ ] Vue
- [ ] Web components
- [ ] Power Apps

### Reproduction case

When opening a sub-navigation menu from the DBNavigation Component, the sub-navigation menu doesn't close after the user has clicked on a navigation item. The sub-navigation menu only closes after the user clicks somewhere outside of the menu.

Image

### Expected Behaviour

When a user clicks on a sub-navigation menu item, the menu should close itself after the navigation has been performed.

### Screenshots

_No response_

### Browser version

None

_No response_

nmerget:
## Proposed changes

closes #4189

## Types of changes

### DIFF
--- a/packages/components/src/components/navigation-item/navigation-item.lite.tsx
+++ b/packages/components/src/components/navigation-item/navigation-item.lite.tsx
@@ -18,10 +18,7 @@ import {
 	getHideProp,
 	uuid
 } from '../../utils';
-import {
-	isEventTargetNavigationItem,
-	NavigationItemSafeTriangle
-} from '../../utils/navigation';
+import { NavigationItemSafeTriangle } from '../../utils/navigation';
 import DBButton from '../button/button.lite';
 import { DBNavigationItemProps, DBNavigationItemState } from './model';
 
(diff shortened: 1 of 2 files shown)


NearW (approved):

packages/components/src/components/navigation-item/navigation-item.lite.tsx
```diff
@@ -41,12 +38,12 @@ export default function DBNavigationItem(props: DBNavigationItemProps) {
 		autoClose: false,
 		subNavigationId: 'sub-navigation-' + uuid(),
 		navigationItemSafeTriangle: undefined,
-		handleNavigationItemClick: (event: unknown) => {
-			if (isEventTargetNavigationItem(event)) {
+		handleNavigationItemClick: (event: any) => {
+			if (event?.target?.nodeName === 'A') {
```
mfranzke, line 42:
```suggestion
			if (event?.target?.localName === 'a') {
```

packages/components/src/shared/model.ts
```diff
@@ -130,7 +130,7 @@ export type PlacementProps = {
 };
 
 export type NavigationBehaviorState = {
-	handleNavigationItemClick: (event: unknown) => void;
+	handleNavigationItemClick: (event: any) => void;
```
mfranzke, line 133: why do you switch from `unknown` to `any` ?

nmerget: Otherwise `event.target` wouldn't work

packages/components/src/components/navigation-item/navigation-item.lite.tsx
nmerget: https://stackoverflow.com/questions/2751556/what-is-the-best-way-to-get-an-elements-tag-name-in-js 

`nodeName` is totally fine

PR #4568, opened 2025-07-17
refactor: renamed `data-hide-icon` to `data-show-icon` to match property

nmerget:
## Types of changes


__snapshots__/checkbox/showcase/chromium/DBCheckbox-should-match-screenshot-1/DBCheckbox-should-match-screenshot.png
mfranzke, line 1: @nmerget I would assume that we would still want to have the icon get displayed next to "false" in the section "Show Label", wouldn't we? This change was mainly related to the "Show Icon" section, wasn't it?

mfranzke: as aligned, this hasn't worked correctly previously, there shouldn't be an icon, so it's actually correct now.

mfranzke (approved):

Issue #3372, opened 2024-10-26
Evaluate `pnpm`/`yarn` usage


mfranzke:
We should evaluate whether a `yarn` or `pnpm` usage might be beneficial instead of `npm`.

PR #4598, opened 2025-07-22
feat: verify `package-lock.json` UTD (up to date)

mfranzke:
## Proposed changes

From time to time we experience out of date `package-lock.json` files. To ensure that those aren't resulting out of our local development (as they are unlikely based on dependabot updates), we should ensure that those are quickly checked in `git push`.

## Types of changes


.husky/pre-push
```diff
@@ -0,0 +1,25 @@
...
+
+## Get the list of files changed between upstream and HEAD
+FILES=$(git diff --name-only $UPSTREAM..HEAD)
+
+## Check if package.json or package-lock.json is in the list
+if echo "$FILES" | grep -qE '^(package\.json|package-lock\.json)$'; then
+  echo "Detected changes in package.json or package-lock.json"
+
+  ## Run your npx command
+  npx --no package-lock-utd
```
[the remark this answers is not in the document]
mfranzke: this isn't incorrect syntax, but mainly telling the pipeline to take the version out of the project we're controlling via dependabot. It's the equivalent to pinning your dependencies.

nmerget:
I think this isn't a good solution.

The package you used just uses [`npm install --package-lock-only --ignore-scripts`](https://github.com/MBuchalik/package-lock-utd/blob/main/src/cli.ts#L38C13-L38C61) to compare the file contents. If we switch to `pnpm` this package wouldn't work.

Anyways I don't see the problem with `npm ci`. The first step in our pipeline would fail if the package-lock.json is out of date. This happens sometimes, but most of the time you have to install the dependency to use it in you branch.

mfranzke:
> I think this isn't a good solution.
> 
> The package you used just uses [`npm install --package-lock-only --ignore-scripts`](https://github.com/MBuchalik/package-lock-utd/blob/main/src/cli.ts#L38C13-L38C61) to compare the file contents. If we switch to `pnpm` this package wouldn't work.
> 
> Anyways I don't see the problem with `npm ci`. The first step in our pipeline would fail if the package-lock.json is out of date. This happens sometimes, but most of the time you have to install the dependency to use it in you branch.
> 
> Instead of adding this we should consider solving #3372 and move to pnpm. A lot of current problems would be solved with pnpm

I mainly agree on wanting to proceed with using the package, but do like the idea of an additional check previous to a push, as we had some more `npm ci` fails in the pipelines previously, which would have been prevented by that conditional qa-test on our localhosts, that doesn't hurt anybody.

.husky/pre-push
```diff
@@ -0,0 +1,78 @@
+# Configuration: Define checks as functions for better maintainability
+# Each check function should:
+# - Define a PATTERN variable for file matching
+# - Define a COMMAND variable for the command to run
+# - Define a DESCRIPTION variable for user feedback
+
+check_npm_files() {
+  PATTERN='^(package\.json|package-lock\.json)$'
```
michaelmkraus, line 8:
```suggestion
  PATTERN='(^|/)(package\.json|package-lock\.json)$'
```

The pattern only match at repo root. Other workspaces are not covered. The recommended pattern should also reach the workspaces

.husky/pre-push
```diff
@@ -0,0 +1,78 @@
...
+# Check for changes in specified files before pushing and run corresponding commands
+## Get the upstream branch
+UPSTREAM=$(git rev-parse --abbrev-ref --symbolic-full-name @{u} 2>/dev/null || echo "")
+if [ -z "$UPSTREAM" ]; then
+  echo "No upstream configured, detecting default branch."
+  # Try to detect the default branch from origin/HEAD
+  DEFAULT_BRANCH=$(git symbolic-ref refs/remotes/origin/HEAD 2>/dev/null | sed 's@^refs/remotes/origin/@@')
+  if [ -z "$DEFAULT_BRANCH" ]; then
+    echo "Could not detect default branch, falling back to 'main'."
+    DEFAULT_BRANCH="main"
```
michaelmkraus, line 43: What if local main branch is not be uptodate?

.config/.lintstagedrc.js
```diff
@@ -8,5 +8,7 @@ export default {
 	'stylelint.config.*': 'stylelint --validate --allow-empty-input',
 	// And elsewhere we don't, compare to https://github.com/stylelint/stylelint/pull/8009
 	'*.{css,scss}': 'stylelint --fix --allow-empty-input --no-validate',
-	'*.{js,ts,tsx,jsx,mjs,cjs}': 'xo --fix'
+	'*.{js,ts,tsx,jsx,mjs,cjs}': 'xo --fix',
+	// ensure that security vulnerabilities are fixed before committing
+	'package-lock.json': 'npm audit fix'
```
michaelmkraus, line 13: This might be very slow. Maybe it could be moved to CI

.husky/pre-push
```diff
@@ -0,0 +1,78 @@
...
+
+## Get the list of files changed between upstream and HEAD
+FILES=$(git diff --name-only "$UPSTREAM"..HEAD)
+
+## Check each pattern and run corresponding command
+for check_function in "${CHECK_FUNCTIONS[@]}"; do
+  # Call the check function to set variables
+  $check_function
+
+  if echo "$FILES" | grep -qE "$PATTERN"; then
```
mfranzke:
```suggestion
  if echo "$FILES" | grep --quiet --extended-regexp --recursive "$PATTERN"; then
```

michaelmkraus (approved):

PR #4624, opened 2025-07-27
feat: added native select styling

### LINKED ISSUE #3654: The position of the options box in the select element looks misaligned.

- [ ] All
- [X] HTML
- [ ] React
- [ ] Angular
- [ ] Vue
- [ ] Web components
- [ ] Power Apps

### Reproduction case

The position of the options always seems off. It is probably automatically positioned by the browser in the top left, which can look broken depending on the browser.

Is there anything that can be done to optimize this?

### Expected Behaviour

I would expect the select options aligned with the select field and not the focus frame.

### Screenshots

image
Bild

### Browser version

None

https://db-ui.github.io/mono/review/main/components/data-input/select/overview

mfranzke:
## Proposed changes

I consolidated the "layer" styling for the opened select options so the CustomSelect dropdown and the native styled `select` (`::picker(select)`) render identically, all driven from `_select-components.scss`.

### What changed:

- `_select-components.scss` now owns the full shared layer definition:
  - `%select-dropdown-container` gained the surface treatment that was previously duplicated inline in `select.scss` and only partly present in the CustomSelect side: it now `@extend`s `%default-card` (border + radius) and sets `background-color`, on top of the existing `position`, `z-index`, `box-shadow`, and sizing.
  - Replaced the `%select-list-max-block-size` placeholder with a new `select-list-styles` **mixin** that bundles the scrollable list styling: `padding`, `overflow-y`/`overflow-block`, the `max-block-size` cap (5.5 items to hint at more content), and the `::-webkit-scrollbar-button` tweak.
  - `%select-list-item` (option box) was already shared and stays as is.
- `custom-select-list.scss` now `@include`s `select-components.select-list-styles` instead of hand-rolling padding/overflow/scrollbar/max-height, and dropped the now-unused `variables` import.
- `select.scss`: the native `::picker(select)` now just `@extend %select-dropdown-container` + `@include select-components.select-list-styles`, removing the "Adapted from ..." inline duplicates (`%default-card`, `background`, `padding`, `overflow-block`, separate `max-block-size`).
- `custom-select-dropdown.scss` is unchanged in intent; it already extends `%select-dropdown-container` and carries `.db-card` in markup, so it now stays in lockstep with the picker.

Why a mixin for the list (not a placeholder): `.db-custom-select-list` applies `all: unset` in its own rule. `@extend` output is emitted before that rule's own declarations in the cascade, so a placeholder's `padding`/`overflow` would be wiped by `all: unset`. A mixin emits the declarations inline right after `all: unset`, so they win. I verified this in the compiled `custom-select-list.css`.

Verified: full `pnpm run build` passes, stylelint passes on all four files, no diagnostics, and I confirmed in the compiled CSS that both `::picker(select)` and `.db-custom-select-dropdown` now emit the same border/radius/background/elevation/position/sizing and list padding/overflow/max-height.

Resolves https://github.com/db-ux-design-system/core-web/issues/3654

## Types of changes

## Further comments

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/feat-multi-select-native-select-styling-new>


packages/components/src/components/select/select.scss
```diff
@@ -45,4 +45,41 @@
...
+			justify-content: space-between;
+			// change order of the included text and pseudo-element within this flexbox item, to make sure the icon is always on the right
+			flex-direction: row-reverse;
+
+			&:checked {
+				font-weight: 700;
+			}
+
+			/* stylelint-disable-next-line selector-pseudo-element-no-unknown */
+			&::checkmark {
```
mfranzke:
```suggestion
			:is(&::checkmark) {
```

packages/components/src/components/select/select.scss
```diff
@@ -45,4 +45,41 @@
...
+			appearance: base-select;
+		}
+
+		/* stylelint-disable-next-line selector-pseudo-element-no-unknown */
+		&::picker-icon {
+			display: none;
+		}
+
+		/* stylelint-disable-next-line selector-pseudo-element-no-unknown */
+		::picker(select) {
```
mfranzke:
```suggestion
		:is(::picker(select)) {
```

michaelmkraus:
The PR #5586 can probably be closed. The shared-placeholder abstraction it proposed has been implemented directly in this PR.

.changeset/native-select-picker-styling.md
```diff
@@ -0,0 +1,5 @@
+---
+"@db-ux/core-components": patch
+---
+
+refactor: consolidate CustomSelect and native Select element styling so the opened options layer looks identical
```
mfranzke, line 5:
```suggestion
refactor(DBSelect): consolidate CustomSelect and native Select element styling so the opened options layer looks identical
```

PR #4690, opened 2025-08-06
refactor: css condition to scss condition

mfranzke:
## Proposed changes

Currently we've checked for the elements to be either `input` or `textarea` at the CSS level with `:is()`. Instead as we're using those in a SCSS mixin, we should that for check for those element types in a SCSS condition.

## Types of changes

### DIFF
--- a/packages/components/src/styles/internal/_form-components.scss
+++ b/packages/components/src/styles/internal/_form-components.scss
@@ -372,7 +372,7 @@ $input-valid-types:
 			background-color: colors.$db-adaptive-bg-basic-transparent-semi-hovered;
 		}
 
-		&:is(input, textarea) {
+		@if ($selector == input or $selector == textarea) {
 			/* see https://developer.mozilla.org/en-US/docs/Web/CSS/field-sizing */
 			&[data-field-sizing="content"] {
 				field-sizing: content;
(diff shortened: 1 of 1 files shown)


packages/components/src/styles/internal/_form-components.scss
```diff
@@ -9,6 +10,11 @@
 
 @forward "../visually-hidden";
 
+// https://developer.mozilla.org/en-US/docs/Web/CSS/:read-only
+$valid-read-only-form-fields: input, textarea;
+// https://developer.mozilla.org/en-US/docs/Web/CSS/field-sizing
+$valid-field-sizing-form-fields: input, textarea;
+
```
michaelmkraus, line 17: Currently, both $valid-field-sizing-form-fields and $valid-read-only-form-fields contain the same elements (input, textarea). Do we need to keep them separate for semantic clarity, or could we merge them into a single list to reduce redundancy?

packages/components/src/styles/internal/_form-components.scss
```diff
@@ -383,11 +389,13 @@ $input-valid-types:
...
 
-		&:is(input, textarea):not(:disabled):read-only {
-			background-color: var(
-				--db-textarea-read-only,
-				#{colors.$db-adaptive-bg-basic-level-1-default}
-			) !important;
+		@if list.index($valid-read-only-form-fields, $selector) {
+			&:not(:disabled):read-only {
+				background-color: var(
+					--db-#{$selector}-read-only,
```
michaelmkraus, line 395: This was --db-textarea-read-only before and is now element-specific. I think that is fine.

packages/components/src/styles/internal/_form-components.scss
mfranzke: I was struggling with this as well. Perhaps there is no clear right or wrong, but when it comes to choosing between DRY and distinguishing between two completely different CSS features, the latter may be slightly more relevant.

packages/components/src/styles/internal/_form-components.scss
mfranzke: > This was --db-textarea-read-only before and is now element-specific. I think that is fine.

In general this isn't a variable we're providing, but only something we provide as a "Styling API". So scoping this to the "correct" component (`input` instead of `textarea`) might even be benefical.

Issue #4671, opened 2025-08-04
DBTag: Tooltip's JS animation leads to breaking tests


annsch:
### Which generators are affected?

- [ ] All
- [ ] HTML
- [x] React
- [ ] Angular
- [ ] Vue
- [ ] Web components
- [ ] Power Apps

### Reproduction case

Our component has a DBTag with onRemove prop. 

```
const firstTagCloseBtn = filterPanelFirstTag.locator('.db-tab-remove-button');
await firstTagCloseBtn.click();
```

In previous versions of DB UX Design System this worked correctly. From v3.x.x we have to use
 it with .hover() otherwise our tests fails because the click cannot be resolved.

```
const firstTagCloseBtn = filterPanelFirstTag.locator('.db-tab-remove-button');
await firstTagCloseBtn.hover();
await firstTagCloseBtn.click();
```

Debugging the DBTag component shows that there a JS style that sets the correct Tooltip position. We assume that playwright is faster than any animation and this leeds to the tooltip overlaying the remove button.

nmerget:
@annsch We add a test [here](https://github.com/db-ux-design-system/core-web/pull/4696) and it works as expected.

Can you add some screenshot, further infromation or a showcase to reproduce it?

Did you disabled all animations for your tests?
You might be able to use `.click({force:true})` as a workaround.

There is a 1MS [delay](https://github.com/db-ux-design-system/core-web/blob/main/packages/components/src/components/tooltip/tooltip.lite.tsx#L67) for the Tooltip position, this happens on "mouseenter" or "focusin" - this could cause the trouble.

PR #4696, opened 2025-08-06
fix(tooltip & popover): playwright visual regression tests

mfranzke:
## Proposed changes

Resolves https://github.com/db-ux-design-system/core-web/issues/4671

## Types of changes


packages/components/src/components/tag/tag.spec.tsx
mfranzke, line 1: @nmerget this PR now only includes a test, how would it solve Annas problem out of https://github.com/db-ux-design-system/core-web/issues/4671?

nmerget: I talked with Anna and add some solutions to the Issue as comment. She is fine by closing the issue.

What we can do is to test the functionality of the event. But everything else is custom code and the developers are responsible for a working test. The `onRemove` event works as expected.

Issue #4385, opened 2025-06-16
Secondary action not reachable on mobile in DBHeader without navigation


psperber:
### Which generators are impacted?

- [x] All
- [ ] HTML
- [ ] React
- [ ] Angular
- [ ] Vue
- [ ] Web components
- [ ] Power Apps

### Reproduction case

Create a DBHeader without a Navigation (no children), but a secondary action.
Open mobile view of the application.

### Expected Behaviour

Hamburger menu is clickable and the secondary action is reachable via drawer

### Screenshots

Desktop View:
![Image](https://github.com/user-attachments/assets/a2b4f402-732d-4a7b-b7f9-d6d5f92362d3)

Mobile View:
![Image](https://github.com/user-attachments/assets/25dc6e04-45fa-466f-8814-3f9e97fa744b)

### Browser version

None

PR #4792, opened 2025-08-20
fix: Keep burger menu visible when drawer has content

michaelmkraus:
* Keep burger visible when drawer has content, hide if nav, meta and secondary are empty

closes https://github.com/db-ux-design-system/core-web/issues/4385

## Types of changes

### DIFF
--- a/packages/components/src/components/header/header.scss
+++ b/packages/components/src/components/header/header.scss
@@ -53,16 +53,21 @@
 		}
 	}
 
-	// Hide the burger menu if there is no navigation
-	&:has(.db-header-navigation:empty) {
+	// Hide the burger ONLY when the drawer would be empty.
+	// This requires: no main navigation AND no meta navigation AND no secondary action.
+	&:has(.db-header-navigation:empty):has(
+			.db-header-meta-navigation:empty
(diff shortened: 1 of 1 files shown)


packages/components/src/components/header/header.scss
```diff
@@ -53,16 +53,21 @@
 		}
 	}
 
-	// Hide the burger menu if there is no navigation
-	&:has(.db-header-navigation:empty) {
+	// Hide the burger ONLY when the drawer would be empty.
+	// This requires: no main navigation AND no meta navigation AND no secondary action.
+	&:has(.db-header-navigation:empty):has(
+			.db-header-meta-navigation:empty
+		):has(.db-header-secondary-action:empty) {
```
mfranzke, line 60:
```suggestion
	&:has(.db-header-navigation:empty, .db-header-meta-navigation:empty, .db-header-secondary-action:empty) {
```

You could [concatenate those in a selector-list within one `:has`](https://developer.mozilla.org/en-US/docs/Web/CSS/Selector_list#relative_selector_list).

mfranzke (approved):

PR #4861, opened 2025-08-28
feat: Switch stable rework

michaelmkraus:
## Proposed changes

* enable in/valid state for switch

## Types of changes

## Further comments

closes https://github.com/db-ux-design-system/core-team/issues/851

### DIFF
--- a/packages/components/src/components/switch/switch.lite.tsx
+++ b/packages/components/src/components/switch/switch.lite.tsx
@@ -1,5 +1,6 @@
 import {
 	onMount,
+	onUpdate,
 	Show,
 	useDefaultProps,
 	useMetadata,
@@ -8,18 +9,29 @@ import {
 	useTarget
 } from '@builder.io/mitosis';
 
+import {
+	DEFAULT_INVALID_MESSAGE,
+	DEFAULT_INVALID_MESSAGE_ID_SUFFIX,
+	DEFAULT_MESSAGE_ID_SUFFIX,
+	DEFAULT_VALID_MESSAGE,
+	DEFAULT_VALID_MESSAGE_ID_SUFFIX
+} from '../../shared/constants';
 import { ChangeEvent, InteractionEvent } from '../../shared/model';
 import {
 	cls,
+	delay,
 	getBoolean,
 	getBooleanAsString,
 	getHideProp,
+	hasVoiceOver,
+	stringPropVisible,
 	uuid
 } from '../../utils';
 import {
 	handleFrameworkEventAngular,
 	handleFrameworkEventVue
 } from '../../utils/form-components';
+import DBInfotext from '../infotext/infotext.lite';
 import { DBSwitchProps, DBSwitchState } from './model';
 
 useMetadata({
@@ -38,6 +50,55 @@ export default function DBSwitch(props: DBSwitchProps) {
 	// jscpd:ignore-start
 	const state = useStore<DBSwitchState>({
 		_id: undefined,
+		_messageId: undefined as string | undefined,
+		_validMessageId: undefined as string | undefined,
+		_invalidMessageId: undefined as string | undefined,
+		_invalidMessage: undefined as string | undefined,
+		_descByIds: '' as string,
+		_voiceOverFallback: '' as string,
+
+		hasValidState: () => {
+			return !!(props.validMessage ?? props.validation === 'valid');
+		},
+		handleValidation: () => {
+			if (!_ref?.validity?.valid || props.validation === 'invalid') {
+				state._descByIds = state._invalidMessageId!;
+				state._invalidMessage =
+					props.invalidMessage ||
+					_ref?.validationMessage ||
+					DEFAULT_INVALID_MESSAGE;
+				if (hasVoiceOver()) {
+					state._voiceOverFallback =
+						state._invalidMessage || DEFAULT_INVALID_MESSAGE;
+					delay(() => {
+						state._voiceOverFallback = '';
+					}, 1000);
+				}
+				return;
+			}
+			if (
+				state.hasValidState() &&
+				_ref?.validity?.valid &&
+				props.required
+			) {
+				state._descByIds = state._validMessageId!;
+				if (hasVoiceOver()) {
+					state._voiceOverFallback =
+						props.validMessage ?? DEFAULT_VALID_MESSAGE;
+					delay(() => {
+						state._voiceOverFallback = '';
+					}, 1000);
+				}
+				return;
+			}
+
+			if (stringPropVisible(props.message, props.showMessage)) {
+				state._descByIds = state._messageId!;
+				return;
+			}
+
+			state._descByIds = '';
+		},
 		handleChange: (event: ChangeEvent<HTMLInputElement>) => {
 			if (props.onChange) {
 				props.onChange(event);
@@ -48,6 +109,7 @@ export default function DBSwitch(props: DBSwitchProps) {
 					handleFrameworkEventAngular(state, event, 'checked'),
 				vue: () => handleFrameworkEventVue(() => {}, event, 'checked')
 			});
+			state.handleValidation();
 		},
 		handleBlur: (event: InteractionEvent<HTMLInputElement>) => {
 			if (props.onBlur) {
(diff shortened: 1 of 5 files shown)


packages/components/src/components/switch/switch.lite.tsx
```diff
@@ -30,14 +42,65 @@ useMetadata({
...
 	}
 });
-useDefaultProps<DBSwitchProps>({});
+useDefaultProps<DBSwitchProps>({
+	labelPosition: 'trailing'
+});
 
 export default function DBSwitch(props: DBSwitchProps) {
 	// This is used as forwardRef
 	const _ref = useRef<HTMLInputElement | any>(null);
```
sarahbrng, line 49: `const _ref = useRef<HTMLInputElement>(null);`

packages/components/src/components/switch/switch.lite.tsx
```diff
@@ -84,8 +166,9 @@ export default function DBSwitch(props: DBSwitchProps) {
 				checked={getBoolean(props.checked, 'checked')}
 				value={props.value}
 				disabled={getBoolean(props.disabled, 'disabled')}
-				aria-invalid={props.validation === 'invalid'}
-				data-custom-validity={props.validation}
+				aria-invalid={
+					props.validation === 'invalid' ? 'true' : undefined
+				}
 				name={props.name}
```
sarahbrng: `aria-describedby={state._descByIds || undefined}`

packages/components/src/components/switch/switch.scss
```diff
@@ -32,9 +31,82 @@ $checked-active-transition-size: calc(
...
+				background-color: colors.$db-successful-on-bg-inverted-default;
+				color: colors.$db-successful-on-bg-inverted-default;
+			} @else if $key == "invalid" {
+				background-color: colors.$db-critical-on-bg-inverted-default;
+				color: colors.$db-critical-on-bg-inverted-default;
+			}
+		}
+	}
+}
+
```
sarahbrng: Der mixin hat große if-else-Blöcke, in denen fast identische CSS-Selektoren wiederholt werden. Das kann über Variablen und Verschachtelungen vereinfacht bzw. verkürzt werden.

packages/components/src/components/switch/switch.scss
```diff
@@ -241,4 +313,78 @@ $checked-active-transition-size: calc(
...
+		--db-switch-label-color: #{colors.$db-successful-on-bg-basic-emphasis-80-default};
+	}
+
+	&[data-custom-validity="invalid"] {
+		--db-switch-label-color: #{colors.$db-critical-on-bg-basic-emphasis-80-default};
+	}
+
+	/* Set color for label */
+	> .db-switch-label,
+	> .db-switch-label * {
```
sarahbrng: ist das * notwendig?

packages/components/src/components/switch/switch.scss
```diff
@@ -241,4 +313,78 @@ $checked-active-transition-size: calc(
...
+
+			@include helpers.hover {
+				background-color: colors.$db-successful-on-bg-basic-emphasis-70-default;
+				border-color: colors.$db-successful-on-bg-basic-emphasis-70-default;
+			}
+
+			@include helpers.active {
+				background-color: colors.$db-successful-on-bg-basic-emphasis-70-default;
+				border-color: colors.$db-successful-on-bg-basic-emphasis-70-default;
+			}
```
sarahbrng: @include helpers.hover, @include helpers.active {
                background-color: colors.$db-successful-on-bg-basic-emphasis-70-default;
                border-color: colors.$db-successful-on-bg-basic-emphasis-70-default;
            }

michaelmkraus: Unfortunately, mixins can"t be combined that way, only selectors.

packages/components/src/components/switch/switch.scss
michaelmkraus: It was a remnant of an earlier version. Good catch. I removed it.

packages/components/src/components/switch/switch.lite.tsx
michaelmkraus: Good point. I added the attribute.

packages/components/src/components/switch/switch.scss
michaelmkraus: Valid point. I simplified the validity mixin. But to use variables within the mixin, I had to disable the linting-rule that checks for valid colors and doesn"t allow variables.

mfranzke:
@michaelmkraus could you please add a changeset to this PR ?

packages/components/src/components/switch/switch.scss
```diff
@@ -32,11 +32,113 @@ $checked-active-transition-size: calc(
 	}
 }
 
+@mixin switch-validity-colored(
```
nmerget: Why don't you use the same code as we have for the checkbox by using a `$variant` to access the same token like this:

```@mixin get-validity-color-check($key: "valid") {
 $variant: successful;

 @if $key != "valid" {
  $variant: critical;
 }

--db-check-element-border-color: var(
  --db-#{$variant}-on-bg-basic-emphasis-70-default
 );
```

packages/components/src/components/switch/switch.scss
```diff
@@ -243,4 +338,84 @@ $checked-active-transition-size: calc(
...
+	}
+
+	&[data-custom-validity="invalid"] {
+		--db-switch-label-color: #{colors.$db-critical-on-bg-basic-emphasis-80-default};
+	}
+
+	/* Set color for label */
+	> .db-switch-label {
+		color: var(--db-switch-label-color, inherit);
+		flex: 1 1 0;
```
nmerget: I think it makes sense to use `display: grid` with some `grid-template-areas` and `grid-area` instead. You can checkout the notification as reference.

packages/components/src/components/switch/switch.scss
```diff
@@ -243,4 +338,84 @@ $checked-active-transition-size: calc(
...
+	/* Set color for helper message */
+	&::after {
+		color: var(
+			--db-switch-label-color,
+			var(--db-check-element-label-color, inherit)
+		);
+	}
+
+	/* Put infotext below switch */
+	& :is(.db-infotext, db-infotext) {
```
nmerget: https://github.com/db-ux-design-system/core-web/pull/4861/files#r2416509673

packages/components/src/shared/model.ts
```diff
@@ -95,6 +95,16 @@ export type IconTrailingProps = {
...
 };
 
+export const LabelPositionList = ['leading', 'trailing'] as const;
+export type LabelPositionType = (typeof LabelPositionList)[number];
+
+export type LabelPositionProps = {
+	/**
+	 * Position of the label.
+	 */
+	labelPosition?: LabelPositionType;
```
nmerget: 2 Questions: 

1. We use `variant` for input/select/textarea to change the position/behavior of the label. Did you ask @leape why we use a different property for this case, should we align it?
2. No matter how we name it, this would be useful for checkbox and radio as well, wouldn't it?

michaelmkraus: 1. Do you recommend to extend the Variant to `['above', 'floating', 'leading', 'trailing']`? Wouldn't we have to support then all variants for each component?
2. I think it will be adapted for checkbox and radio in the future.

packages/components/src/components/switch/switch.scss
michaelmkraus: I switched to the same logic using the variant to calculate the colors.

packages/components/src/shared/model.ts
leape: 1. I think we can use the _variant_ property in this case and don't need to offer all possible values in all corresponding components.
I hadn't thought of that solution. 🙈 sry
If it's not too late to rename the property, we can do that @michaelmkraus 

2. yes - will be aligned of course

michaelmkraus: I will remove the `labelPosition` property and use the `variant`.

showcases/angular-showcase/src/app/components/switch/switch.component.html
```diff
@@ -17,14 +17,21 @@
...
 			[iconTrailing]="exampleProps?.iconTrailing"
+			[iconLeading]="exampleProps?.iconLeading"
 			[validation]="exampleProps?.validation"
+			[invalidMessage]="exampleProps?.invalidMessage"
+			[validMessage]="exampleProps?.validMessage"
+			[message]="exampleProps?.message"
+			[showMessage]="exampleProps?.showMessage"
+			[label]="exampleProps?.label"
+			>{{
+				exampleProps?.label ? exampleProps?.label : exampleName
```
mfranzke, line 32: @nmerget those innerText differ from the other JS frameworks (`showcases/vue-showcase/src/components/switch/Switch.vue` and `showcases/react-showcase/src/components/switch/index.tsx`), is this intentionally?

PR #4931, opened 2025-09-06
refactor: remove CodeQL advanced analysis config files from PRs workflow

mfranzke:
## Proposed changes

Removed CodeQL analysis workflow from pull request actions. This advanced configuration was causing several problems:
- It was only scanning JavaScript, but not GitHub Actions code.
- It was missing onPush hooks, which lead to the problem that it's findings couldn't get correctly displayed in GitHub Security area.

[Switched to the much simpler, default configuration](https://github.com/db-ux-design-system/core-web/settings/security_analysis), that solves the above problems.

This would even also solve topics like e.g. using the `github/codeql-action/autobuild` action directly:

> Actions with special purposes and unlikely to be used directly:
> autobuild: Attempts to automatically build the code. Only used for analyzing languages that require a build. Use the build-mode: autobuild input in the init action instead. For information about input parameters, see the [autobuild action definition](https://github.com/github/codeql-action/blob/main/autobuild/action.yml).

source: https://github.com/github/codeql-action/?tab=readme-ov-file#actions

## Types of changes


.github/workflows/99-codeql-analysis.yml
```diff
@@ -1,28 +0,0 @@
...
-      - name: ⏬ Checkout repo
-        uses: actions/checkout@v5
-
-      - name: 🔄 Initialize CodeQL
-        uses: github/codeql-action/init@v3
-        with:
-          languages: ${{ matrix.language }}
-
-      - name: 🔨 Autobuild
-        uses: github/codeql-action/autobuild@v3
```
mfranzke, line 25: > Actions with special purposes and unlikely to be used directly:

> - autobuild: Attempts to automatically build the code. Only used for analyzing languages that require a build. Use the build-mode: autobuild input in the init action instead. For information about input parameters, see the [autobuild action definition](https://github.com/github/codeql-action/blob/main/autobuild/action.yml).

source: https://github.com/github/codeql-action/?tab=readme-ov-file#actions

PR #4975, opened 2025-09-15
refactor(ci/cd): extend to two parallel dependabot PRs

mfranzke:
## Proposed changes

We'd like to not just run one but two dependabot PRs in parallel (e.g. if one pipeline is running after the merge to `main`) and that for divide them into two groups.

## Types of changes


.github/workflows/default.yml
```diff
@@ -7,12 +7,12 @@ on:
...
       - "main"
 
-# We'd like to run the pipeline for Dependabot PRs sequentially, and all other PRs in parallel, but still only one action run for each PR.
+# We'd like to run the pipeline for Dependabot PRs with up to 2 in parallel, and all other PRs in parallel, but still only one action run for each PR.
 concurrency:
   group: >-
     ${{
       github.actor == 'dependabot[bot]' &&
-      'dependabot' ||
+      format('dependabot-{0}', github.event.number % 2) ||
```
michaelmkraus:
```suggestion
      format('dependabot-{0}', github.event.pull_request.number % 2) ||
```
Github uses this PR-specific field. It is better readable and more specific.

mfranzke: @copilot what's your opinion about using `github.pull_request.number` instead of `github.event.number`, what's the difference in between both? Please provide some URL to the GitHub documentation for those two as well.

mfranzke: @michaelmkraus to my understanding this would be `github.event.pull_request.number` than instead of `github.pull_request.number`, wouldn't it ?

michaelmkraus: You are right!

mfranzke: of course I am.

mfranzke:
We sadly cannot use modulo or arithmetic in workflow expressions.

PR #4989, opened 2025-09-17
refactor: removed obsolete code

mfranzke:
## Proposed changes

Resolves https://github.com/db-ux-design-system/core-web/issues/4988

## Types of changes

### DIFF
--- a/showcases/react-showcase/src/components/default-component.tsx
+++ b/showcases/react-showcase/src/components/default-component.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useState } from 'react';
+import { useCallback, useState } from 'react';
 import { DBCard, DBDivider, DBLink } from '../../../../output/react/src';
 import type {
 	ReactDefaultComponentProps,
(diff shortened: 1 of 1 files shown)


showcases/react-showcase/src/components/default-component.tsx
```diff
@@ -115,27 +115,8 @@ const DefaultComponent = ({
...
-		event: React.MouseEvent<HTMLAnchorElement>,
-		variantName: string
-	) => {
-		if (
-			typeof globalThis === 'undefined' ||
-			!globalThis.location.origin ||
-			!globalThis.location.href
-		) {
-			return;
+		if (typeof globalThis === 'undefined') {
```
mfranzke:
```suggestion
		if (typeof globalThis === 'undefined' || !globalThis.location.href) {
```

showcases/react-showcase/src/components/default-component.tsx
```diff
@@ -196,9 +177,6 @@ const DefaultComponent = ({
 								className="link-headline"
 								content="external"
 								target="_blank"
```
michaelmkraus, line 181:
```suggestion
								target="_blank"
								rel="noopener noreferrer"
```
For security and privacy reasons we could add the [rel-attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/rel/noopener).

showcases/react-showcase/src/components/default-component.tsx
```diff
@@ -115,27 +115,8 @@ const DefaultComponent = ({
...
-			typeof globalThis === 'undefined' ||
-			!globalThis.location.origin ||
-			!globalThis.location.href
-		) {
-			return;
+		if (typeof globalThis === 'undefined' || !globalThis.location.href) {
+			return '';
 		}
 
 		const currentUrl = globalThis.location.href.split('?');
```
michaelmkraus:
```suggestion
		const [baseUrl, query=''] = globalThis.location.href.split('?');
```
We could use it the following lines like this instead of using the index:

`const searchParameters = new URLSearchParams(query);`

showcases/react-showcase/src/components/default-component.tsx
```diff
@@ -115,41 +115,22 @@ const DefaultComponent = ({
...
-		) {
-			return;
-		}
-
-		const currentUrl = globalThis.location.href.split('?');
-		const rawComponentUrl = currentUrl[0];
-		const searchParameters = new URLSearchParams(currentUrl[1] ?? '');
+		const [baseUrl, query=''] = globalThis.location.href.split('?');
+		const rawComponentUrl = baseUrl;
+		const searchParameters = new URLSearchParams(query);
```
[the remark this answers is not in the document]
mfranzke: we handle it at line 122 already.

michaelmkraus (approved):

PR #5090, opened 2025-09-29
Refactor added missing entries

mfranzke:
## Proposed changes

Added missing entries to one of our "private" `package.json` files for compliance with our linting tools.

## Types of changes

### DIFF
--- a/e2e/package.json
+++ b/e2e/package.json
@@ -3,6 +3,11 @@
   "version": "0.0.0",
   "type": "module",
   "description": "",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/db-ux-design-system/core-web.git"
+  },
+  "license": "Apache-2.0",
   "private": true,
   "workspaces": [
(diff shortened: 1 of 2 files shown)


e2e/package.json
```diff
@@ -3,6 +3,11 @@
   "version": "0.0.0",
   "type": "module",
   "description": "",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/db-ux-design-system/core-web.git"
```
michaelmkraus, line 8: I am not sure if we should use the url without git+.
And, we can also define the workspace in a monorepo by eg
`"directory": "packages/components"`

mfranzke: Are you suggesting to define the URL without the prefix or to leave it as implemented, by having that prefix included in the "url" field ? This way it would be equal to the other `package.json` files in this repo. And having this form instead of starting with `https://` directly is some feedback we got from `publint ` I think to remember …

Issue #5103, opened 2025-09-30
Integrate `cooldown` for dependabot


mfranzke:
We'd like to ensure that packages are released for at least ~24~ 72 hours, to decrease recent risks regarding malicious packages.

24 hours seems to be the bare minimum of what makes sense to set for this delay. In discussion I even heard suggestions for up to a week, which seems to be far too long especially for bug fixes. So we should start with something in between.
 
More information:
- https://github.blog/changelog/2025-07-01-dependabot-supports-configuration-of-a-minimum-package-age/
- https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#cooldown-

PR #4631, opened 2025-07-28
refactor: migrated package manager to pnpm

mfranzke:
## Proposed changes

Migrated node package manager from npm to pnpm.

Resolves https://github.com/db-ux-design-system/core-web/issues/3372 and https://github.com/db-ux-design-system/core-web/issues/4311

- [ ] install pnpm in copilot setup
- [x] research for previously included restriction regarding an initial weight of an upload artifact (probably we need to increase some version number)

## Types of changes

- [ ] Bugfix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Refactoring (improvements to existing components or architectural decisions)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation Update (if none of the other choices apply)

## Further comments

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/refactor-migrated-package-manager-to-pnpm>

PR #5104, opened 2025-09-30
feat: added dependabot cooldown feature

mfranzke:
## Proposed changes

resolves https://github.com/db-ux-design-system/core-web/issues/5103

## Types of changes

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,6 +9,8 @@ updates:
       timezone: "Europe/Berlin"
     pull-request-branch-name:
       separator: "-"
+    cooldown:
+      default-days: 7
 
   # Daily: Check minor and patch updates
   - package-ecosystem: "npm"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -18,6 +20,8 @@ updates:
       interval: "daily"
       time: "23:00"
       timezone: "Europe/Berlin"
+    cooldown:
+      default-days: 3
```
michaelmkraus, line 24: It seems as if you could use values for specific releases. Might be interesting for security update.
```suggestion
      semver-major-days: 3
      semver-minor-days: 3
      semver-patch-days: 0
```

mfranzke: I thought about this aspect as well. However, this would counteract the whole idea because we want to delay updates. If we deactivate this for patches and everyone else does the same, providing a malicious version of a package as a patch would circumvent the entire concept.

Additionally, if we receive any notifications regarding security-related issues, dependabot will include them independently from "regular" patch releases and will also mention them in the Security tab here on GitHub. There would also be an override as soon as we use [pnpm](https://github.com/db-ux-design-system/core-web/pull/4631).

PR #5118, opened 2025-10-02
chore: 🤖 Add strict .npmrc file

dkolba:
## Proposed changes

Add .npmrc with sane defaults

## Types of changes

Put an `x` in the boxes that apply.
s9670_AreYouDrunk_COTT_BLK__48426


package.json
```diff
@@ -23,6 +23,9 @@
     "e2e"
   ],
   "packageManager": "npm@11.5.2",
+  "engines": {
+    "node": ">= 24"
+  },
```
mfranzke, line 28:
```suggestion
```

This would be redundant to `.nvmrc`

.npmrc
```diff
@@ -0,0 +1,4 @@
+package-lock=true
```
mfranzke, line 2:
```suggestion
# If set to false, then ignore package-lock.json files when installing. This will also prevent writing package-lock.json if save is true. source: https://docs.npmjs.com/cli/v11/using-npm/config
package-lock=true
```

.npmrc
```diff
@@ -0,0 +1,4 @@
...
+ignore-scripts=true
```
mfranzke, line 4:
```suggestion
# If true, npm does not run scripts specified in package.json files. source: https://docs.npmjs.com/cli/v11/using-npm/config#ignore-scripts
ignore-scripts=true
```

.npmrc
```diff
@@ -0,0 +1,4 @@
...
+save-exact=true
```
mfranzke, line 6:
```suggestion
# Dependencies saved to package.json will be configured with an exact version rather than using npm's default semver range operator. source: https://docs.npmjs.com/cli/v11/using-npm/config#save-exact
save-exact=true
```

.npmrc
```diff
@@ -0,0 +1,4 @@
...
+engine-strict=true
```
mfranzke, line 8:
```suggestion
# If set to true, then npm will stubbornly refuse to install (or even consider installing) any package that claims to not be compatible with the current Node.js version. source: https://docs.npmjs.com/cli/v11/using-npm/config#engine-strict
engine-strict=true
```

package.json
mfranzke: it's still nice to have this included in combination with `engine-strict=true` npm setting.

.npmrc
mfranzke: @dkolba does this `ignore-scripts` only related to not running scripts out of packages that we install? Or even also to the scripts we define in our context regarding `pre` and `post` node scripts to our node scripts ?

mfranzke: <html><head></head><body><p>Great question. As a Node.js web developer, you’re likely familiar with how powerful (and sometimes risky) the <code inline="">scripts</code> field in <code inline="">package.json</code> can be. Let's clarify exactly how the <code inline="">ignore-scripts</code> setting works.</p>
<hr>
<h3>🔧 What is <code inline="">ignore-scripts</code>?</h3>
<p><code inline="">ignore-scripts</code> is a configuration option for <strong>npm</strong> (and similarly in <strong>yarn</strong> or <strong>pnpm</strong>) that prevents <strong>lifecycle scripts</strong> from being executed <strong>during package installation</strong>.</p>
<hr>
<h3>✅ What It <em>Does</em> Affect</h3>
<p>It <strong>only</strong> affects <strong>scripts in packages you install</strong>, such as:</p>
<ul>
<li>
<p><code inline="">preinstall</code>, <code inline="">install</code>, <code inline="">postinstall</code> scripts in the <code inline="">package.json</code> of dependencies.</p>
</li>
<li>
<p>This includes any packages you install from the registry or via Git.</p>
</li>
</ul>
<p>For example:</p>
<pre><code class="language-json">// node_modules/some-package/package.json
"scripts": {
  "install": "node-gyp rebuild"
}
</code></pre>
<p>If <code inline="">ignore-scripts=true</code>, this <code inline="">install</code> script <strong>won’t run</strong> when installing this package.</p>
<hr>
<h3>❌ What It <em>Does Not</em> Affect</h3>
<p>It does <strong>not</strong> prevent your own scripts from running when <strong>you</strong> explicitly invoke them.</p>
<p>That includes:</p>
<ul>
<li>
<p>Any script you define in your own <code inline="">package.json</code>, like:</p>
<pre><code class="language-json">"scripts": {
  "build": "tsc",
  "prebuild": "echo Prebuild",
  "postbuild": "echo Postbuild"
}
</code></pre>
</li>
<li>
<p><code inline="">npm run build</code> will still execute <code inline="">prebuild</code>, <code inline="">build</code>, and <code inline="">postbuild</code> regardless of <code inline="">ignore-scripts</code>.</p>
</li>
</ul>
<hr>
<h3>🧠 Summary</h3>

Use Case | Affected by ignore-scripts
-- | --
Scripts in installed packages | ✅ Yes
Scripts in your own package.json | ❌ No
npm run <script> you execute | ❌ No


<hr>
<h3>🛡️ Why Use <code inline="">ignore-scripts</code>?</h3>
<p>It’s often used in:</p>
<ul>
<li>
<p><strong>CI environments</strong> to prevent unexpected behavior.</p>
</li>
<li>
<p><strong>Security-sensitive projects</strong> to block malicious install scripts from third-party packages.</p>
</li>
</ul>
<p>You can set it via:</p>
<pre><code class="language-bash">npm install --ignore-scripts
# or permanently
npm config set ignore-scripts true
</code></pre>
<p>Let me know if you want help locking down dependencies or auditing scripts safely.</p></body></html>

.npmrc
```diff
@@ -0,0 +1,8 @@
+# If set to false, then ignore package-lock.json files when installing. This will also prevent writing package-lock.json if save is true. source: https://docs.npmjs.com/cli/v11/using-npm/config
+package-lock=true
+# If true, npm does not run scripts specified in package.json files. source: https://docs.npmjs.com/cli/v11/using-npm/config#ignore-scripts
+ignore-scripts=true
```
mfranzke, line 4:
```suggestion
ignore-scripts=true
```

This would break essential tools, such as `husky`. We need to integrate a solution that allows us to provide whitelists.

mfranzke: My bad; the problems I experienced with running Husky seemed to have been related to this setting, but it's not.

PR #5121, opened 2025-10-02
feat(ci/cd): Add GitHub Action for auto-assigning PRs

mfranzke:
## Proposed changes

out of the recipes from https://github.com/actions-ecosystem/recipes this sounded very interesting. As we're sometimes forgetting to set ourselves as the assignee after opening a PR, this action would be very helpful.

## Types of changes


.github/workflows/auto-assign.yml
```diff
@@ -0,0 +1,19 @@
+name: Auto Assign
+permissions:
```
mfranzke, line 3:
```suggestion

permissions:
```

nmerget (approved):

PR #5134, opened 2025-10-03
fix(dependabot): added explicit update types for all dependencies

mfranzke:
## Proposed changes

It seems that currently, no PRs grouped by npmjs organisations would be opened, except the patch ones. According to ChatGPT, we should explicitly define the groups in the Dependabot configuration.

## Types of changes

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -109,6 +109,7 @@ updates:
           - "typescript"
           - "tsx"
           - "tslib"
+        # There is no entry for `major`, as these are configured to be ignored in the `ignore` section of this configuration file.
         update-types:
           - "minor"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -70,50 +70,85 @@ updates:
...
+        update-types:
+          - "minor"
+          - "major"
       # Development tools
       typescript:
         patterns:
           - "typescript"
           - "tsx"
           - "tslib"
+        update-types:
```
mfranzke, line 112:
```suggestion
        # There is no entry for `major`, as these are configured to be ignored in the `ignore` section of this configuration file.
        update-types:
```

PR #5140, opened 2025-10-06
chore: preview branch output inside description instead of comment

nmerget:
## Proposed changes

Add a preview URL only if the deployment was successful instead of adding a comment on opening the PR

## Types of changes

## Further comments

🔭🐙🐈 Test this branch here: https://db-ux-design-system.github.io/core-web/review/chore-preview-url


mfranzke:
To my understanding the URL would get added to the description each pipeline run after deploy. But it wouldn't change over time, as it's a consistant name. What would be the benefit of changing the current simple implementation to his more extended one, except the timing (adding it after a succesful deployment)?

nmerget:
> @nmerget As I understand it, the URL would be added to or replaced in the description after a successful deployment in each pipeline run. However, it wouldn't change over time because it's a consistent name. What would be the benefit of changing the current simple implementation to this more extended one, besides the timing of adding it after a successful deployment, and not already upfront? Our pages are always deployed to that URL, and if I want to ensure I get the latest deployment, I need to check the latest commit's green pipeline status anyhow.
> 
> (I'm not questioning the decision to move the URL from the comment to the PR description.)

Yes you are right, I think it is confusing to have a comment with a link which isn't available.

.github/workflows/03-preview-url-pr-description.yml
```diff
@@ -0,0 +1,22 @@
+name: Preview URL PR Description
+
+permissions:
+  pull-requests: write
+  contents: write
```
michaelmkraus:
```suggestion
  pull-requests: write
```
Isn"t pull-requests: write sufficient?

michaelmkraus (approved):

PR #5144, opened 2025-10-06
Reapply "chore: 🤖 Add strict .npmrc file (#5118)" (#5143)

mfranzke:
## Proposed changes

This reverts commit b81f49d16b448bfa55fc3167e70ab551747cc77a.

## Types of changes

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,8 @@
+# If set to false, then ignore package-lock.json files when installing. This will also prevent writing package-lock.json if save is true. source: https://docs.npmjs.com/cli/v11/using-npm/config
+package-lock=true
+# If true, npm does not run scripts specified in package.json files. source: https://docs.npmjs.com/cli/v11/using-npm/config#ignore-scripts
+ignore-scripts=true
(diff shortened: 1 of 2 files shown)


.npmrc
```diff
@@ -0,0 +1,8 @@
+# If set to false, then ignore package-lock.json files when installing. This will also prevent writing package-lock.json if save is true. source: https://docs.npmjs.com/cli/v11/using-npm/config
+package-lock=true
+# If true, npm does not run scripts specified in package.json files. source: https://docs.npmjs.com/cli/v11/using-npm/config#ignore-scripts
+ignore-scripts=true
```
mfranzke:
```suggestion
ignore-scripts=true
```

This setting also appears to prevent our `pre`- and `post`-scripts from running in our repository workspaces.

mfranzke:
we should merge https://github.com/db-ux-design-system/core-web/pull/4631 previous to this PR, as pnpm seems to allowed our own `pre`- and `post`-Node-scripts to still run even though that we set `ignore-scripts` to `true` (which is `pnpm`s default setting anyhow).

.npmrc
mfranzke: we're bounding this to pnpm, so we're fine.

PR #5159, opened 2025-10-07
fix: select placeholder jumping with label above

### LINKED ISSUE #5139: DBSelect: Jumping placeholder

- [ ] All
- [ ] HTML
- [x] React
- [ ] Angular
- [ ] Vue
- [ ] Web components
- [ ] Power Apps

### Reproduction case

**Note**: This Bug was first reported and fixed for the DBCustomSelect (see https://github.com/db-ux-design-system/core-web/issues/4725) but the same issue is also existing for the DBSelect.

Not sure why, but it looks like the DBSelect placeholder is first placed outside the component (possibly on first render), and is then moved to its final position (possibly on mount). This can be noticed if there are elements after the component. Such elements are then pushed down, and quickly after that moved up to their original position.

But here is an example how to reproduce this issue:

```jsx
import { DBButton, DBSelect } from '@db-ux/react-core-components';
import { useState } from 'react';

const App = () => {
  const [show, setShow] = useState(false);

  const toggle = () => {
    setShow(!show);
  };

  return (
    <>
      <DBButton type="button" onClick={toggle}>
        Toggle DBSelect
      </DBButton>
      <div style={{ minHeight: 32 }}>
        {!show && (
          <div style={{ height: 32 }}>
            DBSelect component placeholder (exactly tall as the
              DBSelect component)
          </div>
        )}
        {show && (
          <DBSelect
            options={[]}
            showLabel={false}
            label="Label"
            placeholder="Placeholder lorem ipsum placeholder"
          />
        )}
      </div>
      <div>
        This text will be pushed down a little bit when you toggle the
        DBSelect, and very soon after it will be pulled back to its
        original position
      </div>
    </>
  );
};

export default App;
```

### Expected Behaviour

No elements jumping happens before or after mounting the DBSelect component.

### Screenshots

_No response_

### Browser version

None

[https://stackblitz.com/edit/db-ux-design-system-examples-lc3zcszn?file=src%2FApp.tsx](https://stackblitz.com/edit/db-ux-design-system-examples-lc3zcszn?file=src%2FApp.tsx)

None

_No response_

_No response_

nmerget:
## Proposed changes

fixes #5139

## Types of changes

## Further comments

🔭🐙🐈 Test this branch here: https://db-ux-design-system.github.io/core-web/review/5139-fix-jumping-placeholder


__snapshots__/select/patternhub/select-overview-should-match-screenshot.png
mfranzke, line 1: the screenshot shows that the "g" is visually cut (vertically at the descender).

nmerget: fixed

mfranzke (approved):

PR #5424, opened 2025-11-07
feat: add breadcrumb component

sarahbrng:
## Proposed changes

add DBBreadcrumb component with documentation and examples

## Types of changes

## Further comments

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/1309-umsetzung-der-komponente-im-code-anhand-konzept-design-für-alle-frameworks>


packages/components/src/components/breadcrumb/breadcrumb.scss
```diff
@@ -0,0 +1,187 @@
...
+			background-color: colors.$db-adaptive-bg-basic-transparent-full-pressed;
+		}
+
+		@media screen and (prefers-reduced-motion: reduce) {
+			transition: none;
+		}
+	}
+
+	// Current page (last item with aria-current="page")
+	li[aria-current="page"] {
```
michaelmkraus: The list-item doesn't have the `aria-current` attribute. It is set on the `span` within the list-item. That is why the `text-decoration` is still shown when hovering the item.

packages/components/src/components/breadcrumb/breadcrumb.lite.tsx
```diff
@@ -0,0 +1,147 @@
...
+import { DBIcon } from '../icon';
+import type { DBBreadcrumbProps, DBBreadcrumbState } from './model';
+
+useMetadata({});
+
+useDefaultProps<DBBreadcrumbProps>({
+	size: 'small',
+	separator: 'chevron',
+	maxItems: undefined,
+	collapsedMenu: false,
```
michaelmkraus, line 19: The `collapsedMenu` prop is part of the props, but it is never used in the implementation.
We should either implement this behaviour as part of this story or drop the property for now to avoid exposing misleading props.

packages/components/src/components/breadcrumb/breadcrumb.scss
michaelmkraus: Fixed with [this](https://github.com/db-ux-design-system/core-web/pull/5424/commits/5b4142bba3e6a45e9fc623d0edf8df428a999440) commit.

packages/components/src/components/breadcrumb/breadcrumb.lite.tsx
```diff
@@ -0,0 +1,155 @@
...
+
+	return (
+		<nav
+			ref={_ref}
+			id={props.id}
+			class={cls('db-breadcrumb', props.className)}
+			data-size={props.size}
+			data-separator={props.separator}
+			aria-label="breadcrumb">
+			<ol class="db-breadcrumb-list" role="list">
```
michaelmkraus:
```suggestion
			<ol class="db-breadcrumb-list">
```
We can drop `role="list"` here. `<ol>` has the list role by default, so the role-attribute is redundant and doesn’t add any benefit.

packages/components/src/components/breadcrumb/breadcrumb.lite.tsx
```diff
@@ -0,0 +1,155 @@
...
+										</span>
+									)}
+								</li>
+
+								{/* Ellipsis button */}
+								<li>
+									<button
+										type="button"
+										class="db-breadcrumb-ellipsis"
+										aria-label={props.ellipsisAriaLabel}
```
michaelmkraus:
```suggestion
										aria-label={props.ellipsisAriaLabel}
										aria-expanded={state.isExpanded ? 'true' : 'false'}
```
`aria-expanded` shows that the controlled element is currently not expanded.

packages/components/src/components/breadcrumb/breadcrumb.lite.tsx
```diff
@@ -0,0 +1,155 @@
...
+			id={props.id}
+			class={cls('db-breadcrumb', props.className)}
+			data-size={props.size}
+			data-separator={props.separator}
+			aria-label="breadcrumb">
+			<ol class="db-breadcrumb-list" role="list">
+				{props.items && props.items.length > 0 ? (
+					<>
+						{props.maxItems &&
+						props.maxItems > 0 &&
```
michaelmkraus, line 46:
```suggestion
						props.maxItems > 2 &&
```
I think we need at least 2 items for the collapsed version.

packages/components/src/components/breadcrumb/breadcrumb.lite.tsx
michaelmkraus: If no property `ellipsisAriaLabel` is set, the button doesn't have an `aria-label`. We should use a default value.

packages/components/src/components/breadcrumb/breadcrumb.lite.tsx
```diff
@@ -0,0 +1,110 @@
...
+								{/* Collapsed view: first item + ellipsis + last items */}
+								<li key={0}>
+									{renderBreadcrumbContent(props.items[0], false)}
+								</li>
+
+								{/* Ellipsis button */}
+								<li key="ellipsis">
+									<button
+										type="button"
+										class="db-breadcrumb-ellipsis"
```
michaelmkraus:
```suggestion
										class="db-breadcrumb-ellipsis"
										title={props.ellipsisAriaLabel}
```
We could add a title (or our tooltip) to show a tooltip when hovering the button.

michaelmkraus: Added tooltip component.

packages/components/src/components/breadcrumb/breadcrumb.lite.tsx
michaelmkraus: Added aria-expanded attribute

PR #5540, opened 2025-11-28
feat: breadcrumb mvp

sarahbrng:
### Problem
- No breadcrumb navigation component existed in the DB UX Design System v3.

### Features
- **Sizes**: `small` (default) und `medium`
- **Separators**: `chevron` (default) und `slash`
- **Collapse/Expand**: Automatisches Kollabieren via `maxItems`; interaktiver Ellipsis-Button mit `ellipsisAriaLabel`
- **Icons**: Icon-Unterstützung pro Item aus DB UX Icon Library
- **Flexible API**: `items`-Array (empfohlen) oder manuelle `DBBreadcrumbItem` Child-Components mit `href`, `text`, `icon`, `ariaCurrent`, `disabled`

### Workarounds
- **Stencil/WC**: Visual-, ARIA- und Axe-Core-E2E-Tests deaktiviert — Custom Element Hosts unterbrechen `<ol>`→`<li>`-Struktur; UX bleibt vollständig accessible
- **Angular**: `CUSTOM_ELEMENTS_SCHEMA`; temporärer Import-Workaround in BreadcrumbItem; Post-Build Overwrites für Non-Null-Assertions bei `maxItems`/`items` (TypeScript strict mode)

### Changes
- **New `DBBreadcrumb` component**: supports `size` (small/medium), `separator` (chevron/slash), `maxItems` with collapsible ellipsis behavior, `ellipsisAriaLabel`, and data-driven `items` prop
- **New `DBBreadcrumbItem` component**: supports `text`, `icon`, `ariaCurrent`, `size`; uses lite imports throughout
- **Accessible disabled links**: `aria-disabled` + `tabIndex={-1}` added on disabled `<a>` elements — previously only `pointerEvents:none` was set, allowing keyboard users to still reach disabled items
- **Storybook controls**: restricted to supported icon props only in `DBBreadcrumbItem`
- **E2E test coverage**: visual snapshots (5 browsers/modes), ARIA snapshots, axe-core, and a11y-checker specs added
- **Framework showcases**: Angular, React, Vue showcases and patternhub docs/routes wired up; `internalListId` state added for robust list ID management

### Affected Areas
- `packages/components/src/components/breadcrumb*` — new component source, models, styles, specs
- `showcases/` (angular, react, vue, e2e, patternhub) — showcase integration and navigation entries

---

**Breaking Changes**: none — new components only  
**Risk**: low — isolated new feature, no modifications to existing components  
**Validated**: visual snapshots regenerated across chromium, firefox, webkit, mobile-chrome, mobile-safari; ARIA and a11y tests passing

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/1493-erstellung-einer-mvp-breadcrumb-komponente>


packages/components/src/components/breadcrumb-item/breadcrumb-item.lite.tsx
```diff
@@ -0,0 +1,39 @@
...
+		<li ref={_ref} id={props.id} class={cls(props.className)}>
+			<Show
+				when={props.href && !props.disabled}
+				else={
+					<span aria-current={props.ariaCurrent}>
+						<Show when={props.icon}>
+							<DBIcon icon={props.icon} />
+						</Show>
+						{props.text ? props.text : props.children}
+					</span>
```
mfranzke:
```suggestion
					<span aria-current={props.ariaCurrent} data-icon={props.icon}>
						{props.text ? props.text : props.children}
					</span>
```

Please do not use `DBIcon` internally. This is more of a leftover and we might remove it the sooner the later.

packages/components/src/components/breadcrumb-item/breadcrumb-item.lite.tsx
```diff
@@ -0,0 +1,39 @@
...
+				}>
+				<a href={props.href} aria-current={props.ariaCurrent}>
+					<Show when={props.icon}>
+						<DBIcon icon={props.icon} />
+					</Show>
+					{props.text ? props.text : props.children}
+				</a>
```
mfranzke:
```suggestion
				<a href={props.href} aria-current={props.ariaCurrent} data-icon={props.icon}>
					{props.text ? props.text : props.children}
				</a>
```

Please do not use `DBIcon` internally. This is more of a leftover and we might remove it the sooner the later.

packages/components/src/components/breadcrumb-item/breadcrumb-item.lite.tsx
```diff
@@ -0,0 +1,39 @@
+import {
+	Show,
+	useDefaultProps,
+	useMetadata,
+	useRef
+} from '@builder.io/mitosis';
+import { cls } from '../../utils';
+import { DBIcon } from '../icon';
```
mfranzke:
```suggestion
```

Please do not use `DBIcon` internally. This is more of a leftover and we might remove it the sooner the later.

packages/components/src/components/breadcrumb/breadcrumb.lite.tsx
```diff
@@ -0,0 +1,282 @@
...
+																1
+													)}>
+													<Show when={item.icon}>
+														<DBIcon
+															weight={state.iconWeight()}
+															icon={item.icon}
+														/>
+													</Show>
+													{item.text}
+												</span>
```
mfranzke:
```suggestion
												<span
													aria-current={state.ariaCurrent(
														item,
														index ===
															state.normalizedItems()
																.length -
																1
														data-icon={item.icon}
													)}>
													{item.text}
												</span>
```

Please do not use `DBIcon` internally. This is more of a leftover and we might remove it the sooner the later.

packages/components/src/components/breadcrumb/breadcrumb.lite.tsx
```diff
@@ -0,0 +1,282 @@
...
+															1
+												)}>
+												<Show when={item.icon}>
+													<DBIcon
+														weight={state.iconWeight()}
+														icon={item.icon}
+													/>
+												</Show>
+												{item.text}
+											</a>
```
mfranzke:
```suggestion
											<a
												href={item.href}
												aria-current={state.ariaCurrent(
													item,
													index ===
														state.normalizedItems()
															.length -
															1
													data-icon={item.icon}
												)}>
												{item.text}
											</a>
```

Please do not use `DBIcon` internally. This is more of a leftover and we might remove it the sooner the later.

packages/components/src/components/breadcrumb-item/breadcrumb-item.lite.tsx
```diff
@@ -0,0 +1,39 @@
...
+useMetadata({});
+
+useDefaultProps<DBBreadcrumbItemProps>({});
+
+export default function DBBreadcrumbItem(props: DBBreadcrumbItemProps) {
+	const _ref = useRef<HTMLLIElement | any>(null);
+
+	return (
+		<li ref={_ref} id={props.id} class={cls(props.className)}>
+			<Show
```
nmerget, line 23: Why did you use this `<Show>` at all?

We should use `props.children` here only. The reason is that most frameworks have custom `<Link>` component which will render as an `<a>` tag but they might do some additional magic based on the framework. Furthermore an `<a>` tag may have a lot of additional properties you can add we already have a `LinkProps` type inside `shared/model.ts`, we use this for our `<DBLink>` to pass all those properties. But it would be better to avoid passing all properties if it is possible.

Maybe you can move `props.disabled` to the `<li>` and you could use [`pointer-events`](https://developer.mozilla.org/de/docs/Web/CSS/Reference/Properties/pointer-events) or JS inside the breadcumb to remove the interactivity of the <a> of the last breadcrumb-item

packages/components/src/components/breadcrumb-item/breadcrumb-item.scss
```diff
@@ -0,0 +1,4 @@
+@charset "utf-8";
+
+// Breadcrumb item styling is handled by the parent breadcrumb component
+// This file exists to maintain the component structure
```
nmerget: Why couldn't we just move [this](https://github.com/db-ux-design-system/core-web/pull/5540/changes#diff-7c91b36b94b28f88d2ef95c6152642a775ae80b3350ed992964563c615867805R20-R72) here?

Right now the showcase for breadcrumb-item wouldn't be styled

packages/components/src/components/breadcrumb/breadcrumb.scss
```diff
@@ -0,0 +1,220 @@
...
+		font-family: var(
+			--db-type-body-font-family,
+			var(--db-font-family-sans)
+		);
+		font-size: var(--db-type-body-font-size-sm, 0.875rem); // 14px (small)
+		font-weight: 400;
+		line-height: var(--db-type-body-line-height-sm, 1.25rem); // 20px
+		text-decoration: none;
+		white-space: nowrap;
+		transition: background-color 0.2s ease;
```
nmerget: We set this globally [with](https://github.com/db-ux-design-system/core-web/blob/main/packages/foundations/scss/animation/_transitions.scss#L3-L8) - it should be applied already

packages/components/src/components/breadcrumb/breadcrumb.scss
```diff
@@ -0,0 +1,220 @@
...
+	li a,
+	li > span:not(.db-icon) {
+		display: flex;
+		align-items: center;
+		gap: variables.$db-spacing-fixed-2xs; // small: spacing between icon and text per Figma
+		padding: variables.$db-spacing-fixed-3xs variables.$db-spacing-fixed-2xs; // 2px 4px (small per Figma)
+		border-radius: variables.$db-border-radius-xs; // 4px
+		background-color: colors.$db-adaptive-bg-basic-transparent-full-default;
+		color: colors.$db-adaptive-on-bg-basic-emphasis-100-default;
+		font-family: var(
```
nmerget: We set this globally - it should be applied already

packages/components/src/components/breadcrumb/breadcrumb.scss
```diff
@@ -0,0 +1,220 @@
...
+			--db-type-body-font-family,
+			var(--db-font-family-sans)
+		);
+		font-size: var(--db-type-body-font-size-sm, 0.875rem); // 14px (small)
+		font-weight: 400;
```
nmerget: We set this globally - it should be applied already

packages/components/src/components/breadcrumb/breadcrumb.scss
```diff
@@ -0,0 +1,220 @@
...
+		gap: variables.$db-spacing-fixed-2xs; // small: spacing between icon and text per Figma
+		padding: variables.$db-spacing-fixed-3xs variables.$db-spacing-fixed-2xs; // 2px 4px (small per Figma)
+		border-radius: variables.$db-border-radius-xs; // 4px
+		background-color: colors.$db-adaptive-bg-basic-transparent-full-default;
+		color: colors.$db-adaptive-on-bg-basic-emphasis-100-default;
+		font-family: var(
+			--db-type-body-font-family,
+			var(--db-font-family-sans)
+		);
+		font-size: var(--db-type-body-font-size-sm, 0.875rem); // 14px (small)
```
nmerget: Please use `@extend %db-overwrite-font-size-sm;` with this @use: `@use "@db-ux/core-foundations/build/styles/fonts";` - This will apply the correct font-size and line-height and it makes sure to set the correct sizes for `[data-icon]`

packages/components/src/components/breadcrumb/breadcrumb.scss
```diff
@@ -0,0 +1,220 @@
...
+		font-weight: 400;
+		line-height: var(--db-type-body-line-height-sm, 1.25rem); // 20px
```
nmerget: https://github.com/db-ux-design-system/core-web/pull/5540/changes#r2715667746

packages/components/src/components/breadcrumb/breadcrumb.scss
```diff
@@ -0,0 +1,220 @@
...
+	// Link and text styling (small)
+	li a,
+	li > span:not(.db-icon) {
+		display: flex;
+		align-items: center;
+		gap: variables.$db-spacing-fixed-2xs; // small: spacing between icon and text per Figma
+		padding: variables.$db-spacing-fixed-3xs variables.$db-spacing-fixed-2xs; // 2px 4px (small per Figma)
+		border-radius: variables.$db-border-radius-xs; // 4px
+		background-color: colors.$db-adaptive-bg-basic-transparent-full-default;
+		color: colors.$db-adaptive-on-bg-basic-emphasis-100-default;
```
nmerget:
```suggestion
```

That's the default, we don't need to set it when we don't overwrite it anywhere

packages/components/src/components/breadcrumb/breadcrumb.scss
```diff
@@ -0,0 +1,220 @@
...
+		}
+
+		&:hover,
+		&:focus {
+			background-color: colors.$db-adaptive-bg-basic-transparent-semi-default;
+		}
+
+		&:active {
+			background-color: colors.$db-adaptive-bg-basic-transparent-full-pressed;
+		}
```
nmerget: We have already defined this:

```

@use "@db-ux/core-foundations/build/styles/helpers";

@include helpers.hover {
			background-color: colors.$db-adaptive-bg-basic-transparent-full-hovered;
		}

		@include helpers.active {
			background-color: colors.$db-adaptive-bg-basic-transparent-full-pressed;
		}```

This will handle the disable state as well

File: packages/components/src/components/breadcrumb/breadcrumb.scss [hunk 15]
```diff
@@ -0,0 +1,220 @@
...
+
+			&:hover,
+			&:focus {
+				background-color: colors.$db-adaptive-bg-basic-transparent-full-default;
+				text-decoration: none;
+			}
+		}
+
+		&:hover,
+		&:focus {
```
nmerget: `&:focus` <- is this a requirement from design?

I remember that we removed it from many components to avoid having the focus and hover effect

packages/components/src/components/breadcrumb/breadcrumb.scss
```diff
@@ -0,0 +1,220 @@
...
+			background-color: colors.$db-adaptive-bg-basic-transparent-semi-default;
+		}
+
+		&:active {
+			background-color: colors.$db-adaptive-bg-basic-transparent-full-pressed;
+		}
+
+		@media screen and (prefers-reduced-motion: reduce) {
+			transition: none;
+		}
```
nmerget: I guess you can remove this when appling https://github.com/db-ux-design-system/core-web/pull/5540/changes#r2715663026

packages/components/src/components/breadcrumb/breadcrumb.lite.tsx
```diff
@@ -0,0 +1,282 @@
...
+	size: 'small',
+	separator: 'chevron',
+	ellipsisAriaLabel: 'Expand to show all breadcrumb items'
+});
+
+export default function DBBreadcrumb(props: DBBreadcrumbProps) {
+	const _ref = useRef<HTMLElement | any>(null);
+
+	const state = useStore<DBBreadcrumbState>({
+		uniqueId: uuid(),
```
nmerget: We have a `export type GlobalState = {
	_id?: string;
};` inside `shared/model.ts` and use `_id` for this case

packages/components/src/components/breadcrumb/breadcrumb.lite.tsx
```diff
@@ -0,0 +1,282 @@
...
+			if (typeof value === 'string') {
+				try {
+					const parsed = JSON.parse(value);
+					return Array.isArray(parsed) ? parsed : [];
+				} catch (error) {
+					return [];
+				}
+			}
+
+			return [];
```
nmerget: This looks nice, we already do something similiar inside `accordion` and `tabs` maybe we can move it to the `utils.ts` with something like `export const parseItems<T> = (items?: T[]|string):T[]=>{...}`

packages/components/src/components/breadcrumb/breadcrumb.lite.tsx
```diff
@@ -0,0 +1,282 @@
...
+												<DBIcon
+													weight={state.iconWeight()}
+													icon={
+														state.normalizedItems()[0]
+															.icon
+													}
+												/>
+											</Show>
+											{state.normalizedItems()[0].text}
+										</a>
```
mfranzke:
```suggestion
										<a
											href={
												state.normalizedItems()[0].href
											}
											aria-current={state.ariaCurrent(
												state.normalizedItems()[0],
												state.normalizedItems()
													.length === 1
											)}
											data-icon={
												state.normalizedItems()[0]
													.icon
											}>
											{state.normalizedItems()[0].text}
										</a>
```

Please do not use `DBIcon` internally. This is more of a leftover and we might remove it the sooner the later.

packages/components/src/components/breadcrumb/breadcrumb.lite.tsx
```diff
@@ -0,0 +1,282 @@
...
+															state.normalizedItems()[0]
+																.icon
+														}
+													/>
+												</Show>
+												{
+													state.normalizedItems()[0]
+														.text
+												}
+											</span>
```
mfranzke:
```suggestion
											<span
												aria-current={state.ariaCurrent(
													state.normalizedItems()[0],
													state.normalizedItems()
														.length === 1
												)}
												data-icon={
													state.normalizedItems()[0]
														.icon
												}>
												{
													state.normalizedItems()[0]
														.text
												}
											</span>
```

Please do not use `DBIcon` internally. This is more of a leftover and we might remove it the sooner the later.

packages/components/src/components/breadcrumb/breadcrumb.lite.tsx
```diff
@@ -0,0 +1,282 @@
...
+				} catch (error) {
+					return [];
+				}
+			}
+
+			return [];
+		},
+		normalizedItems(): DBBreadcrumbItems[] {
+			return state.parseItems();
+		},
```
nmerget, line 38: This is basiclly the same as the [other function](https://github.com/db-ux-design-system/core-web/pull/5540/changes#diff-ba2653f616d0724264ecd2ee69bdaa99b7a00b8baab05f0e63e2689255fd1c0eR37) isn't it?

packages/components/src/components/breadcrumb/breadcrumb.lite.tsx
```diff
@@ -0,0 +1,282 @@
...
+			id={props.id}
+			class={cls('db-breadcrumb', props.className)}
+			data-size={props.size}
+			data-separator={props.separator}
+			aria-label={
+				props.ariaLabel ??
+				(props.id
+					? `Breadcrumb (${props.id})`
+					: `Breadcrumb (${state.uniqueId})`)
+			}>
```
nmerget:
```suggestion
>
```

We don't provide `props.ariaLabel` anymore. Every framework should be enabled to pass `aria-label` to the `_ref` which is the `<nav>` at them moment.

packages/components/src/components/breadcrumb/breadcrumb.lite.tsx
```diff
@@ -0,0 +1,282 @@
...
+			<ol
+				class="db-breadcrumb-list"
+				id={props.id ? `${props.id}-list` : 'db-breadcrumb-list'}>
+				<Show
+					when={state.normalizedItems().length > 0}
+					else={<>{props.children}</>}>
```
nmerget:
```suggestion
					else={props.children}>
```

This should be fine as well

packages/components/src/components/breadcrumb/breadcrumb.lite.tsx
```diff
@@ -0,0 +1,282 @@
...
+					{/* jscpd:ignore-start */}
+					<Show
+						when={state.isCollapsed()}
+						else={
+							<For each={state.normalizedItems()}>
+								{(item, index) => (
+									<li>
```
nmerget: Why did you rebuild the entire `<DBBreadcrumItem>` here? We should just import it and use the correct props.

packages/components/src/components/breadcrumb/breadcrumb.lite.tsx
```diff
@@ -0,0 +1,282 @@
...
+											</Show>
+											{state.normalizedItems()[0].text}
+										</a>
+									</Show>
+								</li>
+							</Show>
+							<li key="ellipsis">
+								<button
+									type="button"
+									class="db-breadcrumb-ellipsis"
```
nmerget, line 126: This should be the `<DBButton size="small" variant="ghost">` or it should use the same styling like the `db-breadcrum-item`. Maybe we can share the paddings/border-radius etc. with the `db-button` to make sure it it is consitent?

packages/components/src/components/breadcrumb/breadcrumb.lite.tsx
```diff
@@ -0,0 +1,282 @@
...
+			data-separator={props.separator}
+			aria-label={
+				props.ariaLabel ??
+				(props.id
+					? `Breadcrumb (${props.id})`
+					: `Breadcrumb (${state.uniqueId})`)
+			}>
+			<ol
+				class="db-breadcrumb-list"
+				id={props.id ? `${props.id}-list` : 'db-breadcrumb-list'}>
```
nmerget: I saw you use this for `aria-controls`, if we want to use it we should get this from the state like this `state.listId`, where we use `uuid`

packages/components/src/components/breadcrumb/breadcrumb.lite.tsx
```diff
@@ -0,0 +1,282 @@
...
+															.icon
+													}
+												/>
+											</Show>
+											{state.normalizedItems()[0].text}
+										</a>
+									</Show>
+								</li>
+							</Show>
+							<li key="ellipsis">
```
nmerget: Overall this solution might not be perfect. 
This feature would be only available for users passing in the `items` which might be the minority.

We need to move this outside of the `items`. Maybe it is possible to fiddle around with `flex` and `order` or `grid` with `grid-area`. In the worst case we need to use JS to mount it to the DOM. 

But at the moment we would hide this feature for many users

packages/components/src/components/breadcrumb/breadcrumb.lite.tsx
```diff
@@ -0,0 +1,282 @@
...
+									aria-expanded={getBooleanAsString(
+										state.isExpanded
+									)}
+									aria-controls={
+										props.id
+											? `${props.id}-list`
+											: 'db-breadcrumb-list'
+									}
+									onClick={() => state.toggleExpanded()}>
+									…
```
nmerget, line 140: We should add a tooltip here with the ariaLabel, and maybe we can add the `...` as `:before` element?

packages/components/src/components/breadcrumb/breadcrumb.scss
```diff
@@ -0,0 +1,220 @@
...
+		&:focus-visible {
+			outline: 2px solid
+				colors.$db-adaptive-on-bg-basic-emphasis-100-default;
+			outline-offset: 2px;
+		}
+
+		@media screen and (prefers-reduced-motion: reduce) {
+			transition: none;
+		}
+	}
```
nmerget, line 34: This is basiclly the `db-breadcrumb-item` with some minor tweaks. No need to add all of them again

packages/components/src/components/breadcrumb/breadcrumb.scss
```diff
@@ -0,0 +1,220 @@
...
+
+	// Accessibility: ensure links show a clear focus indicator
+	a:focus-visible {
+		outline: 2px solid colors.$db-adaptive-on-bg-basic-emphasis-100-default;
+		outline-offset: 2px;
+	}
```
nmerget: We already set a focus for every interactive element [globally](https://github.com/db-ux-design-system/core-web/blob/main/packages/foundations/scss/defaults/default-required.scss#L126-L138)

packages/components/src/components/breadcrumb/breadcrumb.scss
```diff
@@ -0,0 +1,220 @@
...
+	// Default chevron: apply when the breadcrumb does NOT request a slash
+	&:not([data-separator="slash"])
+		.db-breadcrumb-list
+		> li:not(:first-child)::before {
+		@extend %icon;
+
+		--db-icon: "chevron_right";
+		--db-icon-color: #{colors.$db-adaptive-on-bg-basic-emphasis-100-default};
+		--db-icon-font-weight: 16; // Font weight to load 16px icon font variant
+		--db-icon-font-size: 1rem; // Actual size: 16px (small)
```
nmerget, line 50: If you use `@extend %db-overwrite-font-size-sm;` like I said [here](https://github.com/db-ux-design-system/core-web/pull/5540/changes#r2715667746)  under the `<li>` instead of the `<a>` you wouldnt need this

packages/components/src/components/breadcrumb/breadcrumb.scss
```diff
@@ -0,0 +1,220 @@
...
+
+		display: flex;
+		align-items: center;
+		justify-content: center;
```
nmerget, line 54: Can you move this directlly to the `<li>` every element should be centered vertically shouldn't it?

packages/components/src/components/breadcrumb/breadcrumb.lite.tsx
```diff
@@ -0,0 +1,282 @@
...
+																1
+													)}>
+													<Show when={item.icon}>
+														<DBIcon
+															weight={state.iconWeight()}
+															icon={item.icon}
+														/>
+													</Show>
+													{item.text}
+												</span>
```
mfranzke:
```suggestion
												<span
													aria-current={state.ariaCurrent(
														item,
														index ===
															state.collapsedTailItems()
																.length -
																1
													)}
													data-icon={item.icon}>
													{item.text}
												</span>
```

Please do not use `DBIcon` internally. This is more of a leftover and we might remove it the sooner the later.

packages/components/src/components/breadcrumb/breadcrumb.scss
```diff
@@ -0,0 +1,220 @@
...
+	}
+
+	// Slash separator override when component has data-separator="slash"
+	&[data-separator="slash"] .db-breadcrumb-list li:not(:first-child)::before {
+		content: "/"; // ASCII slash (0x2F)
+		font-weight: 400;
+		color: colors.$db-adaptive-on-bg-basic-emphasis-100-default;
```
nmerget, line 66:
```suggestion
```

Those are the defaults no for them

packages/components/src/components/breadcrumb/breadcrumb.lite.tsx
```diff
@@ -0,0 +1,282 @@
...
+															1
+												)}>
+												<Show when={item.icon}>
+													<DBIcon
+														weight={state.iconWeight()}
+														icon={item.icon}
+													/>
+												</Show>
+												{item.text}
+											</a>
```
mfranzke:
```suggestion
											<a
												href={item.href}
												aria-current={state.ariaCurrent(
													item,
													index ===
														state.collapsedTailItems()
															.length -
															1
												)}
												data-icon={item.icon}>
												{item.text}
											</a>
```

Please do not use `DBIcon` internally. This is more of a leftover and we might remove it the sooner the later.

packages/components/src/components/breadcrumb/breadcrumb.scss
```diff
@@ -0,0 +1,220 @@
...
+
+		/* reduce the inline size so the slash doesn't create extra empty box */
+		inline-size: auto;
+		block-size: auto;
+		font-size: var(--db-sizing-functional-xs, 0.875rem); // 14px (0.875rem)
```
nmerget:
```suggestion
```

See this: https://github.com/db-ux-design-system/core-web/pull/5540/changes#r2715826033

packages/components/src/components/breadcrumb/breadcrumb.scss
```diff
@@ -0,0 +1,220 @@
...
+
+	// Slash separator override when component has data-separator="slash"
+	&[data-separator="slash"] .db-breadcrumb-list li:not(:first-child)::before {
+		content: "/"; // ASCII slash (0x2F)
+		font-weight: 400;
+		color: colors.$db-adaptive-on-bg-basic-emphasis-100-default;
+
+		/* reduce the inline size so the slash doesn't create extra empty box */
+		inline-size: auto;
+		block-size: auto;
```
nmerget: Can you ask design how this should behave? It makes sense to me to have the same size as the chevron.

If this is the case you should use this:


```suggestion
		/* reduce the inline size so the slash doesn't create extra empty box */
		inline-size: var(--db-icon-font-size);
		aspect-ratio: 1;
```

packages/components/src/components/breadcrumb/breadcrumb.scss
```diff
@@ -0,0 +1,220 @@
...
+		font-size: var(--db-sizing-functional-xs, 0.875rem); // 14px (0.875rem)
+	}
+
+	// Medium size variant
+	&[data-size="medium"] {
+		.db-breadcrumb-list {
+			gap: variables.$db-spacing-fixed-2xs; // medium: list item gap (per Figma)
+
+			/* ensure list items are vertically centered in medium size */
+			align-items: center;
```
nmerget, line 94: This should be the case for small as well, shouldn't it?

packages/components/src/components/breadcrumb/breadcrumb.scss
```diff
@@ -0,0 +1,220 @@
...
+		}
+
+		// Ensure medium size chevron styling
+		&:not([data-separator="slash"])
+			.db-breadcrumb-list
+			> li:not(:first-child)::before {
+			--db-icon-font-weight: 20; // Font weight to load 20px icon font variant
+			--db-icon-font-size: 1.25rem; // Actual size: 20px (medium)
+			--db-icon-color: #{colors.$db-adaptive-on-bg-basic-emphasis-100-default};
+		}
```
nmerget, line 108: Can be removed by doing [this](https://github.com/db-ux-design-system/core-web/pull/5540/changes#r2715826033)

packages/components/src/components/breadcrumb/breadcrumb.scss
```diff
@@ -0,0 +1,220 @@
...
+
+		// Ellipsis button styling for medium size
+		.db-breadcrumb-ellipsis {
+			padding: variables.$db-spacing-fixed-2xs
+				variables.$db-spacing-fixed-xs; // match medium padding
+
+			font-size: var(--db-type-body-font-size-md, 1rem); // 16px (medium)
+			line-height: var(--db-type-body-line-height-md, 1.5rem); // 24px
+		}
```
nmerget, line 116: Can be removed with [this](https://github.com/db-ux-design-system/core-web/pull/5540/changes#r2715803409)

packages/components/src/components/breadcrumb/breadcrumb.scss
```diff
@@ -0,0 +1,220 @@
...
+				// Keep the medium padding but ensure comfortable tap/click area
+				padding: variables.$db-spacing-fixed-2xs
+					variables.$db-spacing-fixed-xs; // 4px 8px (medium)
+
+				font-size: var(
+					--db-type-body-font-size-md,
+					1rem
+				); // 16px (medium)
+
+				line-height: var(--db-type-body-line-height-md, 1.5rem); // 24px
```
nmerget:
```suggestion
@extend %db-overwrite-font-size-md;
```

https://github.com/db-ux-design-system/core-web/pull/5540/changes#r2715667746

packages/components/src/components/breadcrumb/breadcrumb.scss
```diff
@@ -0,0 +1,220 @@
...
+		}
+	}
+
+	// Remove hover effect for icons within breadcrumb items
+	li a:hover .db-icon,
+	li a:focus .db-icon,
+	li a:active .db-icon {
+		color: colors.$db-adaptive-on-bg-basic-emphasis-100-default;
+		background: none;
+	}
```
nmerget:
```suggestion
```

No need for this we should use `data-icon`

packages/components/src/components/breadcrumb/breadcrumb.scss
```diff
@@ -0,0 +1,220 @@
...
+
+	// Ensure icon size matches weight attribute for small size (20px)
+	.db-icon[data-icon-weight="20"]::before {
+		--db-icon-font-size: 1.25rem; // 20px
+	}
+
+	// Ensure icon size matches weight attribute for medium size (24px)
+	&[data-size="medium"] .db-icon[data-icon-weight="24"]::before {
+		--db-icon-font-size: 1.5rem; // 24px
+	}
```
nmerget: Can be removed by doing [this](https://github.com/db-ux-design-system/core-web/pull/5540/changes#r2715826033)

packages/components/src/components/breadcrumb/breadcrumb.lite.tsx
```diff
@@ -0,0 +1,282 @@
...
+									</Show>
+								</li>
+							</Show>
+							<li key="ellipsis">
+								<button
+									type="button"
+									class="db-breadcrumb-ellipsis"
+									aria-label={
+										props.ellipsisAriaLabel ??
+										'Expand to show all breadcrumb items'
```
nmerget, line 129: We define a defaultProp already do we really need this? Otherwise this should be a constant to share the same string

packages/components/src/components/breadcrumb/model.ts
```diff
@@ -0,0 +1,110 @@
+import { GlobalProps, GlobalState } from '../../shared/model';
+import type { DBBreadcrumbItems } from '../breadcrumb-item/model';
+export type { DBBreadcrumbItems } from '../breadcrumb-item/model';
```
mfranzke:
```suggestion
export type { DBBreadcrumbItems } from '../breadcrumb-item/model';
```

why do you export this one?

packages/components/src/index.ts
```diff
@@ -6,6 +6,10 @@ export * from './components/badge';
 export * from './components/badge/model';
 export * from './components/brand';
 export * from './components/brand/model';
+export * from './components/breadcrumb';
+export * from './components/breadcrumb-item';
+export * from './components/breadcrumb-item/model';
+export * from './components/breadcrumb/model';
```
mfranzke:
```suggestion
export * from './components/breadcrumb';
export * from './components/breadcrumb/model';
export * from './components/breadcrumb-item';
export * from './components/breadcrumb-item/model';
```

packages/components/scripts/post-build/components.ts
```diff
@@ -360,6 +360,29 @@ export const getComponents = (): Component[] => [
...
+				{
+					from: /items\(\)\.slice\(items\(\)!\.length - maxItems\(\)! \+ 1\)/g,
+					to: 'items()!.slice(items()!.length - maxItems()! + 1)'
+				},
+				{
+					from: /@if\(maxItems\(\) && items\(\) && items\(\)\.length > maxItems\(\)\)/g,
+					to: '@if(maxItems() && items() && items()!.length > maxItems()!)'
+				}
+			]
+		}
```
nmerget: Is this a leftover? Couldn't find those in the code

packages/components/src/components/breadcrumb/breadcrumb.lite.tsx
```diff
@@ -0,0 +1,282 @@
...
+					: `Breadcrumb (${state.uniqueId})`)
+			}>
+			<ol
+				class="db-breadcrumb-list"
+				id={props.id ? `${props.id}-list` : 'db-breadcrumb-list'}>
+				<Show
+					when={state.normalizedItems().length > 0}
+					else={<>{props.children}</>}>
+					{/* jscpd:ignore-start */}
+					<Show
```
nmerget, line 83: We shouldn't handle this with JS and `<Show>` instead we could use a `data-*` property on `<ol>` with `li:nth(x)` and `display:none` to hide the button and list items.

This would reduce the amount of JS executing, and we could add some transition and `height: auto` effect in the future if we want to (similar to accordion)

packages/components/scripts/post-build/stencil.ts
```diff
@@ -79,6 +79,13 @@ export default (tmp?: boolean) => {
 			processor: (input: string) => changeFile(upperComponentName, input)
 		});
 
+		// Fix DBIcon imports for Stencil (uses default export)
+		replaceInFileSync({
+			files: [file],
+			from: /import \{ DBIcon \} from ['"]\.\.\/icon['"]/g,
+			to: 'import DBIcon from "../icon"'
+		});
```
nmerget: We don't need this when we repalce the `DBIcon` with `data-icon`

packages/components/src/components/breadcrumb/breadcrumb.lite.tsx
```diff
@@ -0,0 +1,282 @@
+import {
+	For,
+	Show,
+	useDefaultProps,
+	useMetadata,
+	useRef,
+	useStore
+} from '@builder.io/mitosis';
+import { cls, getBooleanAsString, uuid } from '../../utils';
+import { DBIcon } from '../icon';
```
mfranzke:
```suggestion
```

Please do not use `DBIcon` internally. This is more of a leftover and we might remove it the sooner the later.

packages/components/src/components/breadcrumb/breadcrumb.scss
```diff
@@ -0,0 +1,220 @@
...
+@use "@db-ux/core-foundations/build/styles/icons";
+
+.db-breadcrumb {
+	// Default: Small size
+	.db-breadcrumb-list {
+		display: flex;
+		flex-wrap: wrap;
+		align-items: flex-start;
+		gap: variables.$db-spacing-fixed-3xs; // 2px gap for small (per Figma)
+		list-style: none;
```
mfranzke:
```suggestion
		list-style-type: "";
```

[`list-style: none;` would remove the "semantics" of the list element in VoiceOver.](https://www.matuzo.at/blog/2023/removing-list-styles-without-affecting-semantics)

packages/components/src/components/breadcrumb/index.html
```diff
@@ -0,0 +1,30 @@
+<!doctype html>
+<html lang="en">
+	<head>
+		<meta charset="UTF-8" />
+		<title>DBBreadcrumb</title>
+		<link rel="stylesheet" href="/styles/relative.css" />
+	</head>
+	<body style="padding: var(--db-spacing-fixed-md)">
+		<nav class="db-breadcrumb" aria-label="breadcrumb">
+			<ol class="db-breadcrumb-list" role="list">
```
mfranzke:
```suggestion
			<ol class="db-breadcrumb-list">
```

`list` is the existing default role for this element, so we don't need to set it.

packages/components/src/components/breadcrumb/index.html
```diff
@@ -0,0 +1,30 @@
...
+				<li><a href="#">Path</a></li>
+				<li aria-current="page"><span>Current Page</span></li>
+			</ol>
+		</nav>
+
+		<br /><br />
+
+		<!-- Shorter example -->
+		<nav class="db-breadcrumb" aria-label="breadcrumb">
+			<ol class="db-breadcrumb-list" role="list">
```
mfranzke:
```suggestion
			<ol class="db-breadcrumb-list">
```

`list` is the existing default role for this element, so we don't need to set it.

packages/components/src/styles/index.scss
```diff
@@ -5,6 +5,7 @@
 @forward "../components/card/card";
 @forward "../components/input/input";
 @forward "../components/brand/brand";
+@forward "../components/breadcrumb/breadcrumb";
```
mfranzke, line 8:
```suggestion
@forward "../components/breadcrumb/breadcrumb";
@forward "../components/breadcrumb/breadcrumb-item";
```

isn't it that we even also need to include the items styling?

packages/components/src/components/breadcrumb/breadcrumb.scss
```diff
@@ -0,0 +1,220 @@
...
+		line-height: var(--db-type-body-line-height-sm, 1.25rem); // 20px
+		text-decoration: none;
+		white-space: nowrap;
+		transition: background-color 0.2s ease;
+
+		// Current page (last item with aria-current="page")
+		&[aria-current="page"] {
+			font-weight: 700; // Bold for current item
+			cursor: default;
+			pointer-events: none;
```
mfranzke:
```suggestion
```

To my understanding, those are static elements, so these declarations aren't necessary.

showcases/angular-showcase/src/app/components/breadcrumb-item/breadcrumb-item.component.ts
```diff
@@ -0,0 +1,18 @@
...
+import { DBBreadcrumbItem } from '../../../../../../output/angular/src';
+import defaultComponentVariants from '../../../../../shared/breadcrumb-item.json';
+import { environment } from '../../../environments/environment';
+import { DefaultComponent } from '../default.component';
+
+@Component({
+	selector: 'app-breadcrumb-item',
+	templateUrl: './breadcrumb-item.component.html',
+	imports: environment.webComponents
+		? [DefaultComponent, DBBreadcrumbItem] // TODO: Remove DBBreadcrumbItem after stencil component works
```
mfranzke: is this a TODO leftover that could already get removed ?

packages/components/src/components/breadcrumb/docs/Migration.md
```diff
@@ -0,0 +1,15 @@
...
+
+### Breaking Changes
+
+- None (new component)
+
+### New Features
+
+- Added `DBBreadcrumb` component for navigation breadcrumbs
+- Supports semantic HTML structure with proper ARIA labels
+- Responsive design with flexbox layout
```
mfranzke:
```suggestion
## General

> **Note**
> For a general installation or migration process check out this [documentation](https://www.npmjs.com/package/@db-ux/core-components).

## DB UI Core ➡ DB UX Design System Core components

New Component 🥳
```

Let's align this with the Migration guide from other (new) components.

packages/components/src/components/breadcrumb-item/breadcrumb-item.scss
sarahbrng: Fixed: Move breadcrumb-item styling to component SCSS file for encapsulation

packages/components/src/components/breadcrumb-item/breadcrumb-item.lite.tsx
sarahbrng: Fixed: Removed DBIcon internal usage in breadcrumb-item

packages/components/src/components/breadcrumb-item/breadcrumb-item.lite.tsx
sarahbrng: Fixed: Removed DBIcon internal usage in breadcrumb-item

packages/components/src/components/breadcrumb-item/breadcrumb-item.lite.tsx
sarahbrng: Fixed: Removed DBIcon internal usage in breadcrumb-item

packages/components/src/components/breadcrumb-item/breadcrumb-item.lite.tsx
sarahbrng: Fixed: Removed redundant <Show> conditional rendering to allow framework-specific <Link> components. Moved disabled state to <li> with pointer-events, extended LinkProps for consistency.

packages/components/scripts/post-build/components.ts
sarahbrng: Fixed: Removed obsolete breadcrumb Angular post-build overwrites that no longer match generated code, keeping component registration without dead replacement rules.

packages/components/src/components/breadcrumb/breadcrumb.lite.tsx
sarahbrng: Fixed: Extracted generic parseItems to shared utils and removed redundant breadcrumb parsing wrapper to reduce duplication and align with existing component patterns.

packages/components/src/components/breadcrumb/breadcrumb.lite.tsx
sarahbrng: Fixed: Extracted shared parseItems utility to [index.ts](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html), removed duplicate parsing in breadcrumb, aligned state types, restored missing icon import, and added utility tests.

packages/components/src/components/breadcrumb/breadcrumb.lite.tsx
sarahbrng: Fixed: aria-controls now references a stable per-instance list id from state, preventing duplicate fallback ids.

packages/components/src/components/breadcrumb/breadcrumb.lite.tsx
sarahbrng: Fixed: added title on ellipsis button using the same accessible label fallback.

packages/components/src/components/breadcrumb/breadcrumb.lite.tsx
sarahbrng: Fixed: introduced one shared ellipsis label constant and reused it in defaults and runtime fallback.

packages/components/src/components/breadcrumb/breadcrumb.lite.tsx
sarahbrng: Collapsing via pure CSS cannot reliably manage aria-expanded state, dynamic maxItems behavior, and semantic collapsed-tail computation from mixed input data.

packages/components/src/components/breadcrumb/breadcrumb.lite.tsx
sarahbrng: Supporting arbitrary children collapse would require DOM analysis/mount-time transforms and introduces cross-framework inconsistency, so I’d keep this scoped to items for now.

packages/components/src/components/breadcrumb/breadcrumb.lite.tsx
sarahbrng: Fixed: Replaced breadcrumb-specific uniqueId with shared state._id from GlobalState and updated list ID fallback plus types accordingly.

packages/components/src/components/breadcrumb/breadcrumb.lite.tsx
sarahbrng: The ellipsis button is already styled to align with breadcrumb item spacing/typography and avoids coupling to button component internals.

packages/components/scripts/post-build/stencil.ts
sarahbrng: Fixed: Removed the obsolete DBIcon import replacement in stencil post-build. The script now keeps only the required transformations.

packages/components/src/components/breadcrumb/breadcrumb.scss
sarahbrng: Fixed: removed duplicated ellipsis visual/typography states so the button keeps only its button-specific reset and spacing styles.

packages/components/src/components/breadcrumb/breadcrumb.scss
sarahbrng: Thanks, addressed. The redundant breadcrumb-level styling was removed and the item-related styling is now handled in the breadcrumb-item encapsulation.

packages/components/src/components/breadcrumb/breadcrumb.scss
sarahbrng: Good point. Vertical centering is now ensured at list/item level, so alignment is consistent across elements.

packages/components/src/components/breadcrumb/breadcrumb.scss
sarahbrng: Agreed, those redundant declarations were removed from the breadcrumb ellipsis block.

packages/components/src/components/breadcrumb/breadcrumb.scss
sarahbrng: Implemented. The slash separator now follows the icon size variable and uses a square box via aspect-ratio for consistent sizing. I also added a medium-size override so slash and chevron stay aligned.

packages/components/src/components/breadcrumb/breadcrumb.scss
sarahbrng: Addressed together with the related note in r2715826033 by removing the redundant breadcrumb-level declarations.

packages/components/src/components/breadcrumb/breadcrumb.scss
sarahbrng: Yes, this now also applies to small. The default branch already centers the list items vertically.

packages/components/src/components/breadcrumb/breadcrumb.scss
sarahbrng: Done via the cleanup from r2715826033; the removable part is no longer present.

packages/components/src/components/breadcrumb/breadcrumb.scss
sarahbrng: Addressed already in the earlier cleanup (r2715803409). The redundant ellipsis styling is removed.

packages/components/src/components/breadcrumb/model.ts
sarahbrng: Fixed: Removed the redundant re-export from breadcrumb model and kept the type import for local usage to reduce API duplication

showcases/angular-showcase/src/app/components/breadcrumb-item/breadcrumb-item.component.ts
sarahbrng: Fixed: The TODO has already been removed; web-components now import only DefaultComponent, while the Angular component import remains only for the non-web-components path.

packages/components/src/components/breadcrumb/breadcrumb.scss
sarahbrng: Fixed: Removed redundant local typography declarations and aligned item typography with shared font-size placeholders.

packages/components/src/components/breadcrumb/breadcrumb.scss
sarahbrng: Fixed: Removed redundant local transition handling and aligned with shared/global transition behavior.

packages/components/src/components/breadcrumb/breadcrumb.scss
sarahbrng: Fixed: Removed redundant local typography declarations and aligned item typography with shared font-size placeholders.

packages/components/src/components/breadcrumb/breadcrumb.scss
sarahbrng: Fixed: Applied the referenced typography change through shared font-size placeholders.

packages/components/src/components/breadcrumb/breadcrumb.scss
sarahbrng: Fixed: Replaced manual small typography values with the shared small font-size placeholder as suggested.

packages/components/src/components/breadcrumb/breadcrumb.scss
sarahbrng: Fixed: Removed obsolete explicit default declarations from the item styling.

packages/components/src/components/breadcrumb/breadcrumb.scss
sarahbrng: Fixed: Switched hover and active states to shared helper mixins for consistent behavior.

packages/components/src/components/breadcrumb/breadcrumb.scss
sarahbrng: Fixed: Removed the no-longer-needed local reduced-motion transition override with the cleanup of redundant local transitions.

packages/components/src/components/breadcrumb/breadcrumb.scss
sarahbrng: Fixed: Removed local focus-visible override and rely on the existing global interactive focus styles.

packages/components/src/components/breadcrumb/breadcrumb.scss
sarahbrng: Fixed: Replaced manual medium typography values with the shared medium font-size placeholder.

packages/components/src/components/breadcrumb/breadcrumb.scss
sarahbrng: Fixed: Removed redundant icon-size overrides tied to data-icon-weight and kept icon handling in shared patterns.

packages/components/src/components/breadcrumb/breadcrumb.scss
sarahbrng: Fixed: Removed unnecessary static declarations from the old local style block during cleanup.

packages/components/src/components/breadcrumb/breadcrumb.scss
sarahbrng: Fixed: Removed unnecessary static declarations from the old local style block during cleanup.

packages/components/src/components/breadcrumb/breadcrumb.scss
sarahbrng: Fixed: Removed redundant local focus handling and rely on established global focus-visible behavior.

packages/components/src/components/breadcrumb/breadcrumb.lite.tsx
sarahbrng: Fixed: Refactored Breadcrumb to use DBBreadcrumbItem directly across all render paths instead of rebuilding item markup.

mfranzke:
@codex review

PR #5624, opened 2025-12-11
A11y-Update für DBSwitch

sarahbrng:
## Proposed changes

Das Problem: Bisher wurden ARIA-Labels beim Laden der Komponente einmalig gesetzt ("Snapshotted"). Bei dynamischen Änderungen zur Laufzeit (z. B. Sprachwechsel via i18n) blieb der Screenreader-Text veraltet.

Die Lösung: Die Inputs ariaLabel und ariaLabelledBy sind nun reaktive Props somit werden Änderungen sofort an das innere `input[type="checkbox"]` weitergegeben. Der Accessible Name ist immer live aktuell.

`aria-label` bindet an `ariaLabel`.

`aria-labelledby` bindet an `ariaLabelledBy`.

resolves #5495 

## Types of changes

## Further comments

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/5495-changing-language-with-aria-labels-not-correctly-applied-to-elements>


packages/components/src/components/switch/switch.lite.tsx
```diff
@@ -213,7 +215,7 @@ export default function DBSwitch(props: DBSwitchProps) {
 			data-hide-asterisk={getHideProp(props.showRequiredAsterisk)}
 			data-custom-validity={props.validation}
 			class={cls('db-switch', props.className)}>
-			<label htmlFor={state._id}>
+			<label htmlFor={state._id} id={state._labelId}>
```
michaelmkraus, line 219:
```suggestion
			<label id={state._labelId}>
```

`htmlFor` is redundant in this case because the `<input>` is wrapped by the `<label>`.

packages/components/src/components/switch/switch.lite.tsx
```diff
@@ -226,6 +228,8 @@ export default function DBSwitch(props: DBSwitchProps) {
 						props.validation === 'invalid' ? 'true' : undefined
 					}
 					aria-describedby={state._descByIds}
+					aria-label={props.ariaLabel}
+					aria-labelledby={props.ariaLabelledBy ?? state._labelId}
```
michaelmkraus, line 233: According to the ARIA accessible name computation ([W3C](https://www.w3.org/TR/accname-1.2/)) and ARIA Authoring Practices, `aria-labelledby` takes precedence over `aria-label`. If both are present, `aria-label` is ignored. 
So if `aria-labelledby` is present, wo do not need to set `aria-label`.

packages/components/src/components/switch/switch.lite.tsx
mfranzke: it's (still) necessary to connect `label`-HTML-elements explicitly even if their related form elements are nested, as speech recognition software would rely on this, compare to a [LinkedIn article by Nina Jameson](https://www.linkedin.com/posts/nina-jameson-355a201aa_warum-das-schachteln-von-input-und-label-activity-7259825320302657536-Kf5R).

michaelmkraus: Good point! I didn't know and hadn't considered that.

michaelmkraus (dismissed):

packages/components/src/components/switch/switch.lite.tsx
mfranzke, line 1: this is a fix we would like to communicate, so a changeset is missing in this.

mfranzke, line 1: I'm struggling with the approach to only making some of those aria-*-HTML-attribute in one component in our system "reactive" to changes might not be intuitive to the developers. I would have assumed that the existing mechanism for pass-throughs of those attributes even already takes care of this stuff, elsewhere we should optimize that functionality from my point of view, as this missing functionality would be expected equally at other attributes as well.

mfranzke:
@michaelmkraus @sarahbrng this PR is still draft, but approved and in "ready for release" state as well.

sarahbrng:
> only making some of those `aria-*`-HTML-attribute reactive to changes might not be intuitive to the developers. I would have assumed that the existing mechanism for pass-throughs of those attributes even already takes care of this stuff, elsewhere we should optimize that functionality from my point of view, as this missing functionality would be expected equally at other attributes as well.

Yes, that is right:
I suggest we investigate improving the core attribute pass-through mechanism in our component framework (Mitosis) to ensure all ARIA attributes are reactive by default. This would:

- Fix the issue systematically across all components
- Align with developer expectations
- Prevent similar issues in the future

I suggest: Keep this PR as-is (targeted fix for the immediate accessibility bug) and create a follow-up issue to address the architectural concern, what do you think @mfranzke ?

mfranzke:
> > only making some of those `aria-*`-HTML-attribute reactive to changes might not be intuitive to the developers. I would have assumed that the existing mechanism for pass-throughs of those attributes even already takes care of this stuff, elsewhere we should optimize that functionality from my point of view, as this missing functionality would be expected equally at other attributes as well.
> 
> 
> 
> Yes, that is right:
> 
> I suggest we investigate improving the core attribute pass-through mechanism in our component framework (Mitosis) to ensure all ARIA attributes are reactive by default. This would:
> 
> 
> 
> - Fix the issue systematically across all components
> 
> - Align with developer expectations
> 
> - Prevent similar issues in the future
> 
> 
> 
> I suggest: Keep this PR as-is (targeted fix for the immediate accessibility bug) and create a follow-up issue to address the architectural concern, what do you think @mfranzke ?

Let's discuss in the team next week. Providing this fix in the meantime would introduce this "inconsistency" (not a huge thing, but still), and we would be able to wait for the big solution.

mfranzke:
closing in favour of https://github.com/db-ux-design-system/core-web/pull/5728

PR #5655, opened 2025-12-18
feat: add DBFooter component

### LINKED ISSUE #534: 1️⃣ 👁️ Footer (Beta Release)
### Beschreibung

Dieses Epic umfasst alle Aufgaben zur Implementierung, Dokumentation, Qualitätssicherung und Beta-Veröffentlichung des Inhalts **Footer** für das DB UX Design System.
Ziel ist eine konsistente, barrierefreie und dokumentierte Komponente, die DS Nutzenden als Beta zur Verfügung steht.

### Akzeptanzkriterien

Zielzustand → Ready for Beta Release

✅ High Design Quality (Design Perspective)
✅ High Development Quality (Development Perspective)
✅ Understandable (User Perspective)
✅ Usable (User Perspective)

sarahbrng:
## Proposed changes

Adds `DBFooter` as a composable set of components for consistent application footers.

- `DBFooter` renders a native `<footer>` landmark and the layout container, nothing else.
- `DBFooterContent` is the primary footer area, `DBFooterMeta` the secondary one for legal and comparable content. Areas are included or left out by composing the respective subcomponent, so there are no `showMain` / `showMeta` / `showCopyright` toggles and no named slots.
- `DBFooterMeta` takes an optional `copyright` string. There is no Deutsche Bahn specific default; without a value no copyright element is rendered at all.
- `DBFooterMeta` is semantically neutral and creates no navigation landmark. Navigational content therefore has to be wrapped in a labelled `<nav>`, other secondary content such as contact details works as well.
- `width` comes from the shared `ContainerWidthProps` and supports `full`, `large`, `medium` and `small`. The footer and both areas stay full width, only the centred inner content is constrained.
- Uses existing DB background, border, spacing, typography and screen size tokens.
- Copyright and meta content wrap without horizontal overflow on narrow viewports.
- Deliberately no mobile accordion, so footer links stay available to assistive technologies and browser search.
- Supports global IDs and classes including the `propOverrides.id` fallback.
- `db-footer`, `db-footer-content` and `db-footer-meta` are registered as custom element hosts, so the Web Component wrappers get the repository's `display: contents` workaround.
- Documents usage inside `DBPage variant="fixed"` for HTML, React, Angular and Vue with semantic, labelled navigation lists.
- Generates idiomatic default slots for Angular, React, Vue and Web Components without a Footer specific Vue regex workaround.

resolves #1425

## Validation

- `pnpm run build`
- `pnpm run test`
- `pnpm run lint`
- `pnpm run build-outputs`
- React and Vue Footer component tests on Chromium
- React and Vue Footer ARIA snapshots on Chromium and Mobile Chrome
- Covered by tests: width variants, optional areas, boolean strings, IDs, classes, narrow viewports, the `contentinfo` landmark, ARIA snapshots and Axe

### Code Quality

### Validation

### General

## Open points for review

- Design input still needed on three questions raised in review: whether the footer may vary in height, whether a smaller link text size should be enforced (`.db-footer-copyright` is `body-sm` while a default `DBLink` is larger), and real world examples from design.
- `resolves #1425` points at a 2023 dependabot PR rather than a footer issue; the correct reference needs to be filled in.

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/1425-footer-concept-phase-2>


.vscode/settings.json
mfranzke, line 1: @sarahbrng why did you remove this file ?

.gitignore
```diff
@@ -4,6 +4,7 @@ node_modules
 npm-debug.log
 .history
 .idea
+.vscode/
```
mfranzke:
```suggestion
```

this is a legitimate folder that we still would like to use.

packages/components/src/components/footer/docs/HTML.md
```diff
@@ -0,0 +1,54 @@
+## HTML
+
+For general installation and configuration take a look at the [wc-core-components](https://www.npmjs.com/package/@db-ux/wc-core-components) package.
+
+### Use component
+
+#### Simple
+
+```html
+<db-footer>
```
mfranzke: This would need to be the HTML tag instead of the Custom Element (like all the other following Custom Elements as well).

packages/components/src/components/footer/docs/React.md
```diff
@@ -0,0 +1,92 @@
...
+#### With custom content
+
+```tsx App.tsx
+// App.tsx
+import { DBFooter, DBLink } from "@db-ux/react-core-components";
+
+const App = () => (
+	<DBFooter
+		main={
+			<div>
```
mfranzke: This would need to be an unordered list. Let's align on this topic.

showcases/showcase-styles.css
```diff
@@ -1,6 +1,6 @@
 /* Styles for the showcases */
 .desktop-navigation ul {
-	list-style-type: "";
+	list-style-type: none;
```
mfranzke:
```suggestion
	list-style-type: "";
```

The [value 'none' results in a Safari VoiceOver behaviour](https://www.matuzo.at/blog/2023/removing-list-styles-without-affecting-semantics) we'd like to mitigate.

.vscode/settings.json
sarahbrng: Keine Ahnung :) 
Ist reverted !

packages/components/src/components/footer/docs/React.md
sarahbrng: Fixed: Updated the footer navigation example to use semantic unordered-list markup (`ul`/`li`) for links as requested.

mfranzke:
@codex review

packages/components/src/components/footer/footer.lite.tsx
```diff
@@ -0,0 +1,56 @@
...
+});
+
+export default function DBFooter(props: DBFooterProps) {
+	// This is used as forwardRef
+	const _ref = useRef<HTMLDivElement | any>(null);
+
+	return (
+		<footer
+			ref={_ref}
+			id={props.id}
```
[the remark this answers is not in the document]
michaelmkraus: Fixed: DBFooter now uses `props.id ?? props.propOverrides?.id`, and a regression test covers both the explicit ID and fallback cases.

mfranzke:
@codex review

packages/components/src/components/footer/footer.lite.tsx
```diff
@@ -0,0 +1,54 @@
...
+			id={props.id ?? props.propOverrides?.id}
+			class={cls('db-footer', props.className)}
+			data-width={props.width}>
+			{getBoolean(props.showMain, 'showMain') && (
+				<div class="db-footer-main">
+					<div class="db-footer-content-container">
+						{props.children}
+					</div>
+				</div>
+			)}
```
nmerget: We need to create a sub-component instead of having `showMain`.

And I'm not convinced about the "main" naming. Could this lead to some missunderstanding with `<main>`?

`<DBFooterMain>` hmm, not sure about it, maybe you can ask AI about this.

packages/components/src/components/footer/footer.lite.tsx
```diff
@@ -0,0 +1,54 @@
...
+							) && (
+								<p class="db-footer-copyright">
+									© Deutsche Bahn AG
+								</p>
+							)}
+							<Slot name="meta" />
+						</div>
+					</div>
+				</div>
+			)}
```
nmerget: https://github.com/db-ux-design-system/core-web/pull/5655/changes#r3860087423 <- Same here, we should create a sub-component for this instead of having `showMeta` so there will be no `<Slot name="meta" />` at all. There are just `props.children` inside the <footer> and then we can work with a `display:grid` and `:has(.db-footer-main)` etc.

The name "meta" is also something we should think about it. We already have `ControlPanelMeta` for the new Shell. We should think about it if we want to have the same name or choose another one to avoid confusion, this is a discussion with design and maybe also asking AI

packages/components/src/components/footer/footer.lite.tsx
```diff
@@ -0,0 +1,54 @@
...
+				<div class="db-footer-meta">
+					<div class="db-footer-content-container">
+						<div class="db-footer-meta-inner">
+							{getBoolean(
+								props.showCopyright,
+								'showCopyright'
+							) && (
+								<p class="db-footer-copyright">
+									© Deutsche Bahn AG
+								</p>
```
nmerget: The new sub-component should have a new property `copyright` instead of a hardcoded `Deutsche Bahn AG`. We are whitelabel in the first place. We could also think of another way providing the copyright with a 

````
.db-footer-copyright{
gap: 1ch;
&:after{content:(--db-copyright-label)}}
```` 

With this we could move it to the db-theme or sbahn-theme etc. and could change it there. Just some idea, the property might be better.

File: packages/components/src/components/footer/examples/width.example.lite.tsx
Comment (nmerget) [inline packages/components/src/components/footer/examples/width.example.lite.tsx:1]: Image

All width's are identical

File: packages/components/src/components/footer/examples/show-main.example.lite.tsx [hunk 9]
```diff
@@ -0,0 +1,67 @@
...
+				gap: 'var(--db-spacing-fixed-sm)',
+				width: '100%'
+			}}>
+			<DBFooter
+				showMain
+				meta={
+					<nav aria-label="Show main enabled legal navigation">
+						<ul>
+							<li>
+								<a className="db-link" href="#privacy">
```
nmerget: We should use the `<DBLink>` here instead. Same goes for every example

packages/components/src/components/footer/footer.lite.tsx
```diff
@@ -0,0 +1,54 @@
...
+							)}
+							<Slot name="meta" />
```
nmerget: Do we allow other children then `<a>` to come into the meta? Otherweise we could wrap it directly into an `<nav>`?

packages/components/src/components/footer/footer.lite.tsx
```diff
@@ -0,0 +1,54 @@
...
+export default function DBFooter(props: DBFooterProps) {
+	// This is used as forwardRef
+	const _ref = useRef<HTMLElement | null>(null);
+
+	return (
+		<footer
+			ref={_ref}
+			id={props.id ?? props.propOverrides?.id}
+			class={cls('db-footer', props.className)}
+			data-width={props.width}>
```
nmerget, line 17: We need to test this inside the new `DBShellContent`. I guess we need some other way if it is indide the shell.

packages/components/src/components/footer/footer.lite.tsx
```diff
@@ -0,0 +1,54 @@
...
+	showMain: true,
+	showMeta: true
+});
+
+export default function DBFooter(props: DBFooterProps) {
+	// This is used as forwardRef
+	const _ref = useRef<HTMLElement | null>(null);
+
+	return (
+		<footer
```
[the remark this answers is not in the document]
michaelmkraus: I added `db-footer` to the custom-element host registry so it receives the standard `display: contents` workaround.

packages/components/src/components/footer/model.ts
```diff
@@ -0,0 +1,33 @@
...
+
+	/**
+	 * Shows or hides the meta footer area.
+	 */
+	showMeta?: boolean | string;
+
+	/**
+	 * Limits and centres the inner content container; the footer itself stays full width.
+	 */
+	width?: MaxWidthType;
```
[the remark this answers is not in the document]
michaelmkraus: I added `ContainerWidthProps` into `DBFooterProps` and remove the duplicated width declaration.

packages/components/src/components/footer/footer.scss
```diff
@@ -0,0 +1,69 @@
...
+		align-items: center;
+		flex-wrap: wrap;
+		gap: variables.$db-spacing-fixed-lg;
+		min-inline-size: 0;
+	}
+
+	.db-footer-meta-inner > :not(.db-footer-copyright) {
+		flex: 1 1 auto;
+		max-inline-size: 100%;
+		min-inline-size: 0;
```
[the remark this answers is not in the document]
michaelmkraus: I used `DBLink` with wrap in the examples and cover that usage in the narrow-viewport regression test.

packages/components/src/components/footer/footer.lite.tsx
michaelmkraus: The primary footer area is now represented by the new `DBFooterContent` subcomponent. Consumers can include or omit it instead of using the `showMain` property. `DBFooter` itself now only provides the native footer landmark and layout container.

packages/components/src/components/footer/footer.lite.tsx
michaelmkraus: The named meta slot and the `showMeta` property have been replaced by the new `DBFooterMeta` subcomponent. The component remains semantically neutral so consumers can provide navigation or any other secondary footer content as needed.

packages/components/src/components/footer/footer.lite.tsx
michaelmkraus: `DBFooterMeta` now provides an optional `copyright` string property. There is no longer a Deutsche Bahn-specific default, and no copyright element is rendered unless the consumer explicitly provides a value.

packages/components/src/components/footer/examples/width.example.lite.tsx
michaelmkraus: I updated the width example to use the existing showcase marker inside each real `DBFooterContent` container, so the constrained inner widths are now visible.

packages/components/src/components/footer/examples/show-main.example.lite.tsx
michaelmkraus: Done.

packages/components/src/components/footer/footer.lite.tsx
michaelmkraus: The documentation now explicitly states that `DBFooterMeta` is semantically neutral and does not create a navigation landmark. Navigational content must therefore be wrapped in a labelled `<nav>`, while other secondary content such as contact information is also supported.

packages/components/src/components/footer/footer.lite.tsx
michaelmkraus: Passing `DBFooter` through `endSlot` works correctly in both React and Vue: the footer remains outside the scrollable `<main>`, stays fixed while the main area scrolls, preserves the full-width stacked content and meta areas, exposes exactly one `contentinfo` landmark, and passes Axe.

nmerget:
image
We would use an `inline` DBLink for such an example, and also some space between them.

Some other question, which has to be challenged with design. Is it allowed to extend the footer in height? Right now there is a `<p>` with a margin which makes the fotter bigger. If we have just the link inside the height would be smaller.

image

Another point: It looks wired that the Link is way bigger than the copyright, can you ask Design if we need to force a lower text size. Also they should provide some "real world" examples so we can preview some in the examples for testing

nmerget:
image
The width is still not correct inside the examples

nmerget:
@michaelmkraus There is a scrollbar for the size examples:

image


The link sticks to the "Customer services:":

image

.changeset/add-footer-complex-composition-example.md
nmerget, line 1: Can we just use 1 changeset for the footer?

packages/components/src/components/footer-meta/footer-meta.lite.tsx
```diff
@@ -0,0 +1,32 @@
...
+	// This is used as forwardRef
+	const _ref = useRef<HTMLDivElement | any>(null);
+
+	return (
+		<div
+			ref={_ref}
+			id={props.id ?? props.propOverrides?.id}
+			class={cls('db-footer-meta', props.className)}>
+			<div class="db-footer-content-container">
+				<div class="db-footer-meta-inner">
```
nmerget: Why do we need 2 `<div>` here? The  `db-footer-content-container` seems to be false

Maybe we can use `display: grid` to achive the same?

packages/components/src/components/footer-meta/footer-meta.lite.tsx
```diff
@@ -0,0 +1,32 @@
...
+					<Show when={props.copyright}>
+						<p class="db-footer-copyright">{props.copyright}</p>
```
nmerget: There is no `©` symbol before this element. We should add it here to have a real benefit from the `props.copyright`.

packages/components/src/components/footer/index.html
nmerget, line 1: We removed the `index.html` files don't we?

michaelmkraus: Removed.

packages/components/src/components/footer-meta/footer-meta.lite.tsx
michaelmkraus: Added the symbol to the template.

.changeset/add-footer-complex-composition-example.md
michaelmkraus: We now have only one changeset.

packages/components/src/components/footer-meta/footer-meta.lite.tsx
michaelmkraus: Removed the inner wrapper. The shared width container carries the meta row now. Also renamed it to `db-footer-container`. Kept flex. The row has to wrap while the non-copyright child takes the remaining space via `flex: 1 1 auto`.

packages/components/src/components/footer/examples/copyright.example.lite.tsx
```diff
@@ -0,0 +1,87 @@
...
+					<nav aria-label="Legal navigation with copyright">
+						<ul
+							style={{
+								display: 'flex',
+								flexWrap: 'wrap',
+								gap: 'var(--db-spacing-fixed-md)',
+								listStyleType: '""',
+								margin: '0',
+								padding: '0'
+							}}>
```
nmerget: Maybe in future the developers use `DBList` for this case. But for now we should move this to the `DBFooterMeta` to apply it for every `ul,ol` to avoid strange looking lists inside the meta

michaelmkraus: Flex row moved into `footer-meta.scss`. The other declarations were already covered by the foundations' nav reset.

packages/components/src/components/footer/examples/complex-composition.example.lite.tsx
```diff
@@ -48,10 +48,7 @@ export default function FooterComplexComposition() {
 							<ul
 								style={{
 									display: 'grid',
-									gap: 'var(--db-spacing-fixed-xs)',
-									listStyleType: '""',
-									margin: '0',
-									padding: '0'
+									gap: 'var(--db-spacing-fixed-xs)'
```
nmerget, line 55: Why do we override the default here?

michaelkraus: To get vertical spacing. The foundations `nav` reset zeroes the list margins and `li` has no default spacing, so the links would sit with line-height only.

mfranzke:
@codex review

__snapshots__/footer/component/chromium/DBFooter-should-match-screenshot.png
nmerget, line 1: The links in the meta are bigger than the copyright

__snapshots__/footer/showcase/mobile-chrome/DBFooter-should-match-screenshot-1/DBFooter-should-match-screenshot.png
nmerget, line 1: 1. The gap inside the meta between a copyright and links are very big, this might not be right
2. The `width` examples look wrong, on mobile devices they should fill the full space

showcases/showcase-styles.css
```diff
@@ -219,6 +219,53 @@
...
+	gap: var(--db-spacing-fixed-sm);
+}
+
+/*
+ * 1cqw makes 1em equal 1% of the wrapper width, so the em-based max-inline-size of each
+ * variant becomes that percentage. DBFooterContent and DBFooterMeta both use this
+ * container, so content and meta stay aligned.
+ */
+.footer-width-example .db-footer-container {
+	font-size: 1cqw;
```
nmerget: Why do we use this? :O

Our typo is already changing font sizes based on density and device size. This is nothing we want to do in our design system

showcases/showcase-styles.css
```diff
@@ -219,6 +219,53 @@
...
+ * four look identical. Because the clamp is em-based while every typography and spacing
+ * token is rem-based, the container's own font-size scales the clamp and nothing else.
+ * Turning this wrapper into an inline-size container therefore lets the variants render
+ * at 48%, 64%, 90% and 100% of the available width, which keeps them distinguishable at
+ * any card width and needs no scroll container.
+ */
+.footer-width-example {
+	container-type: inline-size;
+	inline-size: 100%;
+}
```
nmerget, line 224: I guess this is not required. The mobile view should fill the complete width. Based on the screen of the snapshot full & large may look equal. But we shouldn't fake something like this, we wouldn't have any hint if something goes wrong

packages/components/src/components/footer/footer.scss
```diff
@@ -0,0 +1,24 @@
+@use "@db-ux/core-foundations/build/styles/screen-sizes";
+
+.db-footer {
+	flex-direction: column;
```
nmerget, line 12: There isn't a `display: flex;` to apply to

packages/components/src/components/footer-meta/footer-meta.scss
```diff
@@ -0,0 +1,59 @@
...
+@use "@db-ux/core-foundations/build/styles/helpers";
+@use "@db-ux/core-foundations/build/styles/colors";
+
+.db-footer-meta {
+	background-color: colors.$db-adaptive-bg-basic-level-2-default;
+	border-block-start: variables.$db-border-width-3xs solid
+		colors.$db-adaptive-on-bg-basic-emphasis-60-default;
+	inline-size: 100%;
+	padding-block: variables.$db-spacing-fixed-xs;
+	padding-inline: variables.$db-spacing-fixed-lg;
```
nmerget:
```suggestion
	padding: variables.$db-spacing-fixed-xs variables.$db-spacing-fixed-lg;
```

packages/components/src/components/footer/examples/optional-areas.example.lite.tsx
```diff
@@ -0,0 +1,58 @@
...
+	storybookNames: ['Content only', 'Meta only'],
+	storybookArgTypes: StorybookFooterArgTypes
+});
+
+export default function FooterOptionalAreas() {
+	return (
+		<div
+			style={{
+				display: 'grid',
+				gap: 'var(--db-spacing-fixed-sm)',
```
nmerget, line 19:
```suggestion
				gap: '0',
```
There shouldn't be a gap between the footer areas

packages/components/src/components/footer/examples/optional-areas.example.lite.tsx
```diff
@@ -0,0 +1,58 @@
...
+					 * link. Mitosis drops explicit JSX whitespace expressions
+					 * during generation, so a text space would not survive
+					 * into the framework outputs. */}
+					<p
+						style={{
+							font: 'var(--db-type-body-sm)',
+							margin: '0',
+							display: 'flex',
+							flexWrap: 'wrap',
+							gap: 'var(--db-spacing-fixed-3xs)'
```
nmerget:
```suggestion
							gap: '0.5ch'
```

showcases/patternhub/components/component-parser/index.tsx
```diff
@@ -374,6 +376,22 @@ const ComponentSwitch = ({
 		);
 	}
 
+	if (type === 'footer-content') {
```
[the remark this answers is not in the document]
michaelmkraus: `DBFooter` is now imported and handled as `type: "footer"`, plus added to the type union in `data.ts`.

showcases/patternhub/tests/default.spec.ts
```diff
@@ -18,6 +18,10 @@ const getDefaultScreenshotTest = async (
 
 for (const group of Components) {
 	for (const component of group.subNavigation) {
+		if (component.isHiddenInMenu === true) {
+			continue;
+		}
```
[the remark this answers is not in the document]
michaelmkraus: `accordion-item` and `tab-item` lost all six baselines' tests. The exclusion is now limited to `footer-content` and `footer-meta` by name. The actual criterion is a missing entry in the Patternhub `nameComponentMap`, which only applies to those two.

showcases/patternhub/data/components.json
```diff
@@ -151,6 +151,30 @@
...
+					},
+					{
+						"label": "DBFooterMeta Properties",
+						"path": "/components/footer-meta/properties"
+					}
+				]
+			},
+			{
+				"label": "DBFooterContent",
+				"name": "footer-content",
```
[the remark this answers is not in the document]
michaelmkraus: Exclude the new subcomponents from the test table

__snapshots__/footer/component/chromium/DBFooter-should-match-screenshot.png
michaelmkraus: Fixed. They now use `DBLink` with `variant="inline"` and `size="small"`.

__snapshots__/footer/showcase/mobile-chrome/DBFooter-should-match-screenshot-1/DBFooter-should-match-screenshot.png
michaelmkraus: The meta gap is now split into row and column: the row gap only kicks in once copyright and content wrap, where the horizontal value was far too wide. Removed the scaling, so on mobile all variants fill the full width as they actually do.

showcases/showcase-styles.css
michaelmkraus: Agreed, removed.

showcases/showcase-styles.css
michaelmkraus: Removed. The variants now look identical when there isn't enough width, which is the real behaviour, and a regression in the breakpoints would actually show up.

packages/components/src/components/footer/examples/optional-areas.example.lite.tsx
michaelmkraus: Applied.

packages/components/src/components/footer-meta/footer-meta.scss
michaelmkraus: Applied.

packages/components/src/components/footer/examples/optional-areas.example.lite.tsx
michaelmkraus: These are two separate `DBFooter` elements, one with content only and one with meta only. 
With `gap: 0` they'd stick together and read as a single footer with both areas, plus you'd see their two `border-block-start` lines stacked.

packages/components/src/components/footer/footer.scss
michaelmkraus: Removed. Was a left-over.

michaelmkraus: It was wrong to remove it. `page.scss` applies display: `flex to > :is(.db-header, .db-footer, header, footer)` inside `.db-page[data-variant="fixed"]`, so the footer is a flex container there and without `column` the content and meta areas lay out side by side (verified: 195px each of 390px). Restored it it.

packages/components/src/components/footer/examples/optional-areas.example.lite.tsx
nmerget: Maybe we could add some Infotext + linebreak there with `'Content only', 'Meta only'` to make sure people aren't as confused as I was. Maybe this makes sense for every example with multiple footers?

packages/components/src/components/footer-meta/footer-meta.scss
```diff
@@ -0,0 +1,65 @@
...
+
+		align-items: center;
+
+		/*
+		 * Row and column gap differ: the row gap only applies once the copyright
+		 * and the secondary content wrap onto separate lines, which is the
+		 * narrow-viewport case. Reusing the horizontal value there would space
+		 * them far too widely apart.
+		 */
+		gap: variables.$db-spacing-fixed-sm variables.$db-spacing-fixed-lg;
```
nmerget:
```suggestion
		gap: variables.$db-spacing-fixed-xs variables.$db-spacing-fixed-lg;
```

michaelmkraus: Applied.

packages/components/src/components/footer-meta/footer-meta.scss
```diff
@@ -0,0 +1,110 @@
...
+		colors.$db-adaptive-on-bg-basic-emphasis-60-default;
+
+	/*
+	 * The whole strip is small text, as in the design. Without it the list items
+	 * keep the inherited body size and their taller line box pushes the links
+	 * a few pixels below the copyright next to them. Note that this also changes
+	 * what `em` means inside the area, which is why the width clamps in
+	 * footer.scss use `rem`.
+	 */
+	font: var(--db-type-body-sm);
```
nmerget: We have a placeholder for this `@extend %db-overwrite-font-size-sm;` which set's the icons sizes as well

packages/components/src/components/footer-meta/footer-meta.scss
```diff
@@ -0,0 +1,110 @@
...
+		 * Deliberately not a container query: making the footer a size container
+		 * stops its contents from contributing to its intrinsic width, so any
+		 * content-sized ancestor - a centred flex row, a `fit-content` box, a
+		 * table cell - collapses to zero width. Deliberately not a media query
+		 * either: that measures the viewport, so a footer narrowed by its
+		 * surroundings (a shell with an open side panel, a `width` variant on a
+		 * wide screen) would keep the side-by-side layout without the room for it.
+		 */
+		> :not(.db-footer-copyright) {
+			flex: 1 1 calc((#{$layout-break} - 100%) * 999);
```
nmerget: This looks way to complex. When I see the design there are 2 variants of the meta.

1x Desktop and 1x Mobile (< breakpoint-md).

We should just add another `<div>` around the `props.children`. If it is empty it gets `display:none` otherwise it gets a `display:flex` with `flex-wrap` then you can set `flex-direction` based on the media-query for `.db-footer-container`. This would reflext 1:1 the design

packages/components/src/components/footer-content/footer-content.scss
```diff
@@ -0,0 +1,29 @@
...
+	nav,
+	[role="navigation"] {
+		ol,
+		ul {
+			@include helpers.display(flex);
+
+			flex-wrap: wrap;
+			gap: variables.$db-spacing-fixed-md;
+		}
+	}
```
nmerget: This is the same inside `footer-meta` we should share it

packages/components/src/components/footer-meta/footer-meta.scss
```diff
@@ -0,0 +1,110 @@
...
+			@include helpers.display(flex);
+
+			flex-wrap: wrap;
+			gap: variables.$db-spacing-fixed-xs variables.$db-spacing-fixed-md;
+		}
+	}
+
+	.db-footer-copyright {
+		color: colors.$db-adaptive-on-bg-basic-emphasis-100-default;
+		font: var(--db-type-body-sm);
```
nmerget: We have a placeholder for this @extend %db-overwrite-font-size-sm; which set's the icons sizes as well

packages/components/src/components/footer-meta/footer-meta.scss
michaelmkraus: Applied.

packages/components/src/components/footer-meta/footer-meta.scss
michaelmkraus: Removed. `font` and the icon variables inherit, so the one `@extend` on `.db-footer-meta` covers this.

packages/components/src/components/footer-meta/footer-meta.scss
michaelmkraus: Removed. `font` and the icon variables inherit.

packages/components/src/components/footer-content/footer-content.scss
michaelmkraus: Shared. The rule now lives once in `footer.scss`.

packages/components/src/components/footer-meta/footer-meta.scss
michaelmkraus: Done. Wrapper is `.db-footer-meta-content`, switch is `screen-sizes.screen("sm", "max")`, empty wrapper gets `display: none`.

nmerget (approved):

PR #5992, opened 2026-02-05
feat: loading indicator

nmerget:
## Proposed changes

Add `DBLoadingIndicator` component

## Types of changes

## Further comments

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/feat-loading-indicator>


packages/components/src/components/loading-indicator/loading-indicator.scss
```diff
@@ -0,0 +1,515 @@
...
+	div {
+		display: flex;
+
+		> label {
+			display: inline-flex;
+			align-items: center;
+
+			> progress {
+				display: none;
+			}
```
mfranzke: this semantic HTML element shouldn't be totally hidden, as it would provide any purpose anymore.

packages/components/src/components/loading-indicator/loading-indicator.lite.tsx
```diff
@@ -0,0 +1,253 @@
...
+							: '32 32 64 64'
+					}
+					aria-hidden="true">
+					<circle class="db-loading-indicator-circle-track" />
+					<circle class="db-loading-indicator-circle-segment" />
+				</svg>
+			</Show>
+
+			<div role={state.getRole()}>
+				<label
```
mfranzke: this element doesn't have a `for`-attribute to connect it to the `progress`-HTML-element.

jonaskuske:
I've found the `autoDisable` behavior to be quite the foot gun, if you do sth. like

```html
<button @click="thingThatLoads()">
@if (loading) {
  <loading-indicator />
}
Click me
</button>
```

then the button will stay permanently disabled after clicking it once and the UI is broken.

Can ofc be avoided through a strict lint rule that enforces usage via state="inactive" instead of conditional rendering, but making the destroy/cleanup callback reset the parent button to enabled/initial state would be a good safeguard still, imo

jonaskuske:
I've added this loading component to our internal component lib to bridge the time until it is officially released fyi – also found some issues with spacing and sizings, and with the timeout functionality.

I'm gone for a week now, can report in more detail once I'm back :)

packages/components/src/components/loading-indicator/loading-indicator.scss
nmerget: Good catch. The native `<progress>` is no longer removed from the accessibility tree: I replaced `display: none` with the design system's `%a11y-visually-hidden` helper, so it stays available to screen readers (announcing the loading state/value) while remaining visually hidden. The visible value is still shown via the adjacent `aria-hidden` `<span>`.

packages/components/src/components/loading-indicator/loading-indicator.lite.tsx
nmerget: Fixed: the `<progress>` now has its own `id` (`{id}-progress`) and the `<label>` points to it via `for`/`htmlFor`, so the label is explicitly associated with the progress element. This renders as `htmlFor` in React and `for` in Vue/Angular.

nmerget:
@jonaskuske thanks for flagging the `autoDisable` foot gun. I've added the safeguard you suggested: on cleanup/unmount the indicator now force-enables its parent, so an indicator removed via conditional rendering (instead of `state="inactive"`) no longer leaves the parent (e.g. a `DBButton`) permanently disabled. Previously the unmount handler reapplied `disabled = _loadingState !== 'inactive'`, which kept it disabled — it now always resets to enabled on destroy.

On your other note about spacing/sizing and the timeout functionality: no worries, please do report those in detail once you're back and we'll address them separately.

packages/components/src/components/loading-indicator/loading-indicator.lite.tsx
```diff
@@ -0,0 +1,297 @@
...
+
+				if (!parent) return;
+
+				const isButton = parent?.localName === 'button';
+
+				if (!(isButton || props.overlay)) return;
+
+				const ariaAttribute = isButton
+					? 'aria-labelledby'
+					: 'aria-describedby';
```
michaelmkraus: `aria-labelledby` on a `<button>` **replaces** the button's accessible name, it does not add to it. With no pre-existing `aria-labelledby`, this sets `aria-labelledby="<indicator-id>"` and the button's own text is dropped from the name computation.

The ARIA snapshot shipped in this PR confirms it — `__snapshots__/loading-indicator/showcase/chromium/should-have-same-aria-snapshot/DBLoadingIndicator-should-have-same-aria-snapshot.yaml:257`:

    - button "Loading" [disabled]:
      - status:
        - text: Loading
        - progressbar "Loading"
      - text: Button

The button is named "Loading" while its visible text is "Button" — for all 8 button examples. In the timeout example the name flips from "Start Timeout" to "Loading" at runtime. A screen reader user can no longer tell *which* action is loading. That is WCAG 4.1.2 (Name, Role, Value), and the visible-text/name mismatch also breaks 2.5.3 (Label in Name).

Suggested fix: use `aria-describedby` for the button branch too, so the name stays "Button" and the loading state becomes a description. As a bonus, `aria-busy` is currently only set for non-button parents (line 128) — the button case is exactly where it carries the most value.

packages/components/src/components/loading-indicator/loading-indicator.scss
```diff
@@ -0,0 +1,537 @@
...
+				inset-inline-start: 50%;
+				transform: translate(-50%, -50%);
+			}
+		}
+	}
+
+	[data-show-label="false"],
+	[data-show-progress-text="false"] {
+		display: none;
+	}
```
michaelmkraus: `display: none` here removes the `<progress>` from the accessibility tree as well, because it is a child of this `<label>` (`loading-indicator.lite.tsx:268`). The `%a11y-visually-hidden` on `> progress` cannot rescue a `display: none` ancestor.

Confirmed by the ARIA snapshot in this PR (`...DBLoadingIndicator-should-have-same-aria-snapshot.yaml:169`):

    - text: "False"
    - status
    - status
    - status

Three empty live regions. With `showLabel={false}` nothing at all is announced — which is the exact opposite of what `showLabel={false}` is for (hide visually, keep it for AT). This is the same defect as the original `display: none` on `<progress>` finding, reintroduced through a different prop.

Suggested fix: use the established pattern from `packages/components/src/styles/internal/_form-components.scss:469-475` — `> label { @extend %a11y-visually-hidden; padding: 0; }`. Switching this component to `getHideProp(props.showLabel)` / `data-hide-label` (what checkbox, input, select, switch, radio and custom-select all use) would give you that for free and remove the second DOM contract.

packages/components/src/components/loading-indicator/loading-indicator.scss
```diff
@@ -0,0 +1,537 @@
...
+				--db-#{$name}-on-bg-basic-emphasis-70-default
+			);
+
+			&[data-variant="circular"][data-orientation="horizontal"] {
+				svg {
+					display: none;
+				}
+
+				div {
+					@include get-state-icon($name);
```
michaelmkraus, line 465: `successful` and `critical` are conveyed by a CSS `::before` icon and colour only. Pseudo-element icons are not in the accessibility tree, and the role stays `progressbar` / `status`, so the two states are indistinguishable for AT.

The ARIA snapshot in this PR produces byte-identical output for both (`...DBLoadingIndicator-should-have-same-aria-snapshot.yaml:138-157`):

    - text: Successful
    - status:
      - progressbar "Circular horizontal": 100 of 100
    - text: Critical
    - status:
      - progressbar "Circular horizontal": 100 of 100

Two things needed: expose the state as text (e.g. a visually hidden, localisable status string), and map the role from the state instead of hardcoding `status`. The design system already has that mapping in `packages/components/src/utils/index.ts:281-307` — `getNotificationRole` returns `alert` for `critical`/`warning` and `status` for `informational`/`successful`, and `DBNotification` uses it (`notification.lite.tsx:45`). This component uses the same semantic vocabulary, so `getRole()` (line 56 in the `.lite.tsx`) both diverges from that convention and duplicates a util that already exists. It also answers the open question in `docs/research/spinner.md` ("role='status' or role='alert'?").

packages/components/src/components/loading-indicator/loading-indicator.lite.tsx
```diff
@@ -0,0 +1,297 @@
...
+		state._loadingState,
+		props.overlay,
+		state._id
+	]);
+
+	onUpdate(() => {
+		if (props.onTimeout) {
+			if (state._loadingState === 'inactive') {
+				state._previousLoadingState = 'inactive';
+			}
```
michaelmkraus, line 356: The `inactive` branch resets `_previousLoadingState` but never clears a running timer. On `active → inactive` within the 5s window the timeout still fires and calls `props.onTimeout(state._loadingState)` — i.e. with `'inactive'`, for a load the consumer explicitly cancelled.

That contradicts the prop's own JSDoc (`model.ts:49-52`: "For 'active' after 5 seconds, for 'successful' and 'critical' after 2 seconds"), which does not include `inactive`. Reproduces in all four targets (React `:170`, Vue `:164`, Angular `:317`, Stencil `watch3Fn`).

Suggested fix: clear the timer here as well —

    if (state._loadingState === 'inactive') {
      if (state._timeoutId) {
        clearTimeout(state._timeoutId);
        state._timeoutId = undefined;
      }
      state._previousLoadingState = 'inactive';
    }

packages/components/src/components/loading-indicator/loading-indicator.lite.tsx
```diff
@@ -0,0 +1,297 @@
...
+				}
+
+				state._timeoutId = setTimeout(
+					() => {
+						state._timeoutId = undefined;
+						if (props.onTimeout) {
+							props.onTimeout(state._loadingState);
+						}
+					},
+					state._loadingState === 'active' ? 5000 : 2000
```
michaelmkraus: Magic numbers.

packages/components/src/components/loading-indicator/loading-indicator.lite.tsx
```diff
@@ -0,0 +1,297 @@
...
+
+	onUnMount(() => {
+		if (state._timeoutId) {
+			clearTimeout(state._timeoutId);
+			state._timeoutId = undefined;
+		}
+
+		state.handleParentAria(true);
+		state.handleParentDisabled(true);
+	});
```
michaelmkraus, line 461: Dead code in the React output. Mitosis emits this as `useEffect(() => { return () => { ... } }, [])` (`output/react/src/components/loading-indicator/loading-indicator.tsx:213-222`), so the cleanup closure captures the *first* render: `initialized === false`, `_timeoutId === undefined`, `_didDisableParent === false`.

Result in React: `handleParentAria(true)` and `handleParentDisabled(true)` both bail out at their `initialized` guard, and `clearTimeout` never runs. Nothing is restored. So the parent keeps a dangling `aria-labelledby` IDREF, `onTimeout` still fires after unmount, and **the button stays `disabled` forever** — which happens on the normal "request finished, remove the indicator" flow.

packages/components/scripts/post-build/components.ts
```diff
@@ -166,6 +166,15 @@ export const getComponents = (): Component[] => [
 		name: 'custom-button'
 	},
 
+	{
+		name: 'loading-indicator',
+		config: {
+			react: {
+				propsPassingFilter: ['autoDisable', 'onTimeout']
```
michaelmkraus: `role` is missing from this filter. It is listed in `reactHtmlAttributes`, so `filterPassingProps` forwards it onto the root `<div>` — while `role={state.getRole()}` also puts it on the inner `<div>`.

packages/components/src/components/loading-indicator/loading-indicator.lite.tsx
```diff
@@ -0,0 +1,297 @@
...
+							: '10 10 20 20'
+					}
+					aria-hidden="true">
+					<circle class="db-loading-indicator-circle-track" />
+					<circle class="db-loading-indicator-circle-segment" />
+				</svg>
+			</Show>
+
+			<div role={state.getRole()}>
+				<label
```
michaelmkraus: The `<label>` sits *inside* the `role="status"` region and is at the same time the accessible name of the `<progress>` via `for`. A live region reads its whole content, so the text is announced twice. 

Could the label move outside the live region, or be referenced via `aria-labelledby` without also being a text node inside it?

packages/components/src/components/loading-indicator/loading-indicator.lite.tsx
```diff
@@ -0,0 +1,297 @@
...
+			state._labelId = mId + DEFAULT_LABEL_ID_SUFFIX;
+			state._progressId = mId + DEFAULT_PROGRESS_ID_SUFFIX;
+		},
+		getPercentage: () => {
+			if (getBoolean(props.indeterminate) || !props.value || !props.max) {
+				return;
+			}
+
+			return `${Math.min(Math.max(Number(props.value) / Number(props.max), 0), 1).toFixed(2)}`;
+		},
```
michaelmkraus, line 90: Two issues with the guards here:

1. `!props.value` treats `value={0}` as "no value". The rendered result happens to be correct because `var(--db-loading-indicator-percentage, 0)` falls back to the same number, but the intent is unclear and it breaks the moment the fallback changes.
2. `value` is typed `string | number`, so `Number('abc')` → `NaN` → `"NaN"`. There is no `@property` registration for `--db-loading-indicator-percentage`, so `inline-size: calc(NaN * ...)` (line 280 in the SCSS) becomes invalid at computed-value time and `inline-size` falls back to `auto` — a typo in `value` renders as an apparently *full* progress bar.

packages/components/src/components/loading-indicator/loading-indicator.lite.tsx
```diff
@@ -0,0 +1,297 @@
...
+		_labelId: undefined,
+		_progressId: undefined,
+		_loadingState: 'inactive',
+		_previousLoadingState: undefined,
+		_timeoutId: undefined,
+		_didDisableParent: false,
+		_style: {},
+		initialized: false,
+		resetIds: () => {
+			const mId = props.id || 'loading-indicator-' + uuid();
```
michaelmkraus: Missing the `propOverrides?.id` fallback.

packages/components/src/components/loading-indicator/loading-indicator.lite.tsx
```diff
@@ -0,0 +1,297 @@
...
+					}
+					aria-hidden="true">
+					<circle class="db-loading-indicator-circle-track" />
+					<circle class="db-loading-indicator-circle-segment" />
+				</svg>
+			</Show>
+
+			<div role={state.getRole()}>
+				<label
+					data-show-label={getBooleanAsString(props.showLabel)}
```
michaelmkraus: Every other component with `showLabel` uses `getHideProp(props.showLabel)` → `data-hide-label` **on the root element**, styled via `&[data-hide-label="true"]`. This introduces a second DOM contract for the same shared prop, on a different element. Consumers who style around the label state now have to special-case this one component.

packages/components/src/components/loading-indicator/loading-indicator.scss
```diff
@@ -0,0 +1,537 @@
...
+	&[data-state="active"] {
+		&:not([data-indeterminate]),
+		&[data-indeterminate="true"] {
+			svg {
+				animation: rotate 1.5s linear 0s infinite normal none running;
+
+				@media screen and (prefers-reduced-motion: reduce) {
+					animation-duration: 5s;
+				}
+			}
```
michaelmkraus, line 127: Under reduced motion the spinner is only slowed to 5s, it never stops. For `DBCustomSelect` that is a regression: the rule this PR removes from `custom-select-dropdown.scss` was wrapped in `@media (prefers-reduced-motion: no-preference)`, i.e. reduced-motion users previously got **no** animation and now get a permanent 5s rotation.

packages/components/src/components/loading-indicator/loading-indicator.scss
```diff
@@ -0,0 +1,537 @@
...
+		animation: loading-indicator-delay 0s linear forwards;
+	}
+
+	&[data-delay="slow"] {
+		animation-delay: 500ms;
+	}
+
+	&[data-delay="fast"] {
+		animation-delay: 250ms;
+	}
```
michaelmkraus, line 621: Hardcoded durations, and inconsistent with the other consumer of the same shared `DelayProps`. `_popover-component.scss:187-193` uses tokens for the identical prop values:

    &[data-delay="slow"] { animation-delay: variables.$db-transition-duration-extra-slow; }  // 0.5s
    &[data-delay="fast"] { animation-delay: variables.$db-transition-duration-medium; }      // 0.3s

So `delay="fast"` means 300ms on tooltip/popover and 250ms here. Please use the tokens.

packages/components/src/components/loading-indicator/loading-indicator.scss
```diff
@@ -0,0 +1,537 @@
...
+
+		&[data-variant="bar"] {
+			div {
+				max-inline-size: calc(100% - #{variables.$db-spacing-fixed-md});
+			}
+		}
+	}
+}
+
+*:has(#{$overlay-child-selector}) {
```
michaelmkraus: A universal selector with `:has()` at global scope. This sets `position: relative` on *any* element in a consumer's document that has an overlay indicator as a direct child — including elements deliberately left `static`, which silently re-parents the consumer's own `position: absolute` children. It also adds style-recalc cost across the whole document.

packages/components/src/styles/internal/_button-components.scss
```diff
@@ -118,6 +120,8 @@
 	@if $selector == "&" {
 		&[data-variant="brand"] {
 			#{$selector} {
+				--db-loading-indicator-button-overlay-color: #{colors.$db-brand-origin-default};
```
michaelmkraus: This sits inside `@if $selector == "&"`, so it only applies to `DBButton`. `DBCustomButton[data-variant="brand"]` (which calls `set-basic-button($interactive-button-selectors)` and handles its brand colours separately in `custom-button.scss:39`) never gets this variable and falls back to `transparent` (`_loading.scss:66-69`) — two button components, two different overlay looks.

packages/components/configs/plugins/undefined-key.cjs
```diff
@@ -0,0 +1,22 @@
...
+ * so it is safe for the generated output it runs on.
+ *
+ * @type {import('@builder.io/mitosis').MitosisPlugin}
+ */
+module.exports = () => ({
+	code: {
+		post: (code) => {
+			return code
+				.replaceAll(':key="undefined"', '')
+				.replaceAll('key={undefined}', '');
```
michaelmkraus: This string replacement now runs over the entire Vue and Stencil output of all components to fix one component. Why does this component emit `key={undefined}` when the other components don't?

packages/components/src/components/loading-indicator/model.ts
```diff
@@ -0,0 +1,79 @@
...
+	'inactive',
+	'active',
+	'successful',
+	'critical'
+] as const;
+export type LoadingIndicatorStateType =
+	(typeof LoadingIndicatorStateList)[number];
+
+export type DBLoadingIndicatorDefaultProps = {
+	indeterminate?: boolean | string;
```
michaelmkraus, line 27: JSDoc is present on 3 of 12 props. Missing on `indeterminate`, `label`, `max`, `progressText`, `value`, `overlay`, `variant`, `state` and `role`. The conventions require it on every prop, and Patternhub renders these comments into the properties table — so the published docs page is incomplete.

packages/components/src/components/loading-indicator/loading-indicator.lite.tsx
nmerget: Fixed in b409ef75c48b. `handleParentAria` now always uses `aria-describedby` (never `aria-labelledby`), so on a button the loading state is added to the accessible name instead of replacing it — the button keeps announcing "Button". I also moved the `aria-busy` handling out of the `!isButton` branch so it is set for the button case too.

packages/components/src/components/loading-indicator/loading-indicator.scss
nmerget: Fixed in b409ef75c48b. The `<progress>` is now a sibling of the `<label>` (still associated via `id`/`htmlFor`) instead of a child, and the `%a11y-visually-hidden` helper moved with it (`div > progress`). Hiding the label therefore no longer removes the progress from the accessibility tree. Hiding the label is done via `label[data-show-label="false"]` using the same visually-hidden pattern rather than `display: none`.

packages/components/src/components/loading-indicator/loading-indicator.scss
nmerget: Addressed in b409ef75c48b. `getRole()` now maps the live-region role from the state via the existing `getNotificationRole` util (`critical` -> `alert`, otherwise `status`), instead of hardcoding `status` — so a critical result is announced assertively. This reuses the same mapping as `DBNotification` and answers the open `role="status"` vs `role="alert"` question from `docs/research/spinner.md`.

On the "expose the state as text" part: the consumer drives the state and passes the matching `label`/children themselves (e.g. `state="critical"` with their own localized "Error" text), and that label renders inside the live region, so the state text is already in the accessibility tree. Adding dedicated `successfulText`/`criticalText` props would just duplicate a string the consumer already provides, so I left that out. Happy to reconsider if you'd prefer the component to own default state strings.

packages/components/src/components/loading-indicator/loading-indicator.lite.tsx
nmerget: Fixed in b409ef75c48b. The `inactive` branch now clears any running timer before resetting `_previousLoadingState`, so an `active -> inactive` cancellation no longer fires `onTimeout` with `'inactive'`.

packages/components/src/components/loading-indicator/loading-indicator.lite.tsx
nmerget: Fixed in b409ef75c48b. The timeout defaults are now named constants (`DEFAULT_LOADING_TIMEOUT_ACTIVE` = 5000, `DEFAULT_LOADING_TIMEOUT` = 2000) in `shared/constants.ts`, used in the `setTimeout`.

packages/components/src/components/loading-indicator/loading-indicator.lite.tsx
nmerget: Fixed in b409ef75c48b. `onUnMount` now calls a dedicated `handleUnmount` that reads everything from the live `_ref`/DOM instead of captured state, so the stale first-render React closure no longer breaks it:
- the timer handle is mirrored onto `dataset.timeoutId` and cleared from there, so a pending `onTimeout` no longer fires after unmount;
- the aria id is removed from the parent by reading the live `root.id` (and `aria-busy` is cleared);
- the "we disabled the parent" flag is mirrored onto `dataset.didDisableParent`, so the parent is re-enabled on unmount and the button no longer stays `disabled` forever.

packages/components/src/components/loading-indicator/loading-indicator.lite.tsx
nmerget: Fixed in 23db8c497b24. `getPercentage` now checks `value`/`max` against `undefined`/`null` explicitly (so `value={0}` is treated as a real value), and bails out when `Number(value)`/`Number(max)` is `NaN` or `max <= 0`. A non-numeric `value` therefore returns no percentage instead of producing `"NaN"` and an invalid `calc()`.

packages/components/src/components/loading-indicator/loading-indicator.lite.tsx
nmerget: Fixed in 23db8c497b24. `resetIds` now uses `props.id || props.propOverrides?.id || 'loading-indicator-' + uuid()`, so the `propOverrides?.id` fallback is honored.

packages/components/src/components/loading-indicator/loading-indicator.scss
nmerget: Fixed in 23db8c497b24. The delay `animation-delay` now uses the shared tokens: `slow` -> `$db-transition-duration-extra-slow` (0.5s) and `fast` -> `$db-transition-duration-medium` (0.3s), matching tooltip/popover. Note this changes `fast` from 250ms to 300ms.

packages/components/src/styles/internal/_button-components.scss
nmerget: Fixed in 23db8c497b24. Moved `--db-loading-indicator-button-overlay-color` out of the `@if $selector == "&"` guard so it is set for both `DBButton` and `DBCustomButton[data-variant="brand"]`. The `%button-brand-colors` `@extend` stays gated to `DBButton` (DBCustomButton applies its brand colours itself, since `@extend` does not compose through the passed-in complex selector).

packages/components/src/components/loading-indicator/model.ts
nmerget: Fixed in 23db8c497b24. Added JSDoc to all remaining props (`indeterminate`, `label`, `max`, `progressText`, `value`, `overlay`, `variant`, `state`, `role`) so the Patternhub properties table is complete.

packages/components/scripts/post-build/components.ts
nmerget: The generated root `<div>` for this component does not spread props via `filterPassingProps`/`getRootProps` at all — it only sets `ref`, `id`, `className`, `style` and the explicit `data-*` attributes. So `props.role` is never forwarded onto the root; `role` only appears once, on the inner live-region `<div>` via `role={getRole()}`. Adding `role` to `propsPassingFilter` therefore has no effect, so I have not changed it. Let me know if you're seeing `role` on the root in a specific output and I'll dig further.

packages/components/configs/plugins/undefined-key.cjs
nmerget: The `key={undefined}` came from the `<progress>` child being written as a ternary that yields `undefined` (`{indeterminate ? undefined : progressText}`) rather than a `<Show>`. In 23db8c497b24 I changed it to `<Show when={!indeterminate}>{progressText}</Show>`, which removes the bare-`undefined` child (the Vue output no longer emits `{{ undefined }}` there). I kept the `undefined-key` plugin for now rather than removing it, since I couldn't fully confirm no other element still triggers it. If we can confirm the `<Show>` change removes the last source, the plugin can be dropped in a follow-up.

packages/components/src/components/loading-indicator/loading-indicator.lite.tsx
nmerget: Acknowledged. The `data-show-label` on the label (instead of `getHideProp` -> `data-hide-label` on the root) was a deliberate choice here because the `<progress>` was moved out to be a sibling of the label, so hiding is scoped to the label element itself. I take the point about the shared DOM contract for consumers styling around the label state — leaving this open for a decision on whether to switch back to the root `data-hide-label` convention for consistency across components.

packages/components/src/components/loading-indicator/loading-indicator.scss
nmerget: Not changed yet — this is a deliberate design call I did not want to make unilaterally. The current behaviour slows the spinner to 5s under reduced motion rather than stopping it, which is a regression from the prior `DBCustomSelect` behaviour (no animation under reduced motion) as you note. Options are (a) stop the animation entirely under `prefers-reduced-motion: reduce` (matches the old DBCustomSelect, but a frozen spinner can read as broken), or (b) keep a very slow spin. Which do you prefer? Same question applies to the bar `wobbling` animation.

packages/components/src/components/loading-indicator/loading-indicator.scss
nmerget: Valid concern about the global `*:has()`. Not changed yet because the universal selector is currently what makes the "drop an overlay indicator into any container" contract work (it turns the parent into a positioning context). Scoping it without breaking that contract needs a decision — e.g. requiring the overlay parent to opt in with a class, or limiting the left-hand side to a known set of containers. Happy to implement whichever direction you prefer.

packages/components/src/components/loading-indicator/loading-indicator.lite.tsx
nmerget: Fixed in ba73d495c502. `showLabel` now uses `getHideProp(props.showLabel)` -> `data-hide-label` on the root element, styled via `&[data-hide-label="true"] > div > label`, matching the shared convention used by checkbox/input/select/switch/radio/custom-select. The `data-show-label` attribute on the label is gone, so there is no longer a second DOM contract for consumers to special-case.

We kept `data-show-progress-text` on the aria-hidden `<span>` as is: `showProgressText` is a loading-indicator-specific prop (not the shared `ShowLabelProps`), so there is no cross-component root convention to align it with.

packages/components/src/components/loading-indicator/loading-indicator.scss
nmerget: Fixed in ba73d495c502. Under `prefers-reduced-motion: reduce` the spinner (and the bar segment) no longer rotate continuously. Instead of slowing to a permanent 5s rotation, they now advance in discrete steps via `animation-timing-function: steps(8, jump-end)` with a 16s duration, i.e. the indicator jumps to the next position once every 2s. That removes the continuous motion the `DBCustomSelect` regression was about, while still visibly updating so it does not look frozen/broken.

packages/components/configs/plugins/undefined-key.cjs
nmerget: Removed the plugin in ba73d495c502. The `key={undefined}` came from the `<progress>` child being a ternary that yields `undefined`; that was already changed to a `<Show>` in 23db8c497b24, so the plugin is no longer needed. I deleted `undefined-key.cjs` and its registration from the vue and stencil configs.

packages/components/src/components/loading-indicator/loading-indicator.lite.tsx
nmerget: After thinking this through I decided to keep the `<label>` inside the `role` region rather than move it out. The reason is the state transitions: when the indicator goes `active -> successful`/`critical`, the consumer typically changes the label too ("Loading" -> "Saved"/"Error"). Because the live region re-reads its content on update, keeping the label inside is what lets that new label be announced. If we moved the label outside the region, the region would only re-announce the progressbar value on that transition and the user would lose the context of *what* finished.

So the tradeoff is: keeping the label in the region causes the label text to be announced twice on the initial render (once as region text, once as the progressbar's `for` name), but it ensures the label is announced on state changes, which we consider the more important behaviour. We're treating the initial duplication as an accepted minor verbosity rather than dropping the state-change announcement. Open to revisiting if you have a way to keep the state-change announcement without the duplication.

packages/components/src/components/loading-indicator/loading-indicator.scss
nmerget: Fixed in cd8bd614d1c6. Replaced the universal `*:has(...)` with a scoped subject:

```scss
:is(.db-button, .db-custom-button, .db-card, .db-popover, [data-loading-overlay="true"]):has(#{$overlay-child-selector})
```

So `position: relative` is only applied to the known overlay hosts (both button components, card, popover), which removes the silent re-parenting of arbitrary `static` consumer elements. Consumers using their own container opt in explicitly via `[data-loading-overlay="true"]`. I included `.db-custom-button` alongside `.db-button` since both pull in `%interactive-element-with-loading` via `set-basic-button` and are equivalent overlay hosts.

mfranzke:
@codex review

packages/components/src/components/loading-indicator/loading-indicator.lite.tsx
```diff
@@ -0,0 +1,430 @@
...
+			state._labelId = mId + DEFAULT_LABEL_ID_SUFFIX;
+			state._progressId = mId + DEFAULT_PROGRESS_ID_SUFFIX;
+		},
+		getPercentage: () => {
+			if (
+				getBoolean(props.indeterminate) ||
+				props.value === undefined ||
+				props.value === null ||
+				props.max === undefined ||
+				props.max === null
```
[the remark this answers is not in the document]
nmerget: Fixed in 85117e53c70f. `getPercentage` now defaults `max` to 100 when it is omitted (matching the native `<progress>` default and the documented default), so a determinate `value` without an explicit `max` produces a percentage instead of nothing.

packages/components/src/components/loading-indicator/loading-indicator.lite.tsx
```diff
@@ -0,0 +1,430 @@
...
+					class="db-loading-indicator-circle"
+					viewBox={
+						props.orientation === 'vertical'
+							? '32 32 64 64'
+							: '10 10 20 20'
+					}
+					aria-hidden="true">
+					<circle class="db-loading-indicator-circle-track" />
+					<circle
+						style={state._segmentStyle}
```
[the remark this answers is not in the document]
nmerget: Fixed in 85117e53c70f. The `--db-loading-indicator-percentage` custom property is no longer only on the circular `<circle>`. It is now applied to the inner content `<div>` (the bar container, whose `::after` reads the property) as well as the circular segment, so the `bar` variant shows determinate progress. It is intentionally not set on the root `_ref` div.

packages/components/src/styles/internal/_loading.scss
```diff
@@ -0,0 +1,91 @@
...
+		}
+
+		span {
+			display: none;
+		}
+	}
+}
+
+%interactive-element-with-loading {
+	&:has(.db-loading-indicator:not([data-overlay="true"])) {
```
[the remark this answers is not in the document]
nmerget: Fixed in 85117e53c70f. The icon-hiding selector in `_loading.scss` now excludes inactive indicators (`:not([data-overlay="true"], [data-state="inactive"])`), so a controlled flow that flips the indicator to `inactive` keeps the host button icon visible instead of leaving an icon-only button blank.

packages/components/src/components/loading-indicator/loading-indicator.lite.tsx
```diff
@@ -0,0 +1,430 @@
...
+			// active/inactive/successful stay a polite "status" region.
+			return getNotificationRole({
+				semantic:
+					state._loadingState === 'critical'
+						? 'critical'
+						: 'successful'
+			});
+		},
+		handleParentDisabled: (forceEnable?: boolean) => {
+			if (_ref && getBoolean(props.autoDisable) && state.initialized) {
```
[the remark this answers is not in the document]
nmerget: Fixed in 85117e53c70f. `handleParentDisabled` no longer gates the whole method on `props.autoDisable`. The disable path is still gated on `autoDisable`, but the re-enable path now also runs when `autoDisable` is turned off (`forceEnable || !autoDisable || state === "inactive"`), as long as this instance was the one that disabled the parent (`_didDisableParent`). So turning off auto-disable re-enables a button this indicator had disabled.

packages/components/scripts/post-build/components.ts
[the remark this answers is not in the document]
nmerget: Fixed in 85117e53c70f. `role` was added to this component's React `propsPassingFilter`, so `filterPassingProps` no longer forwards a consumer-provided `role` onto the `_ref` root. `role` now only lands on the inner live region via `role={getRole()}`. Verified in the generated React output and in the browser with a new `role` example: without the filter the root `.db-loading-indicator` carried the duplicate `role`, with it the root has none.

packages/components/src/components/loading-indicator/loading-indicator.lite.tsx
```diff
@@ -0,0 +1,430 @@
...
+	}, [
+		_ref,
+		state.initialized,
+		state._loadingState,
+		props.overlay,
+		state._id
+	]);
+
+	onUpdate(() => {
+		if (props.onTimeout) {
```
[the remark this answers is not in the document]
nmerget: Fixed in 85117e53c70f. The timeout effect now has an `else` branch: when `props.onTimeout` becomes falsy while a timer is pending, it clears `state._timeoutId` and removes the `dataset.timeoutId` marker, so a removed handler can no longer fire later.

packages/components/src/components/loading-indicator/loading-indicator.lite.tsx
```diff
@@ -0,0 +1,430 @@
...
+
+		if (props.state) {
+			state._loadingState = props.state;
+		} else if (
+			props.indeterminate === undefined ||
+			getBoolean(props.indeterminate)
+		) {
+			state._loadingState = 'active';
+		} else {
+			state._loadingState = 'inactive';
```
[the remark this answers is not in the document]
nmerget: Fixed in 85117e53c70f. When no explicit `state` is provided the loading state now defaults to `active` for both indeterminate and determinate (`indeterminate={false}` with `value`/`max`) modes, instead of `inactive`. So the documented determinate API renders progress without the consumer also having to pass `state="active"`.

packages/components/src/components/loading-indicator/loading-indicator.lite.tsx
```diff
@@ -0,0 +1,430 @@
...
+				let parent = (_ref as HTMLDivElement).parentElement;
+				if (parent && parent.localName === 'db-loading-indicator') {
+					parent = parent.parentElement;
+				}
+
+				if (!parent) return;
+
+				const isButton = parent?.localName === 'button';
+
+				if (!(isButton || props.overlay)) return;
```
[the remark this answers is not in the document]
nmerget: Fixed in 85117e53c70f. `handleParentAria` no longer returns early when the element is neither a button nor an overlay. It now computes `shouldRemove = remove || !(isButton || overlay)`, so when `overlay` is turned off it removes the previously-added `aria-describedby` token and clears `aria-busy` instead of leaving them on the container.

packages/components/src/components/loading-indicator/loading-indicator.lite.tsx
```diff
@@ -0,0 +1,430 @@
...
+
+	onMount(() => {
+		state.resetIds();
+		state.initialized = true;
+	});
+
+	onUpdate(() => {
+		if (props.id) {
+			state.resetIds();
+		}
```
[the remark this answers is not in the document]
nmerget: Fixed in 85117e53c70f. The `onUpdate` keyed on `props.id` now calls `handleParentAria(true)` before `resetIds()`, so the parent relationship referencing the current (old) `_id` is removed before the new id is generated. That avoids leaving a dangling `aria-describedby` token that the later add/remove for the new id could never clean up.

packages/components/src/components/loading-indicator/loading-indicator.scss
```diff
@@ -0,0 +1,598 @@
...
+	}
+}
+
+$progress-bar-full-segment-inline-size: calc(
+	100% - var(--db-loading-indicator-progress-bar-segment-padding) * 2
+);
+
+%progress-bar {
+	inline-size: 100%;
+	min-inline-size: variables.$db-sizing-xl;
```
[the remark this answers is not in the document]
nmerget: Addressed in 85117e53c70f by removing the `width` prop instead of implementing variants. `WidthProps` is outdated for this component, so it was dropped from the model, the `data-width` attribute from the template and the `width` control from the Storybook arg types. The bar keeps its `inline-size: 100%` and there is no longer an advertised prop that has no effect.

packages/components/src/components/loading-indicator/docs/HTML.md
```diff
@@ -0,0 +1,47 @@
...
+			<circle class="db-loading-indicator-circle-track"></circle>
+			<circle class="db-loading-indicator-circle-segment"></circle>
+		</svg>
+		<div role="status">
+			<label
+				id="loading-indicator-1-label"
+				for="loading-indicator-1-progress"
+			>
+				Loading
+				<progress
```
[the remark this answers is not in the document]
nmerget: Fixed in 85117e53c70f. The HTML example now closes the `<label>` before the `<progress>` so the native control is a sibling (matching the framework template and the `div > progress` selector that visually hides it), instead of nesting it inside the label where it stayed visible.

packages/components/src/components/loading-indicator/docs/HTML.md
```diff
@@ -0,0 +1,47 @@
...
+
+```html index.html
+<!-- index.html -->
+...
+<body>
+	<div
+		class="db-loading-indicator"
+		data-variant="circular"
+		data-orientation="horizontal"
+	>
```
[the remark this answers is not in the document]
nmerget: Fixed in 85117e53c70f. The determinate HTML example now sets `data-indeterminate="false"` and `data-state="active"` on the root and provides `style="--db-loading-indicator-percentage: 0.42;"`, since CSS-only consumers have no JS to derive those. The example now renders 42% progress instead of the continuous spinner.

packages/foundations/scss/animation/_animations.scss
```diff
@@ -23,3 +23,17 @@
...
 }
+
+@keyframes wobbling {
+	50% {
+		inset-inline-start: 100%;
+		transform: translateX(
+			calc(
+				-100% - var(
+						--db-loading-indicator-progress-bar-segment-padding,
+						2px
```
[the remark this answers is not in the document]
nmerget: Fixed in 85117e53c70f. The `wobbling` keyframe now multiplies the physical `translateX` by `--db-loading-indicator-wobbling-direction` (default `1`), and the component sets that variable to `-1` under `:dir(rtl)`. So under RTL the translate flips to match the direction-aware `inset-inline-start`, keeping the segment inside the track instead of overshooting.

packages/components/src/components/loading-indicator/loading-indicator.lite.tsx
```diff
@@ -0,0 +1,430 @@
...
+					<circle class="db-loading-indicator-circle-track" />
+					<circle
+						style={state._segmentStyle}
+						class="db-loading-indicator-circle-segment"
+					/>
+				</svg>
+			</Show>
+
+			<div role={state.getRole()}>
+				<label id={state._labelId} htmlFor={state._progressId}>
```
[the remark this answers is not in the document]
nmerget: The component uses `aria-describedby` (not `aria-labelledby`) to reference the indicator from the host button, so the loading state is added as a *description* and does not replace the button's accessible name. The name is still computed from the button's own content, so no separate hiding of the inner region is needed. Note the ARIA snapshot referenced here predates the switch to `aria-describedby`; the snapshots are regenerated in CI on this branch.

packages/components/src/components/loading-indicator/docs/HTML.md
```diff
@@ -0,0 +1,56 @@
...
+		<svg
+			class="db-loading-indicator-circle"
+			viewBox="10 10 20 20"
+			aria-hidden="true"
+		>
+			<circle class="db-loading-indicator-circle-track"></circle>
+			<circle class="db-loading-indicator-circle-segment"></circle>
+		</svg>
+		<div role="status">
```
michaelmkraus: This still documents the nested live region that the changeset says was removed: the `role` now sits on the component root, not on this inner `<div>`. CSS-only consumers following this example end up with a root that is not a live region plus a nested `status`. Please move `role="status"` up to the `.db-loading-indicator` root and update the intro sentence.

packages/components/configs/plugins/attribute-passing/index.cjs
```diff
@@ -114,6 +114,10 @@ module.exports = () => ({
 						// Pass all `data-` and `aria-` attributes, except `data-density`
 						"attr && attr.name !== 'data-density' &&"
 					)
+					.replace(
+						'attr.name.startsWith("aria-")',
+						'attr.name.startsWith("aria-") || attr.name === "role"'
```
michaelmkraus, line 127: This changes attribute passing for every Angular and Stencil component, not just this one — a `role` on the host is now moved to the inner element and `removeAttribute`d from the host. `role` is a declared prop on both `db-loading-indicator` and `db-notification`, so in Stencil the `removeAttribute` triggers the attribute→prop sync, clears the prop, and the next render overwrites the moved value with the state-derived default. Can you verify this in the generated Stencil output and add coverage?

packages/components/src/components/loading-indicator/loading-indicator.lite.tsx
```diff
@@ -0,0 +1,477 @@
...
+
+export default function DBLoadingIndicator(props: DBLoadingIndicatorProps) {
+	// This is used as forwardRef
+	const _ref = useRef<HTMLDivElement | any>(undefined);
+	// jscpd:ignore-start
+	const state = useStore<DBLoadingIndicatorState>({
+		_id: DEFAULT_ID,
+		_labelId: undefined,
+		_progressId: undefined,
+		_loadingState: 'inactive',
```
michaelmkraus, line 46: The initial `'inactive'` means the first render emits `data-state="inactive"` which hides the circular segment, zeroes the bar, and `display: none`s the entire indicator inside a button. So a loading indicator is invisible until hydration, and flashes on CSR. Suggest initializing to `'active'` and rendering `data-state={props.state ?? state._loadingState}` (line 426), so both cases are correct from first paint.

packages/components/src/components/loading-indicator/loading-indicator.lite.tsx
```diff
@@ -0,0 +1,477 @@
...
+					// when autoDisable itself was turned off -- but only if this
+					// instance was the one that disabled the parent.
+					if (
+						(forceEnable ||
+							!autoDisable ||
+							state._loadingState === 'inactive') &&
+						state._didDisableParent
+					) {
+						state._didDisableParent = false;
+						(_ref as HTMLDivElement).dataset['didDisableParent'] =
```
michaelmkraus: The stale-closure reason is clear, but `data-did-disable-parent` is set to `'false'` instead of being deleted, so it stays in the consumer's DOM forever — inconsistent with delete `dataset['timeoutId']` a few lines down.

packages/components/src/components/loading-indicator/loading-indicator.scss
```diff
@@ -0,0 +1,629 @@
...
+	&[data-state="critical"] {
+		div {
+			&::after {
+				/* stylelint-disable-next-line db-ux/use-sizing */
+				inline-size: $progress-bar-full-segment-inline-size;
+			}
+		}
+	}
+
+	div {
```
michaelmkraus: Descendant `div` instead of `> div`: `props.children` render inside the `<label>`, so a consumer `<div>` in the label gets `position: relative` and a second `::before`/`::after` progress track. Same for `%text-container` at line 27 (`display: flex`).

mfranzke:
@codex review

mfranzke:
@codex review

packages/components/src/components/loading-indicator/docs/HTML.md
nmerget: Fixed. The HTML example now puts `role="status"` on the `.db-loading-indicator` root and the inner wrapper is a plain `<div class="db-loading-indicator-content">` with no role, so CSS-only consumers get a single live region on the root, matching the framework output. The intro sentence was updated accordingly (role sits on the root, not the inner wrapper).

packages/components/src/components/loading-indicator/loading-indicator.lite.tsx
nmerget: Fixed, but via `onMount` rather than changing the initial value. `_loadingState` stays `inactive` as the declared initial state, and the template still reads `data-state={state._loadingState}`. `onMount` now resolves it once with `state._loadingState = props.state ?? "active"`, so the indicator is correct from the first paint (visible when uncontrolled, or the consumer's `props.state` when controlled) without hardcoding `active` into the initial state or adding a `props.state ??` fallback in the template.

packages/components/src/components/loading-indicator/loading-indicator.lite.tsx
nmerget: Fixed. The re-enable path now `delete`s `dataset["didDisableParent"]` instead of setting it to `"false"`, so it no longer lingers in the consumer DOM (consistent with `delete dataset["timeoutId"]` a few lines down).

packages/components/src/components/loading-indicator/loading-indicator.scss
nmerget: Fixed. The inner content element now carries a dedicated `db-loading-indicator-content` class and the styles target `> .db-loading-indicator-content` (in `%text-container`, `%progress-bar`, `%spinner` and `%states`) instead of a bare descendant `div`/`> div`. A consumer `<div>` in the projected children (which render inside the `<label>`) therefore no longer matches the layout rules and no longer gains `display: flex` / `position: relative` / a second `::before`/`::after` progress track. The HTML example and its prose were updated to include the class.

packages/components/configs/plugins/attribute-passing/index.cjs
nmerget: Kept the general `role` forwarding (added an explanatory comment above it). `role` is a global attribute: a consumer can set it on any custom-element host, and since the host is a `display: contents` wrapper the attribute has to move onto the real inner element, exactly like `data-*`/`aria-*`. Restricting it to individual components would leave `role` stranded on the contents-wrapper everywhere else. Overwriting a component's own inner role (e.g. `DBTabItem` `role="tab"`) when a consumer explicitly sets `role` on the host is intended — the consumer owns the global attribute.

On the Stencil attribute->prop point: for `db-loading-indicator`/`db-notification` the consumer-facing path is the declared `role` prop, which is rendered on the correct inner element via `getRole()`/`getNotificationRole()`, so the intended value is not lost. If you'd still prefer the observer to skip `role` for components that declare it as a prop (to avoid the redundant move + `removeAttribute` churn), I'm happy to add that narrowing — let me know.

packages/components/src/components/loading-indicator/loading-indicator.lite.tsx
```diff
@@ -0,0 +1,477 @@
...
+					(_ref as HTMLDivElement).dataset['timeoutId'] =
+						`${Number(timeoutId)}`;
+				}
+			}
+		} else if (state._timeoutId) {
+			// onTimeout was removed while a timer was still pending: this effect
+			// reruns (onTimeout is a dependency) but would otherwise leave the
+			// old timer and its captured callback alive, so clear it here.
+			clearTimeout(state._timeoutId);
+			state._timeoutId = undefined;
```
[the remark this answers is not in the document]
nmerget: Fixed. The `onTimeout`-removed branch now also resets `state._previousLoadingState = undefined` after clearing the timer, so re-adding the callback while the state is unchanged no longer hits the state-equality guard and a replacement timer is scheduled again.

packages/components/src/components/loading-indicator/loading-indicator.lite.tsx
```diff
@@ -0,0 +1,477 @@
...
+				if (describedByElements.length) {
+					parent.setAttribute(
+						'aria-describedby',
+						describedByElements.join(' ')
+					);
+				} else {
+					parent.removeAttribute('aria-describedby');
+				}
+
+				parent.ariaBusy = null;
```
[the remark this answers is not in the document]
nmerget: Fixed. The component now tracks whether it set `aria-busy` (`_didSetAriaBusy`, mirrored onto `dataset["didSetAriaBusy"]` so it survives the React unmount closure). `handleParentAria` only clears `aria-busy` when this instance had set it, and `handleUnmount` only clears it when `dataset["didSetAriaBusy"] === "true"`. So a parent already carrying `aria-busy="true"` for unrelated/concurrent work is no longer reset to ready by this indicator.

packages/components/src/components/loading-indicator/loading-indicator.lite.tsx
```diff
@@ -0,0 +1,477 @@
...
+			}
+		},
+		// Runs on unmount. Mitosis compiles onUnMount to a React
+		// useEffect(() => cleanup, []) whose closure captures the first-render
+		// state (initialized: false, _timeoutId/_didDisableParent unset), so
+		// the normal handlers would all bail out and never restore the parent.
+		// This cleanup therefore reads everything from the live _ref/DOM
+		// instead of from captured state.
+		handleUnmount: () => {
+			if (!_ref) return;
```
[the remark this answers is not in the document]
nmerget: Not changed in this round. The cleanup still reads the node from `_ref` at unmount, which is the same pattern every other component here uses for unmount cleanup (`tabs`, `tab-item`, `drawer-header` all read `_ref`/state in their `onUnMount`). A robust cross-framework fix for the React "object ref nulled before the passive-effect cleanup" case is not expressible cleanly in Mitosis source: a second `useRef` gets the ref-name transform applied to it (`_nodeRef.current.current = _ref.current`) and a state field is stale in the `[]` cleanup — both verified in the generated output. Doing this properly belongs in a Mitosis plugin that rewrites the React `onUnMount` to capture the node in the effect body, which is a repo-wide change rather than something specific to this component. Flagging it so we can decide whether to tackle that separately.

packages/components/src/components/loading-indicator/loading-indicator.lite.tsx
```diff
@@ -0,0 +1,477 @@
...
+
+	// jscpd:ignore-end
+
+	onMount(() => {
+		state.resetIds();
+		state.initialized = true;
+	});
+
+	onUpdate(() => {
+		if (props.id) {
```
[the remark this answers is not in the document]
nmerget: Left as-is. The `onUpdate` keyed on `props.id` intentionally only runs when a `props.id` is actually provided (`if (props.id)`). Clearing a previously-set controlled `id` back to `undefined`/empty is an uncommon flow, and reacting to it would regenerate a uuid-based id on removal, which is a behavior change we did not want to introduce here. Happy to revisit if this turns out to bite a real consumer.

packages/components/src/components/loading-indicator/loading-indicator.lite.tsx
```diff
@@ -0,0 +1,477 @@
...
+					: Number(props.max);
+
+			// Guard against non-numeric props (e.g. Number("abc") -> NaN) and a
+			// zero/negative max, which would otherwise produce "NaN" and render
+			// as an invalid calc() that falls back to a full bar.
+			if (Number.isNaN(value) || Number.isNaN(max) || max <= 0) {
+				return;
+			}
+
+			return `${Math.min(Math.max(value / max, 0), 1).toFixed(2)}`;
```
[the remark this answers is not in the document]
nmerget: Left the `toFixed(2)` rounding as-is. The flattening only becomes visible at sub-0.5% resolution (e.g. `value=3, max=1000`), which is not a realistic use of a progress value, and at 2-decimal precision the visual difference is imperceptible. `data-percentage-full` continues to key off the rounded `"1.00"`; a value rounding up to full is close enough to completion for the visualization that we are comfortable treating it as full.

packages/components/configs/plugins/attribute-passing/index.cjs
```diff
@@ -114,6 +114,10 @@ module.exports = () => ({
...
+					)
```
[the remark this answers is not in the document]
nmerget: Intended behavior, so not scoped. `role` is a global attribute and forwarding it off the `display: contents` host onto the real inner element is correct for every component (same rationale as `data-*`/`aria-*`). If a consumer deliberately sets `role` on a fixed-role widget's host (e.g. `DBTabItem`), overriding the component's own `role="tab"` is the consumer exercising ownership of that global attribute, not a bug. Added a comment in the plugin documenting this.

packages/components/src/components/loading-indicator/loading-indicator.scss
```diff
@@ -0,0 +1,629 @@
...
+					}
+				}
+			}
+
+			&[data-variant="bar"],
+			&[data-variant="circular"][data-orientation="vertical"] {
+				div {
+					> label {
+						@include get-state-icon($name);
+					}
```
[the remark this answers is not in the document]
nmerget: Not changed. Moving the terminal-state icon from the `<label>` to the content wrapper shifts where the icon pseudo-element sits in the flex/grid flow, which changes the layout for the normal (visible-label) case — the common case — to fix a color-only edge case that only happens when the label is hidden. We decided the layout regression is not worth it, so the icon stays on the `<label>` (bar and vertical-circular unchanged). Open to revisiting the hidden-label a11y aspect separately if needed.

.changeset/loading-indicator-review-fixes.md
mfranzke, line 1: this component hasn't been released, why does it need another changeset for ongoing changes/fixes?

.changeset/loading-indicator-review-fixes-2.md
mfranzke, line 1: this component hasn't been released, why does it need another changeset for ongoing changes/fixes?

__snapshots__/custom-select/showcase/chromium/should-have-same-aria-snapshot/DBCustomSelect-should-have-same-aria-snapshot.yaml
```diff
@@ -962,7 +962,10 @@
   - status
   - text: "True"
   - group:
-    - article: Loading
+    - article:
+      - status:
+        - text: Loading
+        - progressbar "Loading"
```
mfranzke, line 968: the text and the semantic element with the accessible name `Loading` are redundant, the first one would be the one being obsolete here.

.changeset/loading-indicator-review-fixes.md
nmerget: Agreed. Removed in 14dcced465b9. Since `DBLoadingIndicator` is not released yet, all the review fixes ship under the original feature changeset (`plenty-jobs-fetch.md`, `minor`). A separate `patch` changeset just adds noise to a changelog that has no prior published version to amend, so I deleted both `loading-indicator-review-fixes.md` and `loading-indicator-review-fixes-2.md`.

.changeset/loading-indicator-review-fixes-2.md
nmerget: Same as above - removed in 14dcced465b9. Both `loading-indicator-review-fixes.md` and `loading-indicator-review-fixes-2.md` are gone; the unreleased component keeps just the single feature changeset.

__snapshots__/custom-select/showcase/chromium/should-have-same-aria-snapshot/DBCustomSelect-should-have-same-aria-snapshot.yaml
nmerget: This is the standard way a `<label for>` association serializes in the aria snapshot, not an actual double announcement. The `text` node is the `<label>`, and the `progressbar "Loading"` is the native `<progress>` it names via `for`/`id` - so the label text shows up both as its own node and as the accessible name of the element it labels.

It is the same pattern `DBSelect`/`DBCustomSelect` themselves produce:

```yaml
- text: Label
- combobox "Label":
```

Nobody flags that on the select, because to a screen reader the label and the control it names are announced together as one thing. It resolves to a single announcement here too, so I have left the markup as is - the `<label for>`/`<progress id>` association is intentional and correct.

packages/components/src/components/loading-indicator/loading-indicator.lite.tsx
```diff
@@ -0,0 +1,514 @@
...
+				parent.disabled = false;
+			}
+		}
+	});
+
+	// jscpd:ignore-end
+
+	onMount(() => {
+		state.resetIds();
+		state._loadingState = props.state ?? 'active';
```
michaelkraus, line 315: This doesn't fix the reported [problem](https://github.com/db-ux-design-system/core-web/pull/5992#discussion_r3988885506). `onMount` compiles to `useEffect(() => …, [])` and runs *after* the first render, while `_loadingState` still starts as `'inactive'`(line 46) and the template still binds `data-state={state._loadingState}` (line 463). So the first paint and the SSR markup carry `data-state="inactive"` — inside a button that's `display: none` until hydration, standalone the segment is hidden / the bar sits at 0.

Initializing to `'active'` and rendering
`data-state={props.state ?? state._loadingState}` fixes it in the markup
instead of in an effect.

packages/components/src/styles/internal/_loading.scss
```diff
@@ -0,0 +1,107 @@
...
+		 * aria-hidden progress-text span is fully removed.
+		 */
+		label {
+			@extend %a11y-visually-hidden;
+		}
+
+		span {
+			display: none;
+		}
+	}
```
michaelmkraus: The `.db-loading-indicator-content` fix didn't reach this file: these are still bare descendant selectors. Since `props.children` render inside the `<label>`, `div span { display: none }` also matches consumer markup —
`<DBButton><DBLoadingIndicator><span>Saving…</span></DBLoadingIndicator>Save</DBButton>`
drops the loading text from the accessibility tree, which is the opposite of
what the comment above promises, and leaves the button's `aria-describedby` pointing at an empty description. Applies to `[data-overlay="true"]` too via the `@extend`.

Please target `> .db-loading-indicator-content > label` / `> span` here as
well. Same for `svg, div { z-index: 2 }` at line 99.

nmerget: Fixed in beafab7be688. `_loading.scss` no longer uses bare descendant selectors: `%hide-label-and-progress-text` now targets `> .db-loading-indicator-content > label` / `> span`, and the overlay `z-index` rule targets `> svg` / `> .db-loading-indicator-content`.

So a consumer `<span>` passed as children (which renders inside the `<label>`) no longer matches `> span { display: none }` — the loading text stays in the accessibility tree and the button's `aria-describedby` points at real content. This flows through the placeholder to both the button `@extend` and the standalone `[data-overlay="true"]` case.

Verified in the compiled output:
- `button.css`: `.db-button .db-loading-indicator:not([data-overlay=true]) > .db-loading-indicator-content > span { display: none; }` (and `> label` visually-hidden)
- `loading-indicator.css`: `[data-overlay=true].db-loading-indicator > .db-loading-indicator-content > span { display: none; }`

Also added a comment on the rule explaining why direct-child scoping is required.

packages/components/src/components/loading-indicator/loading-indicator.lite.tsx
nmerget: You're right about the mechanism — `onMount` compiles to a post-render effect, so the SSR / first-paint markup does ship `data-state="inactive"` until it runs. After weighing it we've decided to keep it in `onMount` as a documented tradeoff rather than move the value into the render, for two reasons:

1. `state._loadingState` has to stay the template binding. It carries the internal transitions driven by `onTimeout` and state changes, so `data-state={props.state ?? state._loadingState}` would pin the DOM to the prop and stop reflecting those internal changes once a consumer passes `state`.
2. We don't want to reference `props.state` inside the `useStore` initializer — props are not reliably in scope there across all four targets, and nothing else in the codebase does it.

On the visual impact: the `inactive` first frame degrades gracefully rather than looking broken. For the default (uncontrolled) case it paints the bare track (bar) / empty circle, then `onMount` promotes it to `active` within a frame — reading as the indicator "starting" rather than flashing a wrong segment. On CSR the effect flushes before paint, so it isn't perceptible; the only genuinely-wrong case is a consumer passing `state="active"` in SSR seeing one inactive frame, which is brief and self-correcting.

I've added a comment at the `onMount` (beafab7be688) documenting this tradeoff so it isn't re-litigated. Happy to revisit if the SSR frame turns out to bite a real consumer.

__snapshots__/loading-indicator/showcase/chromium-highContrast/should-have-same-aria-snapshot/DBLoadingIndicator-should-have-same-aria-snapshot.yaml
```diff
@@ -0,0 +1,316 @@
...
+  - status:
+    - text: Circular vertical
+    - progressbar "Circular vertical"
+  - status:
+    - text: Bar
+    - progressbar "Bar"
+  - text: Successful
+  - status:
+    - text: Circular horizontal
+    - progressbar "Circular horizontal"
```
mfranzke, line 110: the component is both colored and an icon is being chosen for the `successful` and `critical` values of the `state` property. But that state isn't provided to the accessibility tree, e.g. by an accessible label (like `aria-label`).

We shouldn't fully rely on the authors providing this information e.g. in the included `label`, but provide a property instead, that we could than even also enforce e.g. via an eslint rule, if `state="successful"` or `state="critical"` have been defined.

__snapshots__/loading-indicator/showcase/mobile-chrome/should-have-same-aria-snapshot/DBLoadingIndicator-should-have-same-aria-snapshot.yaml
```diff
@@ -0,0 +1,316 @@
...
+  - status:
+    - text: Circular vertical
+    - progressbar "Circular vertical"
+  - status:
+    - text: Bar
+    - progressbar "Bar"
+  - text: "True"
+  - status:
+    - text: Circular vertical
+    - progressbar "Circular vertical"
```
mfranzke, line 237: even though that we define a live-region, this content doesn't include any "Loading" text inside. Even though that this is just an example, and there's isn't any visible "Loading" text anyhow, we should still provide a good example that even already includes this text even if visually hidden. The alternative would be to enforce it via property / eslint rule.

nmerget: Changed the examples

__snapshots__/loading-indicator/showcase/chromium-highContrast/should-have-same-aria-snapshot/DBLoadingIndicator-should-have-same-aria-snapshot.yaml
nmerget: I updated the `propsOverrides` you can set an ariaLabel if you want. But I don't like the idea of new properties like `successfulLabel`, `criticalLabel` etc. here. The updated example should make clear that you can use `ariaLabel` or the normal `label` or `progressText` that should be fine.

packages/components/src/components/loading-indicator/examples/_loading-indicator.arg.types.ts
```diff
@@ -0,0 +1,24 @@
...
+export const StorybookLoadingIndicatorArgTypes: Record<string, InputType> = {
+	variant: {
+		control: 'select',
+		options: ['bar', 'circular']
+	},
+	orientation: {
+		control: 'select',
+		options: ['horizontal', 'vertical']
+	},
+	size: { control: 'select', options: ['small', 'medium'] },
```
michaelmkraus: `size` supports `'large'` (model type, SCSS `[data-size="large"]` and the size example all use it) but this control only offers `['small', 'medium']`, so large can't be selected in Storybook.

packages/components/configs/plugins/attribute-passing/index.cjs
```diff
@@ -114,6 +114,18 @@ module.exports = () => ({
...
+					// component-specific: a consumer may set `role` on any
+					// custom-element host, and because the host is a
+					// display: contents wrapper the role has to move onto the
+					// real inner element (like data-/aria-). Forwarding is
+					// intentionally allowed to overwrite a component's own inner
+					// role (e.g. DBTabItem role="tab") -- the consumer owns the
+					// global attribute.
+					.replace(
+						'attr.name.startsWith("aria-")',
+						'attr.name.startsWith("aria-") || attr.name === "role"'
```
michaelmkraus, line 127: This adds global `role` forwarding for every Angular/Stencil component, not just DBLoadingIndicator, and `removeAttributes` it from the host. For components that declare `role` as a prop (e.g. `db-notification`), the Stencil attribute→prop sync can clear the prop.

packages/components/src/components/loading-indicator/examples/_loading-indicator.arg.types.ts
nmerget: Good catch, fixed. Added `'large'` to the `size` control options in `_loading-indicator.arg.types.ts`, so it now matches the model type (`SizeType | 'large'`) and the `[data-size="large"]` SCSS. The `large` variant is now selectable in Storybook.

packages/components/configs/plugins/attribute-passing/index.cjs
nmerget: This is the same concern we already discussed and resolved in the earlier threads on this file ([r4003422179](https://github.com/db-ux-design-system/core-web/pull/5992#discussion_r4003422179) and [r3997383147](https://github.com/db-ux-design-system/core-web/pull/5992#discussion_r3997383147)), so I'll keep the behavior as is.

Short version: `role` is a global attribute, so forwarding it off the `display: contents` custom-element host onto the real inner element is correct for every component (same rationale as `data-*`/`aria-*`) — leaving it on the contents wrapper would strand it. If a consumer deliberately sets `role` on a fixed-role widget's host, overriding the component's own inner role is the consumer exercising ownership of that global attribute, not a bug.

On the Stencil attribute→prop point for components that declare `role` as a prop (`db-loading-indicator`, `db-notification`): the consumer-facing path is the declared `role` prop, which is rendered onto the correct inner element via `getRole()` / `getNotificationRole()`, so the intended value is not lost. As offered before, if you'd prefer the observer to skip `role` for components that declare it as a prop (to avoid the redundant move + `removeAttribute` churn), I'm happy to add that narrowing — but the current behavior is intentional and correct.

mfranzke:
@nmerget review

packages/components/configs/plugins/attribute-passing/index.cjs
michaelmkraus: Agreed on keeping the global `role` forwarding — no need for the per-component narrowing, that would just add a special-case list without shrinking the actual blast radius. But could you add a test on the generated Stencil output that asserts `role` lands correctly on the inner element for the components that declare `role` as a prop (`db-notification`, `db-loading-indicator`)?

nmerget: Add a test for db-notification as well

packages/components/src/components/loading-indicator/docs/HTML.md
```diff
@@ -0,0 +1,62 @@
...
+			aria-hidden="true"
+		>
+			<circle class="db-loading-indicator-circle-track"></circle>
+			<circle class="db-loading-indicator-circle-segment"></circle>
+		</svg>
+		<div class="db-loading-indicator-content">
+			<label
+				id="loading-indicator-1-label"
+				for="loading-indicator-1-progress"
+			>
```
mfranzke, line 52:
```suggestion
			<label for="loading-indicator-1-progress">
```

The `id` seems to be some kind of a leftover to me.

michaelmkraus (approved):

PR #6056, opened 2026-02-14
refactor(CustomSelect): `disable` simplification

mfranzke:
## Proposed changes

Instead of deactivating every single interactive element in that CustomSelect, we should set the [`inert`](https://web.dev/articles/inert) attribute, and could remove unnecessary `aria-disabled` and `tabIndex` props.

## Types of changes

### DIFF
--- a/packages/components/src/components/custom-select/custom-select.lite.tsx
+++ b/packages/components/src/components/custom-select/custom-select.lite.tsx
@@ -922,7 +922,8 @@ export default function DBCustomSelect(props: DBCustomSelectProps) {
 			data-selected-type={props.multiple ? props.selectedType : 'text'}
 			data-hide-label={getHideProp(props.showLabel)}
 			data-icon={props.icon}
-			data-show-icon={getBooleanAsString(props.showIcon)}>
+			data-show-icon={getBooleanAsString(props.showIcon)}
+			inert={tabIndex={props.disabled ? "true" : undefined}}>
 			<label id={state._labelId}>
 				{props.label ?? DEFAULT_LABEL}
 				<select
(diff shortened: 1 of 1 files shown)


packages/components/src/components/custom-select/custom-select.lite.tsx
```diff
@@ -922,7 +922,8 @@ export default function DBCustomSelect(props: DBCustomSelectProps) {
 			data-selected-type={props.multiple ? props.selectedType : 'text'}
 			data-hide-label={getHideProp(props.showLabel)}
 			data-icon={props.icon}
-			data-show-icon={getBooleanAsString(props.showIcon)}>
+			data-show-icon={getBooleanAsString(props.showIcon)}
+			inert={tabIndex={props.disabled ? "true" : undefined}}>
```
nmerget, line 926: This would remove the element from the accessibility tree when I check the example on [mdn](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/inert).

I guess this might be bad, because a screen-reader user has another behavior. Without `inert` it reads something like `disabled summary` with `inert` the user can't set the screen-reader cursor on the element

mfranzke: > This would remove the element from the accessibility tree when I check the example on [mdn](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/inert).
> 
> I guess this might be bad, because a screen-reader user has another behavior. Without `inert` it reads something like `disabled summary` with `inert` the user can't set the screen-reader cursor on the element

I actually thought about this being a 'feature' rather than a 'bug' having them "removed" from the accessibility tree. On the other hand, however, it would differ from the visual experience, even though it's mainly a lack of information that this element is there instead of functionality.

I'll think this through again.

mfranzke:
for the moment, we won't proceed with this approach.

PR #6059, opened 2026-02-16
feat: table

nmerget:
## Proposed changes

Add Table component

resloves https://github.com/db-ux-design-system/core-team/issues/2503

## Types of changes

## Further comments

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/feat-table>


.github/workflows/01-build-stories.yml
mfranzke, line 1: what for did you change these path? The commit message just says "fix: issues with pipeline", could you please elaborate further?

nmerget: We have a new folder `templates` inside the storybooks. The build didn't work because we just uploaded the components

packages/components/src/components/table-body/index.html
```diff
@@ -0,0 +1,11 @@
+<!DOCTYPE html>
+<html lang="en">
+	<head>
+		<meta charset="UTF-8" />
+		<title>DBTableBody</title>
+		<link rel="stylesheet" href="/styles/relative.css" />
+	</head>
+	<body style="padding: var(--db-spacing-fixed-md)">
+		<div class="db-table-body">Test</div>
```
mfranzke: this doesn't look like a table at all. I totally understand that this is related to that we're including table contents in here like in any `children`, but even the template includes `<tbody>` , so is this actually correct?

same for all the other similar HTML files. Those even also mainly include a `div` element even for any other semantic elements that a table is based on.

packages/components/src/components/table/docs/HTML.md
```diff
@@ -0,0 +1,13 @@
...
+
+For general installation and configuration take a look at the [components](https://www.npmjs.com/package/@db-ux/core-components) package.
+
+### Use component
+
+```html index.html
+<!-- index.html -->
+...
+<body>
+
```
mfranzke: it's empty

packages/components/src/components/table/docs/Migration.md
```diff
@@ -0,0 +1,8 @@
+## General
+
+> **Note**
+> For a general installation or migration process check out this [documentation](https://www.npmjs.com/package/@db-ux/core-components).
+
+## DB UI Core ➡ DB UX Design System Core components
+
+New Component 🥳
```
mfranzke: that's not true – there as a component both in core and elements, compare to https://db-ui.github.io/core/?p=viewall-components-table and https://db-ui.github.io/elements/?path=/docs/04-data-display-table-intro--docs

packages/components/src/shared/model.ts
```diff
@@ -759,3 +759,48 @@ export interface PatternhubProps {
...
+export type DBTableCellProps = {
+	/**
+	 * The **`colSpan`** read-only property of the HTMLTableCellElement interface represents the number of columns this cell must span; this lets the cell occupy space across multiple columns of the table.
+	 *
+	 * [MDN Reference](https://developer.mozilla.org/docs/Web/API/HTMLTableCellElement/colSpan)
+	 */
+	colSpan?: number | string;
+
+	/**
+	 * The **`colSpan`** read-only property of the HTMLTableCellElement interface represents the number of columns this cell must span; this lets the cell occupy space across multiple columns of the table.
```
mfranzke: even though that this is the way of writing this attribute, as this is the description for providing this property with a small `s`.

packages/components/src/shared/model.ts
```diff
@@ -759,3 +759,48 @@ export interface PatternhubProps {
...
+	 *
+	 * [MDN Reference](https://developer.mozilla.org/docs/Web/API/HTMLTableCellElement/colSpan)
+	 */
+	colspan?: number | string;
```
mfranzke, line 777: what's the reason to provide both types of writings (lower- and uppercase `s`)? My assumption is that it's based on the consuming template language, but regarding developer experience, do we replace/remove either one based on the JS framework or would they get both attributes provided ? This might be confusing.

packages/components/src/components/table-row/table-row.lite.tsx
```diff
@@ -0,0 +1,139 @@
...
+	});
+
+	// jscpd:ignore-end
+
+	return (
+		<tr
+			ref={_ref}
+			id={props.id}
+			class={cls('db-table-row', props.className)}
+			data-interactive={getBooleanAsString(props.interactive)}
```
mfranzke, line 43: This reminds me again of that we even also do this elsewhere, having the users set the `interactive` property besides even inserting interactive elements as children (or at least controlling their properties). This seems to be an unintuitive redundancy. Wouldn't it be possible to prevent this and determine it by ourselves out of our logic ?

packages/components/src/components/table-row/table-row.scss
```diff
@@ -0,0 +1,118 @@
+@use "@db-ux/core-foundations/build/styles/variables";
+@use "@db-ux/core-foundations/build/styles/colors";
+@use "@db-ux/core-foundations/build/styles/helpers";
+@use "./../../styles/internal/table-components";
+
+$interactive-elements: ':is(input,button,a,select,textarea,[role="button"],[role="link"],[tabindex]:not([tabindex="-1"]))';
```
mfranzke: What about other interactive elements like e.g. `summary`? And further ARIA roles like e.g. `listbox`?

mfranzke: declaring the roles `[role="button"],[role="link"]` on elements doesn't make them interactive, but only by setting `[tabindex]:not([tabindex="-1"])` on those elements when not using a semantic HTML element. So we probably shouldn't promote just setting the `aria-role` directly without `[tabindex]:not([tabindex="-1"])` to be interactive, add that selector to the `role` selector parts (this would enforce a "correct" setting of `rrole`+ `tabindex`, but we might still miss some `role` definitions, see the other comment on this code part), or just keep the `role` declarations out of there in general.

mfranzke: shouldn't we move a CSS selector like this to a general central place in our codebase, so that we would prevent redundancy and enable reusage and updates at a central place in case of new developments ?

packages/components/src/components/table-row/table-row.scss
```diff
@@ -0,0 +1,118 @@
...
+		@content;
+	}
+}
+
+@mixin table-row-interactive-state($state: "hover") {
+	&:has(
+		#{$interactive-child}:#{$state}:not(
+				:disabled,
+				[aria-disabled="true"],
+				[tabindex="-1"]
```
mfranzke, line 29: this is a candidate for our new SCSS mixins, isn't it?

packages/components/src/components/table-row/table-row.scss
```diff
@@ -0,0 +1,118 @@
...
+			@extend %db-informational-variables;
+
+			background-color: colors.$db-adaptive-bg-basic-level-2-default;
+
+			&,
+			span,
+			p,
+			a {
+				font-weight: bold;
+			}
```
mfranzke, line 60:
```suggestion
			font-weight: bold;
```

Wouldn't setting `font-weight` directly be enough here, as it should get inherited to the included HTML elements ? Or is this specificity preventing incorrect elements getting this `bold` `font-weight`?

packages/components/src/components/table/docs/React.md
```diff
@@ -0,0 +1,17 @@
...
+
+### Use component
+
+```tsx App.tsx
+// App.tsx
+import { DBTable } from "@db-ux/react-core-components";
+
+const App = () => (
+	<DBTable>
+	</DBTable>
```
mfranzke: shouldn't we provide at least a very simple table with like e header and a body and some cells in it ?

packages/components/src/components/table/docs/Vue.md
```diff
@@ -0,0 +1,17 @@
...
+
+```vue App.vue
+<!-- App.vue -->
+<script>
+import { DBTable } from "@db-ux/v-core-components";
+</script>
+
+<template>
+	<DBTable>
+	</DBTable>
```
mfranzke: shouldn't we provide at least a very simple table with like a header and a body and some cells in it ?

packages/components/src/components/table-row/table-row.scss
nmerget: Yes we just want to set it to text-like elements. Otherwise something like a `<label>` for a checkbox becomes bold, which might indicate the checkbox is checked without the checkmark.

packages/components/src/components/table-body/index.html
nmerget: fixed

packages/components/src/components/table/docs/HTML.md
nmerget: fixed

packages/components/src/components/table/docs/Migration.md
nmerget: fixed

packages/components/src/shared/model.ts
nmerget: fixed

packages/components/src/shared/model.ts
nmerget: Add comment

packages/components/src/components/table-row/table-row.scss
nmerget: they have to add it by themselfs. This is just a nice to have way of auto detecting most of the elements

packages/components/src/components/table-row/table-row.lite.tsx
nmerget: We do something like this inside the DBTag, but there we assume that passing e.g. `<button>` as children that the DBTag become interactive. So the DBTag has a automatism.

For the TableRow we could also check for `data-table-row-trigger` with a `:has` that would be possible. But then there wouldn't be any hint inside the props. The `interactive` has a comment that you need to add `data-table-row-trigger` to some of you components. This information might not be working otherwise

packages/components/src/components/table-row/table-row.scss
nmerget: fixed

nmerget: fixed

packages/components/src/components/table-row/table-row.scss
nmerget: fixed

storybooks/react-storybook/src/templates/table-kitchen-sink/tableModels.tsx
```diff
@@ -0,0 +1,192 @@
...
+			<DBCheckbox
+				data-table-row-trigger="true"
+				label="Select All"
+				showLabel={false}
+				checked={row.getIsSelected()}
+				indeterminate={row.getIsSomeSelected()}
+				onChange={row.getToggleSelectedHandler()}>
+				<DBTooltip placement="top">Select All</DBTooltip>
+			</DBCheckbox>
+		),
```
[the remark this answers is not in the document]
nmerget: fixed

packages/components/src/components/table/docs/React.md
nmerget: fixed

packages/components/src/components/table/docs/Vue.md
nmerget: fixed

packages/components/src/components/table-row/table-row.lite.tsx
```diff
@@ -0,0 +1,139 @@
...
+			data-sub-header-emphasis={props.subHeaderEmphasis}>
+			<Show when={props.cells} else={props.children}>
```
mfranzke, line 45: at most other places (expect select and navigation-item; components that get build up our of composed contents) we've switched to having `props.children` being included additionally to some other input, like e.g. a `text` property. Is this implementation equally to select and navigation-item happening on purpose here, or does this need to get changed as well?

packages/components/src/components/table/examples/complex.example.lite.tsx
```diff
@@ -0,0 +1,830 @@
...
+										display: 'flex',
+										gap: 'var(--db-spacing-fixed-xs)'
+									}}>
+									Link
+									<DBButton
+										size="small"
+										variant="ghost"
+										icon="arrows_vertical"
+										noText>
+										Sort Link
```
mfranzke:
```suggestion
										Sort button
```

packages/components/src/components/table/examples/complex.example.lite.tsx
```diff
@@ -0,0 +1,830 @@
...
+										<DBTooltip placement="top">
+											Sort Link
```
mfranzke:
```suggestion
											Sort button
```

packages/components/src/components/table/examples/complex.example.lite.tsx
```diff
@@ -0,0 +1,830 @@
...
+										display: 'flex',
+										gap: 'var(--db-spacing-fixed-xs)'
+									}}>
+									Link
+									<DBButton
+										size="small"
+										variant="ghost"
+										icon="arrows_vertical"
+										noText>
+										Sort Link
```
mfranzke:
```suggestion
										Sort button
```

packages/components/src/components/table/examples/complex.example.lite.tsx
```diff
@@ -0,0 +1,830 @@
...
```
mfranzke:
```suggestion
										Sort button
```

packages/components/src/components/table/examples/complex.example.lite.tsx
```diff
@@ -0,0 +1,830 @@
...
+										<DBTooltip placement="top">
+											Sort Link
```
mfranzke:
```suggestion
											Sort button
```

packages/components/src/components/table/examples/complex.example.lite.tsx
```diff
@@ -0,0 +1,830 @@
...
```
mfranzke:
```suggestion
											Sort button
```

packages/components/src/components/table/examples/complex.example.lite.tsx
```diff
@@ -0,0 +1,830 @@
...
+										display: 'flex',
+										gap: 'var(--db-spacing-fixed-xs)'
+									}}>
+									Link
+									<DBButton
+										size="small"
+										variant="ghost"
+										icon="arrows_vertical"
+										noText>
+										Sort Link
```
mfranzke:
```suggestion
										Sort button
```

packages/components/src/components/table/examples/complex.example.lite.tsx
```diff
@@ -0,0 +1,830 @@
...
+										<DBTooltip placement="top">
+											Sort Link
```
mfranzke:
```suggestion
											Sort button
```

packages/foundations/scss/defaults/default-required.scss
```diff
@@ -65,6 +65,7 @@ body,
 /* stylelint-enable selector-max-id */
 
 body {
+	color: inherit;
```
mfranzke: what's the element this declaration would inherit from, `html`? Do we actually set colors in this one ourselves, or is this a hook for our users?

packages/foundations/scss/helpers/_interactive.scss
```diff
@@ -0,0 +1,2 @@
+$interactive-elements: "body,button,input,textarea,summary,select,a,[tabindex]:not([tabindex="-1
+	"])";
```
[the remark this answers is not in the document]
nmerget: fixed

packages/components/src/components/table-row/table-row.scss
```diff
@@ -0,0 +1,110 @@
+@use "@db-ux/core-foundations/build/styles/variables";
+@use "@db-ux/core-foundations/build/styles/colors";
+@use "@db-ux/core-foundations/build/styles/helpers";
+@use "./../../styles/internal/table-components";
+
+$interactive-row-action: '[data-table-row-trigger="true"]';
+$interactive-nested-child: "#{$interactive-row-action} input,textarea,select";
```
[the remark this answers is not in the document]
nmerget: fixed

packages/components/src/components/table/examples/complex.example.lite.tsx
nmerget: fixed

packages/components/src/components/table/examples/complex.example.lite.tsx
nmerget: fixed

packages/components/src/components/table/examples/complex.example.lite.tsx
nmerget: fixed

packages/components/src/components/table/examples/complex.example.lite.tsx
nmerget: fixed

packages/components/src/components/table/examples/complex.example.lite.tsx
nmerget: fixed

packages/components/src/components/table/examples/complex.example.lite.tsx
nmerget: fixed

packages/components/src/components/table/examples/complex.example.lite.tsx
nmerget: fixed

packages/components/src/components/table/examples/complex.example.lite.tsx
nmerget: fixed

packages/components/src/components/table-row/table-row.lite.tsx
nmerget: No we've splitted the `text` and `label` props with the `children`. In this case we consume a `JSON` with cells, similar to the `DBAccordion` you can use option API *or* composition API in this case not both

packages/foundations/scss/defaults/default-required.scss
nmerget: This is for our users. I added here to avoid a conflict from the [previous selector](https://github.com/db-ux-design-system/core-web/pull/6059/changes#diff-31654d0d5346be5fbf8d509f83a5e58e8d96b7236f7da006d55cae74a724a7c3L86-L95).

showcases/e2e/table/table-a11y-checker.spec.ts
```diff
@@ -0,0 +1,8 @@
+import { test } from '@playwright/test';
+import { hasWebComponentSyntax, runA11yCheckerTest } from '../default.ts';
+
+// The a11y checker will clash with the wrapping custom element from angular and stencil
```
mfranzke, line 4: wouldn't any QA tool on the users side even also report this problem ?

packages/foundations/scss/helpers/_interactive.scss
```diff
@@ -0,0 +1 @@
+$interactive-elements: 'body,button,input,textarea,summary,select,a,[tabindex]:not([tabindex="-1"])';
```
mfranzke:
```suggestion
$interactive-elements: 'button,input,textarea,summary,select,a,[tabindex]:not([tabindex="-1"])';
```

why would `body` be an interactive element ? To my understanding this is obsolete due to the other selector added with https://github.com/db-ux-design-system/core-web/pull/6059/changes#diff-31654d0d5346be5fbf8d509f83a5e58e8d96b7236f7da006d55cae74a724a7c3R68

mfranzke:
```suggestion
$interactive-elements: 'body,button,input,textarea,summary,select,a[href],[tabindex]:not([tabindex="-1"])';
```

An `a`-HTML-element without a `href`-HTML-attribute is only an anchor.

mfranzke:
@codex please review

showcases/e2e/table/table-a11y-checker.spec.ts
nmerget: Yes but that's the problem of those tools in combination of custom elements. They just follow the DOM with some self written rules, which don't reflect the accessibilty tree. It is accessible in the browser, but a flase positive in any tool

.changeset/add-table-component.md
```diff
@@ -0,0 +1,9 @@
+---
+"@db-ux/core-components": minor
+"@db-ux/react-core-components": minor
+"@db-ux/ngx-core-components": minor
+"@db-ux/v-core-components": minor
+"@db-ux/wc-core-components": minor
```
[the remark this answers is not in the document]
mfranzke: we need to further adjust these AGENTS instructions.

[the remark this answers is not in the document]
mfranzke: this has been corrected with https://github.com/db-ux-design-system/core-web/pull/6950

packages/components/scripts/post-build/components.ts
```diff
@@ -29,6 +29,38 @@ export type Component = {
 };
 
 export const getComponents = (): Component[] => [
+	{
+		name: 'table-data-cell'
+	},
```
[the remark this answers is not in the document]
nmerget: We need to keep this right now, we can't remove it until we remove the post-scripts

packages/components/src/components/table-row/table-row.lite.tsx
```diff
@@ -0,0 +1,139 @@
...
+											hreflang={cell.link?.hreflang}
+											target={cell.link?.target}
+											rel={cell.link?.rel}
+											referrerPolicy={
+												cell.link?.referrerPolicy
+											}
+											role={cell.link?.role}
+											showIcon={cell.link?.showIcon}
+											text={cell.link?.text}
+											wrap={cell.link?.wrap}>
```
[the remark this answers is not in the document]
nmerget: It's a link it should not have click handlers. If users wants to add this they have to use composition

packages/components/src/styles/internal/_table-components.scss
```diff
@@ -0,0 +1,60 @@
...
+	display: flex;
+
+	@for $span from 2 through 20 {
+		&[colspan="#{$span}"] {
+			grid-column: span var(--db-table-cell-span, #{$span});
+		}
+
+		&[rowspan="#{$span}"] {
+			grid-row: span var(--db-table-cell-span, #{$span});
+		}
```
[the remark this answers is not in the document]
nmerget: We'll add this as comment to the SCSS

packages/components/src/components/table/table.scss
```diff
@@ -0,0 +1,289 @@
...
+%set-grid-cells {
+	@for $num from 1 through 20 {
+		&:has(
+				tr
+					> :is(
+						td,
+						th,
+						db-table-data-cell,
+						db-table-header-cell
+					):nth-child(#{$num})
```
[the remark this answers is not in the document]
nmerget: We'll add this as comment to the SCSS

packages/foundations/scss/helpers/_interactive.scss
nmerget: fixed

mfranzke (approved):

PR #6109, opened 2026-02-20
feat: add `_id` property, prevent duplicate `id` in Angular and WC

### LINKED ISSUE #2017: Use a prefix for native attributes
In general we shouldn't use native attributes that we'd like to transfer to included elements within our components written as their native name, as this might interfere (think of e.g. `id` or the `aria-*`-attributes).

That for adding a dash or underscore in front of it would help to differentiate. If we'd like to target more than one specific element within a component, we could still add an identifier in the future previous to that attribute, like e.g. `image-id`.

### LINKED ISSUE #5462: Setting `id` attribute on db-ux custom elements breaks screenreaders

- [ ] All
- [ ] HTML
- [ ] React
- [x] Angular
- [ ] Vue
- [ ] Web components
- [ ] Power Apps

### Reproduction case

Screenreaders, like the one built into Windows 11, are not able to read certain text from db-ux custom elements if the **id** attribute is present on the element.

For example in this case, the screenreader is unable to read the "Label of username" text from the **label** attribute if the **id** attribute is present:
`<db-input id="username" label="Label of username"></db-input>`

And in this example the screenreader is also not able to read the text "My Switch" out loud:
`<db-switch id="my-switch">My Switch</db-switch>`

If the **id** attribute is removed, the screenreader works as expected.
Interestingly, attributes like **placeholder** are not affected by the **id** attribute and always work as expected.

### Expected Behaviour

Screenreaders should be able to read all accessibility related text from a db-ux element regardless whether the **id** attribute is set or not.

### Screenshots

_No response_

### Browser version

Chrome

_No response_

DB InfraGO AG

EBuLa Rückfallebene (IRE)

nmerget:
## Proposed changes

This PR introduces a new _id prop across core components to help avoid id attribute collisions/host-attribute side effects in Angular and Web Components outputs, and refactors many internal store fields to a new m* naming scheme to prevent clashes with the new prop.

- add `_id` property, prevent duplicate `id` in Angular and Web Component, because of the "wrapper" component tags still being rendered with the exact same `id`-HTML-attribute.

- refactored internal `_` state variables with `m` prefix to avoid confusion with `_` properties like `_id`

resolves #5462
resolves #2017

## Types of changes

## Further comments

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/feat-undescore-id-angular>


.changeset/flat-swans-enjoy.md
```diff
@@ -0,0 +1,9 @@
+---
+"@db-ux/core-components": minor
```
mfranzke:
```suggestion
```

this package is not related to this enhancement.

packages/components/scripts/post-build/react.ts
```diff
@@ -151,10 +151,10 @@ export default DB${upperComponentName};`
 						`{...getRootProps(props,${JSON.stringify(rootProps)})}` +
 						'\nclassName={'
 				},
-				/* We need to overwrite the internal state._value property just for react to have controlled components.
+				/* We need to overwrite the internal state.mValue property just for react to have controlled components.
 				 * It works for Angular & Vue, so we overwrite it only for React.  */
 				{
-					from: 'props.value ?? _value',
+					from: 'props.value ?? mValue',
```
mfranzke, line 157: we introcued new naming conventions, these even also need to get mentioned in the AI instructions.

PR #6137, opened 2026-02-24
refactor(dependency scans): ignore a specific CVE

mfranzke:
## Proposed changes

This [vulnerability](https://github.com/advisories/GHSA-3ppc-4f35-3m26) doesn't affect us, so we could ignore it.

## Types of changes

## Further comments

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/refactor-dependency-scans-ignore-a-specific-CVE>


mfranzke:
> I already ignored it on https://github.com/db-ux-design-system/core-web/security/dependabot, why shall we add this here?

you're right, I wasn't aware that those are in sync and I could do it this way around as well.

PR #6198, opened 2026-03-01
refactor(dependabot): added necessary update types for all dependencies

mfranzke:
## Proposed changes

by the fix with https://github.com/db-ux-design-system/core-web/commit/39f9d26e66a78fa61fd6210fc25049e0643b2c39 it became pretty clear that we need `update-types` for all `groups` entries, as dependabot elsewhere wouldn't group anymore, but only opens individual PRs, like e.g. https://github.com/db-ux-design-system/core-web/pull/6165, and afterwards opens the groups again, compare to https://github.com/db-ux-design-system/core-web/pull/6195

Todos:
- [ ] compare the entries with https://github.com/db-ux-design-system/core-web/pull/6135/changes/b4befd81d023df50c2f93742673d1af94f4e96d9
- [ ] check for further update groups except Angular, maybe they are special anyhow because if their `major` update type ignore

## Types of changes

## Further comments

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/refactor-added-necessary-update-types-for-various-dependencies-in-dependabot-yml>


mfranzke:
@nmerget with commit https://github.com/db-ux-design-system/core-web/pull/6135/changes/b4befd81d023df50c2f93742673d1af94f4e96d9 all `update-types` have been removed. The commit message or description doesn't state a reason for that, what problem were these definitions causing previously?

mfranzke:
we need to try this out first and it even also doesn't seem to be that big of a problem – the only remaining patch dependencies that haven't been grouped lately seem to have been based on dependabot alerts, which seem to get triggered independently from the grouped ones, and that for aren't getting grouped at all.

PR #6269, opened 2026-03-09
refactor: Simplify `cursor-pointer` mixin for inputs

mfranzke:
## Proposed changes

- We could simplify the `:is` selector by merging two declarations

## Types of changes

## Further comments

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/refactor-simplify-cursor-pointer-mixin-for-inputs>


packages/foundations/scss/helpers/_functions.scss
```diff
@@ -16,15 +16,9 @@ $cursor-pointer: var(--db-overwrite-cursor, pointer);
 	cursor: $cursor-pointer;
 	@content;
 
-	&:is(textarea),
-	&:is(input) {
+	&:is(textarea, input) {
 		cursor: initial;
 	}
-
-	&:is(input[type="checkbox"]),
```
nmerget, line 24: We need at least this for [DBTag](https://design-system.deutschebahn.com/core-web/review/refactor-simplify-cursor-pointer-mixin-for-inputs/components/data-display/tag/overview?page=behavior)

mfranzke: @nmerget within https://github.com/db-ux-design-system/core-web/blob/refactor-simplify-cursor-pointer-mixin-for-inputs/packages/components/src/components/tag/tag.scss this mixin isn't used, that property is being set directly within that file https://github.com/db-ux-design-system/core-web/blob/refactor-simplify-cursor-pointer-mixin-for-inputs/packages/components/src/components/tag/tag.scss#L146

Nevertheless do we need these lines of code at all, as we're even already setting `cursor: var(--db-overwrite-cursor, pointer)` via line 16 ?

nmerget: So we need some other logic for DBTag with a checkbox included. The checkbox gets `cursor:auto` on hover and active

packages/foundations/scss/helpers/_functions.scss
```diff
@@ -11,45 +11,25 @@
...
-			[aria-disabled="true"],
-			[tabindex="-1"],
-			:has(:disabled)
-		) {
-		@include cursor-pointer {
-			@content;
-		}
+@mixin hover($cursor: pointer) {
+	&:hover,
+	&:focus {
```
michaelmkraus: The new `hover()` mixin now automatically applies to `&:focus` as well. This means hover-styles will now apply as focus-styles everywhere this mixin is used! So this is actually a major change.

packages/foundations/scss/helpers/_functions.scss
```diff
@@ -11,45 +11,25 @@
...
-			[aria-disabled="true"],
-			[tabindex="-1"],
-			:has(:disabled)
-		) {
-		@include cursor-pointer {
-			@content;
-		}
+@mixin active($cursor: pointer) {
+	&:active {
+		cursor: var(--db-overwrite-cursor, #{$cursor});
```
michaelmkraus: The description mentions 'We don't need to set cursor: pointer', but the mixins still set it. We just handle it smarter now with parameters.

nmerget: We still need it for DBTag because with a radio we shouldn't show the pointer it it is selected. I removed the description

packages/foundations/scss/helpers/_functions.scss
nmerget:
```suggestion
```

packages/foundations/scss/helpers/_functions.scss
```diff
@@ -11,45 +11,24 @@
...
-			:disabled,
-			[aria-disabled="true"],
-			[tabindex="-1"],
-			:has(:disabled)
-		) {
-		@include cursor-pointer {
-			@content;
-		}
+@mixin hover($cursor: pointer) {
+	&:hover,
```
nmerget:
```suggestion
	&:hover {
```

packages/foundations/scss/helpers/_functions.scss
nmerget: fixed

packages/components/src/styles/internal/_form-components.scss
```diff
@@ -423,8 +423,10 @@ $input-valid-types:
 				#{variables.$db-spacing-fixed-sm}
 			);
 
-		@include helpers.hover {
-			background-color: colors.$db-adaptive-bg-basic-transparent-semi-hovered;
+		@include helpers.not-disabled {
+			@include helpers.hover(initial) {
```
mfranzke:
```suggestion
			@include helpers.hover(revert) {
```

this might be a bigger topic to have a look in our codebase in general, [`initial`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/initial) might not be the best choice, as it would reflect the CSS specs default, and neither the browsers default stylesheet, nor any  users choices:
> The initial value can be found in the formal syntax section of each CSS property reference page. For example, the [initial value of font-size is medium](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/font-size#formal_definition). The initial value should not be confused with the value specified by the browser's style sheet.

source: https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Cascade/Property_value_processing#initial_value

`revert` might actually be the better choice, as it would still lead to `initial` in the cascade of choices, but take either the browser stylesheet or users configurations (think of e.g. settings for accessibility) into account:
> it resets the property either to user agent set value, to user set value, to its inherited value (if it is inheritable), or to initial value. It can be applied to any CSS property

source: https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/revert

mfranzke: I've created https://github.com/db-ux-design-system/core-web/issues/6665 for the general followup

packages/foundations/scss/helpers/_functions.scss
```diff
@@ -11,45 +11,25 @@
...
-	}
-
-	&:is(input[type="checkbox"]),
-	&:is(input[type="radio"]:not(:checked)) {
-		cursor: pointer;
+@mixin not-disabled() {
+	&:not(#{$disabled-state}) {
+		@content;
 	}
 }
```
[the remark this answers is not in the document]
mfranzke: false positive, this is purely components specific.

packages/foundations/scss/helpers/_functions.scss
```diff
@@ -11,45 +11,25 @@
...
-			:has(:disabled)
-		) {
-		@include cursor-pointer {
-			@content;
-		}
+@mixin active($cursor: pointer) {
+	&:active {
+		cursor: var(--db-overwrite-cursor, #{$cursor});
+		@content;
 	}
```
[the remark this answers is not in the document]
nmerget: We moved the not-disabled to a own @mixin to avoid chaning unused selectors

packages/components/src/styles/internal/_form-components.scss
nmerget: This mixin sets `cursor:initial` instead of `cursor:pointer` that's all...

packages/components/src/components/tab-item/tab-item.scss
```diff
@@ -36,23 +36,27 @@ $with-icon-padding-calc: calc(
...
+					background-color: colors.$db-adaptive-bg-basic-transparent-full-hovered;
+				}
 
-			&:has(input:disabled) {
-				cursor: initial;
+				&:has(input:disabled) {
+					cursor: initial;
+				}
 			}
 		}
```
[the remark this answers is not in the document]
nmerget: fixed

packages/components/src/styles/internal/_form-components.scss
mfranzke: but why would we want to use `initial` instead of `revert` at all ?

packages/components/src/styles/internal/_form-components.scss
```diff
@@ -423,8 +423,11 @@ $input-valid-types:
 				#{variables.$db-spacing-fixed-sm}
 			);
 
-		@include helpers.hover {
-			background-color: colors.$db-adaptive-bg-basic-transparent-semi-hovered;
+		@include helpers.not-disabled {
+			// cursor: initial — text/textarea inputs should use the text cursor, not pointer
+			@include helpers.hover(revert) {
```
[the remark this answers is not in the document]
mfranzke: we'll update the comment

packages/components/src/styles/internal/_form-components.scss
```diff
@@ -423,8 +423,11 @@ $input-valid-types:
 				#{variables.$db-spacing-fixed-sm}
 			);
 
-		@include helpers.hover {
-			background-color: colors.$db-adaptive-bg-basic-transparent-semi-hovered;
+		@include helpers.not-disabled {
+			// cursor: initial — text/textarea inputs should use the text cursor, not pointer
```
mfranzke:
```suggestion
			// cursor: revert — text/textarea inputs should use the text cursor, not pointer
```

michaelmkraus (dismissed):

PR #6468, opened 2026-04-10
feat(mcp-server): add agentic code verification loop to migrate tool

michaelmkraus:
## Proposed changes

This PR upgrades the migration engine, introduces active code scaffolding, solidifies the data architecture to prevent LLM hallucinations, and adds visual context capabilities to the MCP server. 

### Key Changes
`analyze_v2_migration` — Deterministic Code Scanner

- New tool that acts as a local regex scanner for legacy v2 files. It identifies old components (<cmp-*, <elm-*), colors (db-color-*), and icons.

- Uses lazy-cached migration maps to cross-reference findings with our Single Source of Truth guides, returning a precise plan (120-char context limit per finding). This drastically reduces token consumption and eliminates AI hallucinations during large file migrations.

`scaffold_component` — Boilerplate Generator

- New tool to generate fully DB UX v3-compliant component structures on the fly.

- Supports 5 technology contexts out-of-the-box (React, Angular, Vue, Web Components, HTML).

- Automatically handles framework-specific naming conventions (PascalCase vs. kebab-case), file splitting (e.g., TS/HTML/SCSS for Angular, SFCs for Vue), and enforces mandatory SCSS foundation imports (default-variables & colors/variables).

### Single Source of Truth & Prebuild Pipeline

- Moved all migration guides out of the MCP package into the monorepo root (docs/migration/db-ui/) to establish a single source of truth.

- Implemented a robust npm-run-all prebuild pipeline using cpr. Guides and compiled tokens are now automatically copied into an assets/ directory during build, ensuring the server remains fully functional when executed standalone via npx.

### Defense-in-Depth Docs Filtering

- Eliminated context noise (and token waste) by strictly limiting docs_search.

- Added a build-time whitelist and a runtime safety net to exclusively allow components/ and foundations/docs/. ADRs, migration guides are now explicitly blocked from the general search.

### Compiled Design Tokens Extraction

- Rerouted `get_design_tokens` to read actual compiled CSS/SCSS values instead of raw SCSS loops for spacing, elevation, and density.

- Added multiline regex support to correctly parse complex shadow strings. The LLM now receives actionable primitive values (e.g., 0.75rem) and density class overrides.

`verify_migrated_code` — Compiler Self-Correction

- Added `verify_migrated_code` to run framework-specific compiler checks (e.g., tsc --noEmit) via secure, hidden temp files.

- Agentic Workflow: Updated the migrate.ts prompt to enforce a mandatory verification loop with a maximum of 3 retries.

`get_component_visual` — Visual Context for LLMs

- MCP tool that returns Base64-encoded reference images (dashboard, form, landingpage, table) for visual context during code generation.

- JPEG Compression & Downsampling: Images are dynamically processed via sharp (max 800x800 resolution, image/jpeg with quality 75) to strictly adhere to the 100k character limit for Base64 payloads, protecting LLM context windows.

- Opt-in only: The tool is explicitly marked as optional. LLMs should only invoke it when textual docs are insufficient.

## Types of changes

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/feat-mcp-agentic-feedback-loop>


.config/cspell.config.ts
```diff
@@ -34,6 +34,7 @@ export default defineConfig({
 		'**/test/**',
 		'packages/migration/**',
 		'packages/mcp-server/docs/**',
+		'packages/mcp-server/assets/**',
```
mfranzke:
```suggestion
```

This path (and the one above) actually include a lot of written contents – why would we want to exclude them ?

.stylelintignore
```diff
@@ -16,3 +16,4 @@ packages/foundations/scss/_normalize.scss
 packages/stylelint/test/**
 .stylelintrc.json
 packages/foundations/assets/icons/fonts/**
+packages/mcp-server/assets/
```
mfranzke, line 22: Why should we ignore those stylings? Are they mostly generated? Let's leave a comment here in this case so that we could remember.

packages/mcp-server/package.json
```diff
@@ -20,25 +20,28 @@
...
     "copy-build:readme": "cpr README.md ../../build-outputs/mcp-server/README.md --overwrite",
     "dev": "npm run build && node dist/index.js",
     "lint": "eslint src/ --ext .ts",
-    "start": "node dist/index.js",
     "test": "npm run build && vitest run",
     "test:watch": "vitest"
   },
   "dependencies": {
     "@modelcontextprotocol/sdk": "1.29.0",
+    "sharp": ">=0.33.5",
```
mfranzke: couldn't we pin this dependency instead ?

.config/cspell.config.ts
michaelmkraus: You are right, I added the missing words to the cspellingnorewords.txt and removed the two entries.

.stylelintignore
michaelmkraus: I added a comment why these are ignored (autogenerated, copied during prebuild).

.config/cspellignorewords.txt
```diff
@@ -171,3 +171,23 @@ cspellignorewords
 lintstagedrc
 carsharing
 bikesharing
+carsharing
+bikesharing
```
mfranzke:
```suggestion
```

.config/cspellignorewords.txt
```diff
@@ -171,3 +171,21 @@ cspellignorewords
...
+crosshairs
+giftcard
+iceportal
+komfort
+piggybank
+regioguide
+rollator
+sepa
+stepless
+steppless
```
mfranzke:
```suggestion
```

Was this word added intentionally? As far as I know, this isn't the correct spelling.

docs/migration/db-ui/component-migration.md
```diff
@@ -1,583 +1,105 @@
-# DB UI to DB UX Design System v3 Component Migration Guide
+# DB UI to DB UX v3 Component Migration
 
-> **Note**
-> This document provides migration guidance for DB UI components that don't have direct equivalents in DB UX Design System.
+Prefer JS framework packages: Angular `@db-ux/ngx-core-components` (`<db-button>`), React `@db-ux/react-core-components` (`<DBButton>`), Vue `@db-ux/v-core-components` (`<DBButton>`). CSS-only fallback: `@db-ux/core-components`.
```
mfranzke:
```suggestion
Prefer JS framework packages: Angular `@db-ux/ngx-core-components` (`<db-button>`), React `@db-ux/react-core-components` (`<DBButton>`), Vue `@db-ux/v-core-components` (`<DBButton>`) in case that the user wants to build a JavaScript application with one of these frameworks. If it's still a JavaScript framework, but none of these apply, use `@db-ux/core-components` node package. In all other cases, use the HTML and CSS basis: `@db-ux/core-components`.
```

To me this sounds biased regarding the tech stack, so we should differentiate by the users needs.

docs/migration/db-ui/component-migration.md
```diff
@@ -1,583 +1,105 @@
...
+**accordion-item** — `cmp-accordion`->`db-accordion-item` (was single item, now multiple items in one accordion). Props: `summary`->`headline`, `emphasis`->[removed], `size`->[removed] (use density). New: `disabled`, `content`.
 
-New Component (new component)
+**badge** — New component
 
-### accordion-item
+**brand** — `cmp-brand`->`db-brand`. All anchor-related props removed (`siteNameLink`,`alt`,`anchorRef`,`anchorTitle`,`anchorRelation`,`src`->[removed]). Wrap with your own `<a>` if needed. New: `hideLogo`.
 
-#### class
+**button** — `elm-button`->`db-button`. Sizes: `large`->[removed], `regular`->[removed] (default is `medium`, use density). Variants: `primary`->[removed] (use `outlined`), `brand-primary`->`primary`, `secondary-outline`->`outlined`, `secondary-solid`->`solid`, `tertiary-plain`->`text`. Icons: `iconTrailing`->[removed] (only leading icons), `iconOnly`->`noText`.
```
mfranzke, line 17: is this a file being generated? How do we keep those up to date elsewhere?

docs/migration/db-ui/component-migration.md
```diff
@@ -1,583 +1,105 @@
...
 
-| Before          |    Status     | After               | Description                                                                                                                      |
-| --------------- | :-----------: | ------------------- | -------------------------------------------------------------------------------------------------------------------------------- |
-| `cmp-accordion` | [replacement] | `db-accordion-item` | The old accordion was just 1 item, we split it to combine multiple accordion-items into 1 accordion (which is another component) |
+**card** — `cmp-card`->`db-card`. Variant: `header`->[removed], `content`->[removed] (use children/slot), `variant="interactive"` makes it act like a button. Image: `alt`->`imgAlt`, `image`->`imgSrc`. New: `imgWidth`, `imgHeight`. Removed: `illustration`, `uiCoreIllustrationPath`.
 
-#### props
+**checkbox** — `elm-checkbox`->`db-checkbox`. `input_id`->`id`.
```
mfranzke: this might not be correct, as we would prefer `propOverrides` for Web Components and Angular.

docs/migration/db-ui/component-migration.md
```diff
@@ -1,583 +1,105 @@
...
+| ---------------------- | :-----------: | ----------------------------------------------------------------- |
+| `rea-main`             |   [removed]   | `db-page` (includes `<main>`)                                     |
+| `rea-grid`             |   [removed]   | CSS Grid or `db-stack`                                            |
+| `rea-footer`           |   [planned]   | Semantic `<footer>` + `db-link`, planned Q4/2025                  |
+| `elm-headline`         | [replacement] | Semantic headings + `db-infotext` for subtitles                   |
+| `elm-headline` (pulse) |   [removed]   | Heading + CSS animation                                           |
+| `elm-loadingindicator` |   [planned]   | Planned Q4/2025, use CSS spinner + `aria-live`                    |
+| `elm-progress`         |   [planned]   | Planned Q4/2025, use `<progress>`                                 |
+| `elm-chip`             | [replacement] | `db-tag` with `db-button`/`db-checkbox`/`db-radio` inside         |
+| `cmp-breadcrumb`       |   [planned]   | Planned Q4/2025, use `db-link` in `<nav aria-label="Breadcrumb">` |
```
mfranzke, line 83: we're releasing this tool nowadays, but  still include a Q4/2025 in there. This doesn't feel right …

packages/mcp-server/agent/_instructions.md
```diff
@@ -53,6 +59,17 @@ Or add it to your MCP client config:
...
+
+1. **File Scan** — Call `analyze_v2_migration` with the file path. This returns a deterministic JSON report of all v2 patterns (components, colors, icons) with line numbers and suggestions. Use this as your migration checklist.
+2. **Migration Analysis** — Call `list_migration_guides`, then `get_migration_guide` for each relevant guide. Call `docs_search` for component-specific migration docs. Produce a mapping table: Legacy Element → DB UX v3 Component → Rationale.
+3. **Component Discovery & Props Retrieval** — Call `list_components` to verify every mapped component. For each: `get_component_props`, `get_component_details`, `get_example_code`. Call `get_design_tokens` to replace hardcoded values. Call `list_icons` to verify icon names. Optionally call `get_component_visual` for layout uncertainty.
+4. **Code Generation** — Generate the complete migrated code. Do **NOT** show it to the user yet.
+5. **Code Verification (MANDATORY for react/angular/vue)** — Call `verify_migrated_code`. If errors are returned, fix and retry (max 3 attempts). For `web-components` and `html` targets, skip this step.
```
mfranzke: maybe we should add the reason at the end of this sentence both for us to still remember as well as having this information available for the AI.

packages/mcp-server/assets/migration/component-migration.md
mfranzke, line 1: having a markdown file in the `assets` instead of `docs` folder is kind of … unexpected. Is this move really necessary?

packages/mcp-server/assets/migration/general-migration.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Inline Styles
+
+Remove all inline `style` attributes. They bypass theming, density, dark mode. Priority: 1) use component props, 2) CSS/SCSS class with tokens, 3) inline with tokens only as last resort.
+
+## Spacing
+
+Replace hardcoded `px`/`rem` in `padding`, `margin`, `gap`, `inset`, `top`/`right`/`bottom`/`left` with `--db-spacing-fixed-*` tokens. For responsive spacing use `--db-spacing-responsive-*`.
+
+Fixed tokens (regular density): `3xs`=2px, `2xs`=4px, `xs`=8px, `sm`=12px, `md`=16px, `lg`=24px, `xl`=32px, `2xl`=48px, `3xl`=80px. Values adapt to active density automatically. SCSS: `$db-spacing-fixed-md`, etc.
```
mfranzke: same here, how would we ensure that those stay up to date ?

packages/mcp-server/assets/migration/icon-migration.md
mfranzke, line 1: the v2 icon names have been divided by a dash instead of an underscore, compare to https://db-ui.github.io/core/?p=viewall-base-icons

packages/mcp-server/assets/visuals/.gitkeep
mfranzke, line 1: we don't need this file as we do have contents in there, do we ?

packages/mcp-server/assets/migration/general-migration.md
```diff
@@ -0,0 +1,33 @@
+# DB UI → DB UX General Migration
+
+Scan files: `.css`, `.scss`, `.html`, `.tsx`, `.jsx`, `.ts`, `.js`, `.vue`
+
+## Typography
+
+Replace hardcoded `font-size` with `font` shorthand using `--db-type-*` tokens. Use `font:` not `font-size:`. Remove redundant `line-height`/`font-weight`/`font-family` — the token sets all four.
```
mfranzke:
```suggestion
Replace hardcoded `font-size` CSS property with `font` shorthand using `--db-type-*` tokens. Use `font:` not `font-size:`. Remove redundant `line-height`/`font-weight`/`font-family` — the token sets all four.
```

Wouldn't it be necessary to mention "CSS property" at least once for clarity ?

packages/mcp-server/assets/migration/general-migration.md
```diff
@@ -0,0 +1,33 @@
...
+
+NEVER use `opacity` to simulate lighter colors. NEVER use `--db-elevation-*` as color values.
+
+## box-shadow → elevation tokens
+
+Do NOT use color tokens for `box-shadow`. Use: `--db-elevation-sm` (subtle/pressed), `--db-elevation-md` (default/cards), `--db-elevation-lg` (prominent/hover). SCSS: `$db-elevation-sm`/`md`/`lg`. Interactive: `data-interactive="elevation"` or `.db-interactive-elevation` (auto `md`→`lg` hover→`sm` active).
+
+## Inline Styles
+
+Remove all inline `style` attributes. They bypass theming, density, dark mode. Priority: 1) use component props, 2) CSS/SCSS class with tokens, 3) inline with tokens only as last resort.
```
mfranzke:
```suggestion
Replace all inline `style` attributes by external CSS. They bypass theming, density, dark mode. Priority: 1) use component props, 2) CSS/SCSS class with tokens, 3) inline with tokens only as last resort.
```

It's not just removing, but instead replacing, isn't it ?

packages/mcp-server/assets/migration/icon-migration.md
mfranzke, line 1: @leape have all icons by generation 2 been migrated to generation 3 already, or would we still need additions to this list afterwards ?

At least if (file)names change, we might need to update this list again.

packages/mcp-server/assets/tokens/default-variables.scss
mfranzke, line 1: do we need to add/commit this file into the codebase, or wouldn't that be part of the (pre)build process anyhow?

packages/mcp-server/assets/tokens/density-all.css
mfranzke, line 1: what's the source of these files contents, how do we keep it up to date ?

packages/mcp-server/scripts/prebuild.mjs
```diff
@@ -0,0 +1,123 @@
+/**
+ * Prebuild asset copy script for @db-ux/mcp-server.
```
mfranzke:
```suggestion
 * Prebuild asset copy script for @db-ux/mcp-server node package.
```

packages/mcp-server/scripts/prebuild.mjs
```diff
@@ -0,0 +1,123 @@
+/**
+ * Prebuild asset copy script for @db-ux/mcp-server node package.
+ *
+ * Copies migration guides and compiled token files from the monorepo into the
+ * local assets/ directories so they are available for standalone npx usage.
+ *
+ * All copies are OPTIONAL — if a source file or directory does not exist
+ * (e.g. in CI before foundations has been built, or in a partial checkout),
+ * the copy is skipped with a warning instead of aborting the build.
```
mfranzke: wouldn't this approach be a problem, as we might lack smaller or larger portions of our basic data out of that package? And if the pipeline wouldn't break, we wouldn't even notice. What's the reason for that?

packages/mcp-server/README.md
```diff
@@ -91,22 +91,53 @@ This will copy the correct rules for DB UX component usage and design token refe
...
+| `list_design_token_categories` | Lists all available token categories to pass to `get_design_tokens`.                                                                                                                                                                                                                                                     |
+| `list_icons`                   | Returns all valid DB UX icon names (e.g. `arrow_down`, `chevron_right`, `x_placeholder`). Always call this before using any `icon` prop — never guess a name.                                                                                                                                                            |
+| `docs_search`                  | Searches the DB UX conceptual documentation (guidelines, A11y, migration, ADRs) or component-specific markdown docs. Acts as our Retrieval-Augmented Generation (RAG) engine.                                                                                                                                            |
```
mfranzke:
```suggestion
| `docs_search`                  | Searches the DB UX conceptual documentation (guidelines, Accessibility, migration, ADRs) or component-specific markdown docs. Acts as our Retrieval-Augmented Generation (RAG) engine.                                                                                                                                            |
```

clarity

packages/mcp-server/src/tools/schemas.ts
```diff
@@ -63,7 +63,7 @@ export const getExampleCodeSchema = {
 
 export const docsSearchSchema = {
 	description:
-		'Searches the DB UX conceptual documentation (guidelines, A11y, migration, ADRs) or component-specific markdown docs.',
+		'Searches the DB UX component and foundation documentation (guidelines, A11y, framework-specific docs). Only docs from packages/components/ and packages/foundations/docs/ are included. For migration guides, use list_migration_guides and get_migration_guide instead.',
```
mfranzke:
```suggestion
		'Searches the DB UX component and foundation documentation (guidelines, Accessibility, framework-specific docs). Only docs from packages/components/ and packages/foundations/docs/ are included. For migration guides, use list_migration_guides and get_migration_guide instead.',
```

clarity

packages/mcp-server/src/tools/scanner.ts
```diff
@@ -0,0 +1,297 @@
...
+
+// ---------------------------------------------------------------------------
+// Tool Handler
+// ---------------------------------------------------------------------------
+
+/**
+ * Analyzes a file for DB UI v2 patterns that need migration to DB UX v4.
+ *
+ * Deterministically scans for:
+ * - v2 component tags (<cmp-*, <elm-*, <rea-*)
```
mfranzke: the components haven't been called like this – these have been values of the `class`-HTML-attribute, the Custom Elements / Web Components tags have been `db-*`, in which the asterisk was e.g. `brand`, resulting in e.g. `<db-brand>`, compare to e.g.
- https://db-ui.github.io/elements/?path=/docs/02-navigation-brand-intro--docs
- https://db-ui.github.io/core/?p=viewall-elements-buttons

packages/mcp-server/src/tools/schemas.ts
```diff
@@ -116,7 +109,84 @@ export const getMigrationGuideSchema = {
...
+			.describe(
+				"Name of the visual reference to load. Available visuals: 'dashboard', 'form', 'landingpage', 'table'."
+			)
+	}
+};
+
+export const analyzeV2MigrationSchema = {
+	description:
+		'IMPORTANT: Call this tool FIRST when asked to migrate a file. ' +
+		'Scans a source file for DB UI v2 patterns (components like <cmp-*, <elm-*, <rea-*>, ' +
```
mfranzke: the components haven't been called like this – these have been values of the `class`-HTML-attribute, the Custom Elements / Web Components tags have been `db-*`, in which the asterisk was e.g. `brand`, resulting in e.g. `<db-brand>`, compare to e.g.
- https://db-ui.github.io/elements/?path=/docs/02-navigation-brand-intro--docs
- https://db-ui.github.io/core/?p=viewall-elements-buttons

packages/mcp-server/README.md
```diff
@@ -91,22 +91,53 @@ This will copy the correct rules for DB UX component usage and design token refe
...
+| `verify_migrated_code`         | Saves generated code to a temp file and runs a compiler check (`tsc --noEmit`). Must be called after code generation and before showing code to the user. Returns diagnostics on failure so the AI can self-correct (max 3 attempts).                                                                                    |
+| `get_component_visual`         | Returns a downsampled screenshot (max 800×800 px, JPEG q75, bilinear interpolation) of a DB UX component or page layout as a Base64-encoded image. Use sparingly — only when visual context is needed for complex layouts, z-index dependencies, or visual hierarchies.                                                  |
+| `analyze_v2_migration`         | **Call FIRST when migrating a file.** Scans a source file for DB UI v2 patterns (`<cmp-*>`, `<elm-*>`, `<rea-*>` tags, `db-color-*` tokens, legacy icon names) and returns a JSON report with exact line numbers and deterministic migration suggestions from the official guides. No LLM guessing needed.               |
```
mfranzke: the components haven't been called like this – these have been values of the `class`-HTML-attribute, the Custom Elements / Web Components tags have been `db-*`, in which the asterisk was e.g. `brand`, resulting in e.g. `<db-brand>`, compare to e.g.
- https://db-ui.github.io/elements/?path=/docs/02-navigation-brand-intro--docs
- https://db-ui.github.io/core/?p=viewall-elements-buttons

packages/mcp-server/package.json
```diff
@@ -20,25 +20,28 @@
   "files": [
     "CHANGELOG.md",
     "agent",
+    "assets",
     "dist"
   ],
   "scripts": {
+    "prebuild": "node scripts/prebuild.mjs",
     "build": "node esbuild.js",
```
mfranzke:
```suggestion
    "build": "node scripts/prebuild.mjs && node esbuild.js",
```

we've deactivated the lifecycle scripts, which even also results in that our own lifecycle scripts would be disabled. A simple solution for this would to combine the commands.

packages/mcp-server/package.json
```diff
@@ -20,25 +20,28 @@
...
     "copy-build": "npm-run-all copy-build:*",
     "copy-build:agent": "cpr agent ../../build-outputs/mcp-server/agent --overwrite",
+    "copy-build:assets": "cpr assets ../../build-outputs/mcp-server/assets --overwrite",
     "copy-build:context": "cpr CONTEXT.md ../../build-outputs/mcp-server/CONTEXT.md --overwrite",
     "copy-build:dist": "cpr dist ../../build-outputs/mcp-server/dist --overwrite",
     "copy-build:package.json": "cpr package.json ../../build-outputs/mcp-server/package.json --overwrite",
     "copy-build:readme": "cpr README.md ../../build-outputs/mcp-server/README.md --overwrite",
     "dev": "npm run build && node dist/index.js",
     "lint": "eslint src/ --ext .ts",
-    "start": "node dist/index.js",
```
mfranzke, line 36: why has this script been removed ?

packages/mcp-server/assets/tokens/default-variables.scss
```diff
@@ -0,0 +1,1317 @@
...
+	--db-opacity-md: 0.4;
+	--db-opacity-lg: 0.68;
+	--db-opacity-xl: 0.76;
+	--db-opacity-2xl: 0.84;
+	--db-opacity-3xl: 0.92;
+	--db-opacity-full: 1;
+	--db-typography-regular-desktop-headline-3xl:
+		bolder 5rem/1.2 "OpenSans Head", helvetica, arial, sans-serif;
+	--db-typography-regular-desktop-headline-2xl:
+		bolder 4rem/1.25 "OpenSans Head", helvetica, arial, sans-serif;
```
mfranzke: this file currently includes the variables of the whitelabel theme, that most likely nobody wants to use.

.config/cspellignorewords.txt
michaelmkraus: We need both. stepless is correct, but in old icons it was spelled with pp. Therefore, we need both.

docs/migration/db-ui/component-migration.md
michaelmkraus: No, this file is manually maintained (Single Source of Truth). We keep the official migration guides here in the docs/ folder. The MCP server copies these files into its own assets/ folder during prebuild. When we build the MCP, it automatically pulls the latest version from this folder.

docs/migration/db-ui/component-migration.md
michaelmkraus: You are right. Since the MCP reads this markdown file to learn know to migrate, we might need to be more precise here so it generates the framework-specific code.

I update this to explicitly distinguish between the frameworks.

docs/migration/db-ui/component-migration.md
michaelmkraus: You are right. I’ve updated the file to remove the specific dates and changed the status to a general 'Planned'.

packages/mcp-server/agent/_instructions.md
michaelmkraus: Good point. Added this.

packages/mcp-server/assets/migration/component-migration.md
michaelmkraus: These files should not be committed. You are totally right! I will fix this.

packages/mcp-server/assets/migration/general-migration.md
michaelmkraus: Same issue here. I fixed this.

packages/mcp-server/assets/migration/icon-migration.md
michaelmkraus: Good catch. I adapted this!

packages/mcp-server/assets/visuals/.gitkeep
michaelmkraus: Correct, I removed it already.

packages/mcp-server/assets/tokens/default-variables.scss
michaelmkraus: Correct. I already removed it earlier.

packages/mcp-server/assets/tokens/density-all.css
michaelmkraus: The source of this file (and default-variables.scss) is the compiled output from our @db-ux/foundations package. When the MCP server is built, the prebuild script copies the compiled CSS from the foundations into this assets/tokens/ folder.

packages/mcp-server/assets/tokens/default-variables.scss
michaelmkraus: You are right. I just fixed the build script. It now correctly pulls the DB theme tokens directly from @db-ux/db-theme (and saves them as db-variables.scss instead of the generic default)

packages/mcp-server/package.json
michaelmkraus: It was removed by accident. I will add it back now.

packages/mcp-server/package.json
michaelmkraus: You are right, chaining them in the build command is the best solution

packages/mcp-server/scripts/prebuild.mjs
michaelmkraus: This was originally set up as a workaround for local development (so the MCP server build wouldn't crash if someone hadn't built the foundations yet). I will update the prebuild script to enforce strict existence of these source files and throw an error if anything is missing.

packages/mcp-server/src/tools/scanner.ts
michaelmkraus: I will update the JSDoc to reflect that the scanner looks for v2 CSS classes as well as the legacy Web Components correctly.

packages/mcp-server/README.md
michaelmkraus: I will update this in the README

packages/mcp-server/src/tools/schemas.ts
michaelmkraus: same here. I fixed it.

packages/mcp-server/assets/migration/icon-migration.md
michaelmkraus: image

.config/cspellignorewords.txt
mfranzke: let's use `// cSpell:ignore steppless` inline in those specific files than. Elsewhere we'll have an incorrect wording that would be the basis for mistakes in all other files as well.

docs/migration/db-ui/color-migration.md
```diff
@@ -1,148 +1,150 @@
 # DB UI → DB UX Color Migration
 
-| DB UI Token                       | DB UI Color                                                                                                                                                | Best Background Match                                  | Background Color                                                                                                                                           | Best Foreground Match                                | Foreground Color                                                                                                                                           |
```
nmerget, line 3: My idea was to have a human readable migration and one for the mcp. We can merge them if we don't need the human readable documentation

packages/mcp-server/src/utils/path.ts
```diff
@@ -29,6 +68,42 @@ export const TOKEN_FILES: Record<string, string> = {
 	transitions: join(FOUNDATIONS_DIR, 'scss/animation/_transitions.scss')
 };
 
+/**
+ * Compiled token files that contain the actual primitive values
+ * (rem, px, box-shadow strings) instead of raw SCSS with @each loops.
+ *
+ * Primary: monorepo source paths (packages/foundations/...).
```
nmerget: Why do we search in our monorepo for those files? Shouldn't we bundle them always inside the mcp to test if the shipped package works as expected? 

We should include our own mcp inside the monorepo like we would do when we use it from outside

packages/mcp-server/src/prompts/migrate.ts
```diff
@@ -0,0 +1,158 @@
...
+   - Call 'get_example_code' with the most relevant example and framework="${target_framework}" to obtain the exact generated source. Adapt this code — do NOT rewrite from scratch.
+c. Call 'list_design_token_categories', then 'get_design_tokens' for every token category used in the legacy code (colors, spacing, typography, etc.). Replace ALL hardcoded values.
+d. If any icon is used, call 'list_icons' and copy the exact name from the returned array.
+e. OPTIONAL — Visual validation: If you are uncertain about layout structures, z-index stacking, or visual hierarchies of a component, call 'get_component_visual' with the component name. This returns a downsampled screenshot for visual reference. Use sparingly — only when textual docs are insufficient.
+
+Native HTML replacement rules (enforce strictly):
+- <button>      → DBButton
+- <input>       → DBInput
+- <select>      → DBSelect
+- <a>           → DBLink
```
nmerget: It's valid to have an `<a>` somewhere or also some `<Link>` component which comes from e.g. `react-router`

packages/mcp-server/src/prompts/migrate.ts
```diff
@@ -0,0 +1,158 @@
...
+- <textarea>    → DBTextarea
+- <div> layout  → DBStack, DBSection, DBCard
```
nmerget: Also a <div> is completly valid. Otherwise the the agent will stack 10 Cards into each other which might be bad

packages/mcp-server/src/prompts/migrate.ts
```diff
@@ -0,0 +1,158 @@
...
+	target_framework
+}: {
+	legacy_code: string;
+	source_context: string;
+	target_framework: string;
+}) {
+	const boundary = `LEGACY_CODE_${Date.now()}_${Math.random().toString(36).slice(2)}`;
+	const pkg =
+		FRAMEWORK_PKG[target_framework as Framework] ??
+		`@db-ux/${target_framework}-core-components`;
```
nmerget, line 26: There are some projects which might use plain HTML+JS with a e.g. Python backend. In this case we need to use `@db-ux/core-components`

docs/migration/db-ui/color-migration.md
michaelmkraus: I understand the thought behind splitting. However, maintaining two migration mappings (one for humans, one for the MCP) is risky because they will drift out of sync over time. Maybe we should use a single, cleanly formatted Markdown table. It is readable for developers and for the AI.

packages/mcp-server/src/tools/scaffold.ts
```diff
@@ -0,0 +1,365 @@
...
+		.toLowerCase()
+		.replace(/[^a-z0-9-]/g, ''); // Strip anything except alphanumeric + hyphens (path traversal protection)
+}
+
+// ---------------------------------------------------------------------------
+// SCSS Foundation Imports (mandatory for every generated SCSS block)
+// ---------------------------------------------------------------------------
+
+const SCSS_IMPORTS = `@use "@db-ux/foundations/scss/defaults/default-variables" as db-vars;
+@use "@db-ux/foundations/scss/colors/variables" as db-colors;`;
```
nmerget: Those are not mandatory. 

You would use `@use "@db-ux/foundations/colors` like `colors.$db-bg-basic-level-1-default` if you need color. And you would use `@use "@db-ux/foundations/variables` like `variables.$db-spacing-fixed-md`. But only if you need it.

packages/mcp-server/src/tools/scaffold.ts
```diff
@@ -0,0 +1,365 @@
...
+ * - @use "@db-ux/foundations/scss/colors/variables"
+ *
+ * ### Security
+ * - **Path traversal protection:** `targetPath` is resolved against `process.cwd()`
+ *   and must stay within it. The component name is sanitized to `[a-z0-9-]` only.
+ * - **No silent overwrites:** Refuses to write if any target file already exists.
+ * - **Graceful error handling:** FS errors (permissions, disk full) are caught and
+ *   returned as MCP error results instead of crashing the server.
+ */
+export async function handleScaffoldComponent({
```
nmerget: This scaffold always adds SCSS which is bad.

It should check which CSS styling is used. Maybe the project uses TW or plain CSS, CSS modules. We should respect this. Furthermore Angular has it's own scaffolding via CLI, we should repect this two, because many would use it.

There are also multiple ways to build your projects layout for every Framework. the scaffolding should repect this. In Vue you could have a `<styles>` tag or you import the CSS file with an import.


Can we move this into another branch? It's not related with the migration directly isn't it?

packages/mcp-server/src/utils/path.ts
michaelmkraus: The original idea behind reading directly from the monorepo was for local development experience – so we wouldn't have to rebuild the MCP every time we adapt a markdown guide etc. However, you are right. If our build script fails to copy sth into the MCP's assets/ folder, we wouldn't notice it locally, but the published npm package would be broken.

packages/mcp-server/src/tools/scanner.ts
```diff
@@ -0,0 +1,302 @@
...
+
+	const manifest = await getManifest();
+	const guides = manifest.migrationGuides ?? {};
+
+	componentMap = parseComponentMap(
+		findGuide(guides, 'component-migration') ?? ''
+	);
+	colorMap = parseColorMap(findGuide(guides, 'color-migration') ?? '');
+	iconMap = parseIconMap(findGuide(guides, 'icon-migration') ?? '');
+}
```
nmerget: Why do we need this in the first place?

We use the .md files to provide the context for AI. Then we parse the .md files to JSON.
Can we just save those files as JSON/JS/TS instead of .md?

The AI reads token it doesn't matter the format of the file. So we can consume those directly?

packages/mcp-server/src/tools/scanner.ts
```diff
@@ -0,0 +1,302 @@
...
+ *
+ * Deterministically scans for:
+ * - v2 CSS classes (cmp-*, elm-*, rea-*) and v2 Web Components (<db-*)
+ * - v2 color tokens (db-color-*)
+ * - v2 icon names (cross-referenced against the icon migration guide)
+ *
+ * Returns a JSON report with line numbers, findings, and migration suggestions
+ * resolved from docs/migration/db-ui/*.md — no LLM guessing needed.
+ */
+export async function handleAnalyzeV2Migration({
```
nmerget: This is the analayze tool, but the files name is `scanner.ts` we should align it and rename the function or the file.

packages/mcp-server/src/prompts/migrate.ts
michaelmkraus: You are right. An instructing to migrate every `<a>` tag to DBLink is not good and would e.g. break application routing. I will remove the `<a>` mapping from this list.

packages/mcp-server/src/prompts/migrate.ts
michaelmkraus: Good point! As the `<a>` tags, `<div>` elements are necessary for grouping. I will remove the `<div>` mapping from the prompt instructions right away as well!"

packages/mcp-server/src/prompts/migrate.ts
michaelmkraus: Good point, didn't think of that. I will update the mapping to resolve html and js to the @db-ux/core-components package.

packages/mcp-server/src/tools/verify.ts
```diff
@@ -0,0 +1,200 @@
...
+
+/**
+ * Builds the shell command used to verify the temporary file.
+ * Uses `npx --no-install tsc --noEmit` for React/Angular (TypeScript) and
+ * `npx --no-install vue-tsc --noEmit` for Vue single-file components.
+ *
+ * File paths are always double-quoted to handle spaces and special characters
+ * on all operating systems (Unix + Windows).
+ */
+function buildCheckCommand(filePath: string, framework: string): string {
```
nmerget: This is only for TS with `tsc` compiler. If a project uses JS this would fail. If a project uses `vite` this might fail as well, because vite runs its own compiling which isn't 100% compatible with tsc. 

I wouldn't recommend doing it like this. Instead we can return something like `Check for lint, build, check scripts in the to test if the app still works` this would allow the AI to read the project setup and run the correct commands

packages/mcp-server/src/tools/visuals.ts
```diff
@@ -0,0 +1,138 @@
...
+ * as JPEG (quality 75), and returns it as a Base64-encoded MCP image block.
+ *
+ * Images are resolved from the `assets/visuals/` directory that ships
+ * with the `@db-ux/mcp-server` package.
+ *
+ * This tool is **opt-in only** — it should be invoked by the LLM when visual
+ * context is needed for complex layouts, z-index reasoning, or verifying
+ * visual hierarchies.
+ */
+export async function handleGetComponentVisual({
```
nmerget: Is this part of the migration as well? We could separate it to another branch couldn't we?

packages/mcp-server/src/tools/scaffold.ts
michaelmkraus: I will remove the scaffold tool from this branch so we can focus on getting the migration. We can handle scaffolding in a separate ticket/branch later.

packages/mcp-server/src/tools/tokens.ts
```diff
@@ -1,34 +1,174 @@
...
+				) {
+					i++;
+					result.push(allLines[i]);
+				}
+			}
+		}
+	}
+
+	return result.join('\n');
+}
```
nmerget: Again here, we do a prebuild step with a script to copy `.md`, `.scss` files etc.

Can we just transform it to JSON/JS/TS files with the data as object to avoid parsing the files always inside the MCP?

packages/mcp-server/scripts/prebuild.ts
```diff
@@ -0,0 +1,117 @@
+/**
```
nmerget, line 1: Can we use `.ts` here as well, we use node 24 so this shouldn't be a problem

packages/mcp-server/src/tools/scanner.ts
michaelmkraus: The reason they are .md files right now is because they are used as human-readable documentation. The best architecture would be to use structured JSON/TS, consume that in the MCP server, and auto-generate the .md files for the humans from that data.
That is a big change. Let's merge the regex parsers for now and create a dedicated follow-up ticket to refactor the migration to structured data.

packages/mcp-server/src/tools/verify.ts
```diff
@@ -0,0 +1,200 @@
...
+ *   compiler output is fully captured before {@link truncate} reduces it.
+ */
+export async function handleVerifyMigratedCode({
+	code,
+	framework
+}: {
+	code: string;
+	framework: string;
+}): Promise<ToolResult> {
+	const ext = FRAMEWORK_EXT[framework] ?? '.ts';
```
nmerget: What if someone uses JS?

For Angular only the `.ts` file is verified. But there migth be issues with the `.html` file which is connected with the `.ts` file. It would fail inside the `ng build`

packages/mcp-server/src/tools/scanner.ts
michaelmkraus: I will rename the handler function and ensure the registered tool name itself is aligned to use 'scan' instead of 'analyze'.

packages/mcp-server/src/tools/scanner.ts
nmerget: https://github.com/db-ux-design-system/core-web/pull/6468#discussion_r3151997672

The question is, do we need a human readable file at all? 

No one will do a migration hand by hand now, when there is AI. And also a JSON/TS file is readable/searchable for humans. Let's convert them all to to JSON/TS to reduce maintenance

packages/mcp-server/src/tools/verify.ts
nmerget: `{
  "output": {
    "kind": "text",
    "content": "{\"content\":[{\"type\":\"text\",\"text\":\"❌ Verification failed for angular code.\\n\\n.tmp_verify_18d0ac60-e63a-47de-9c83-8ce20dd8e0c8.ts(2,34): error TS2307: Cannot find module './components/content/content.component' or its corresponding type declarations.\\r\\n.tmp_verify_18d0ac60-e63a-47de-9c83-8ce20dd8e0c8.ts(3,35): error TS2307: Cannot find module './components/db-footer/db-footer.component' or its corresponding type declarations.\\r\\n.tmp_verify_18d0ac60-e63a-47de-9c83-8ce20dd8e0c8.ts(4,35): error TS2307: Cannot find module './components/db-header/db-header.component' or its corresponding type declarations.\\n\\nPlease fix the errors above and call verify_migrated_code again.\"}],\"isError\":true}"
  }
}
Die Fehler sind nur relative Import-Pfade, die in der isolierten Sandbox nicht aufgelöst werden können – das ist erwartetes Verhalten bei verify_migrated_code für Dateien mit lokalen Imports. `

There might be issues like this if we just use temporary files with `tsc` :/

michaelmkraus: You are right! I'll refactor the command builder to return a dynamic prompt instead of a hardcoded command.

packages/mcp-server/src/tools/visuals.ts
michaelmkraus: You are right, it is not actually in the scope of this branch. The background here is: we received the visual assets from our colleagues, so I wanted to wire them up directly while I was working on the MCP tools.

nmerget:
@michaelmkraus did you verify those changes with an real project?

In my testing some of the tools were not called, and some of them did not work as expected ^.^

But overall the result is good :)

packages/mcp-server/src/tools/tokens.ts
michaelmkraus: You are right, this is the same as the Markdown parsing comment earlier.
We can use the current approach as a MVP for this PR and clean up the asset pipeline (Markdown + SCSS -> JSON) in a follow-up ticket.

packages/mcp-server/scripts/prebuild.ts
michaelmkraus: I will rename the script to .ts and update the package.json script references.

packages/mcp-server/src/tools/verify.ts
michaelmkraus: Good point. I will remove out this hardcoded extension mapping and the file creation.

docs/migration/db-ui/color-migration.md
michaelmkraus: I merged this into a standard Markdown table.

packages/mcp-server/src/tools/scanner.ts
michaelmkraus: Ich habe hierfür ein Ticket erstellt https://github.com/db-ux-design-system/core-team/issues/2428

packages/mcp-server/src/tools/visuals.ts
michaelmkraus: I removed it from this branch.

nmerget:
## Code Review: `feat-mcp-agentic-feedback-loop`

### 1. Syntax error in `build-tokens.ts` line 57
`return '__skip__\n'` has a literal newline inside the string. The `=== '__skip__'` check in `addToken` will never match, causing base tokens to leak into the `other` category instead of being skipped.

**Fix:** `return '__skip__';`

### 2. Path traversal check fragile on Windows (`scanner.ts`)
`absolutePath.startsWith(cwd + sep)` uses `sep` (`\` on Windows), which can break with mixed separators. Normalize both paths before the check, consistent with `resolveSafePath` in `path.ts`.

### 3. `resolveVisualPath` calls `existsSync` in a loop (`visuals.ts`)
Iterates over extensions with one `existsSync` call per candidate. A single `readdir` + lookup would be more consistent with `listAvailableVisuals`.

### 4. Superfluous `inputSchema: {}` in `verifyMigratedCodeSchema` (`schemas.ts`)
The handler takes no parameters, so `inputSchema` should be omitted entirely — consistent with other parameter-less schemas like `listVisualsSchema`.

### 5. Stray newline in `docsSearchSchema` description (`schemas.ts`)
A raw newline mid-string in the template literal will render as a literal newline in the MCP tool description.

### 6. `Math.random()` used for prompt boundary (`migrate.ts`)
Not cryptographically random. Fine for injection prevention, but a comment clarifying that this is not a security primitive would help future readers.

### 7. Breaking Node.js engine bump (`package.json`)
`engines.node` bumped from `>=22` to `>=24`, dropping support for Node 22 LTS. Should be explicitly called out in the changelog/PR description.

.github/copilot-instructions.md
```diff
@@ -113,6 +113,63 @@ pnpm run regenerate:screenshots
...
+- **`patch`** — for bug fixes
+- **`minor`** — for new features
+- **`major`** — for breaking changes (e.g. a property in any `model.ts` has been added, removed, renamed, or its type has changed)
+
+### How to Add a Changeset
+
+Run the following command and follow the interactive prompts:
+
+```bash
+npx changeset
```
mfranzke, line 140:
```suggestion
pnpm exec changeset
```

.github/copilot-instructions.md
```diff
@@ -310,18 +311,33 @@ Or add it to your MCP client config:
...
+12. `scan_v2_migration` — **call FIRST when migrating a file.** Scans a source file for v2 CSS classes (`cmp-*`, `elm-*`, `rea-*`), v2 Web Components (`<db-*>`), `db-color-*` tokens, and legacy icons. Returns a JSON report with line numbers and deterministic suggestions.
 
 #### DON'Ts — these are hard violations
 
-- **NEVER** use native HTML elements (`<button>`, `<input>`, `<select>`, `<a>`) when a DB UX component exists (e.g. `DBButton`, `DBInput`, `DBSelect`, `DBLink`)
-- **NEVER** use `<div>` or `<span>` for layout when `DBStack`, `DBSection`, or `DBCard` apply
+- **NEVER** use native HTML elements (`<button>`, `<input>`, `<select>`, `<textarea>`) when a DB UX component exists (e.g. `DBButton`, `DBInput`, `DBSelect`, `DBTextarea`)
+- **NEVER** blindly replace `<a>` tags with `DBLink` — this breaks framework routing (e.g. react-router `<Link>`). Only replace `<a>` when it is explicitly styled as a UI action component
```
mfranzke:
```suggestion
- **NEVER** replace any `<a>` tags with `DBLink` — this breaks framework routing (e.g. react-router `<Link>`). Only replace `<a>` when it is explicitly styled as a UI action component
```

.github/copilot-instructions.md
```diff
@@ -310,18 +311,33 @@ Or add it to your MCP client config:
...
+- **NEVER** force-replace generic `<div>` elements with `DBStack`/`DBSection`/`DBCard` — plain `<div>` is valid HTML. Only use DB UX layout components when the design explicitly calls for them
```
mfranzke:
```suggestion
- **NEVER** force-replace generic `<div>` elements with `DBStack`/`DBSection`/`DBCard` — plain `<div>` is valid HTML. Only use DB UX layout components when the design explicitly calls for them, and `DBStack`/`DBSection`/`DBCard` would provide semantics or a specific grouping
```

.github/copilot-instructions.md
mfranzke: there's another follow up issue for this.

nmerget (dismissed):

nmerget (dismissed):

nmerget (approved):

PR #6477, opened 2026-04-13
refactor(select): hide empty `option` for placeholder and floating label right away

mfranzke:
## Proposed changes

As a follow up to https://github.com/db-ux-design-system/core-web/pull/5888, we should even already hide the first, empty `option` that we're inserting for placeholder and floating label DBSelect components, right away, in case that `showEmptyOption` is defined. That for we use the `:open` CSS pseudo selector as progressive enhancement.

As soon as `:open` is supported widely ([only Safari is missing support currently](https://caniuse.com/wf-open-pseudo), but will release it most likely with version 26.5), we could even also remove the first part of this `:has` selector.

## Types of changes

## Further comments

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/refactor-select-hide-our-empty-option-right-away>


packages/components/src/components/select/select.scss
```diff
@@ -44,7 +44,7 @@
 		background-color: colors.$db-adaptive-bg-basic-level-3-default;
 	}
 
-	&:has(> select option:checked:not(.placeholder)) {
+	&:has(> select option:checked:not(.placeholder), > select:open) {
```
mfranzke:
```suggestion
	// TODO: we could remove the `> select option:checked:not(.placeholder), ` selector part as soon as `:open` is baseline.
	&:has(> select option:checked:not(.placeholder), > select:open) {
```

packages/components/src/components/select/select.scss
```diff
@@ -44,7 +44,8 @@
 		background-color: colors.$db-adaptive-bg-basic-level-3-default;
 	}
 
-	&:has(> select option:checked:not(.placeholder)) {
+	// TODO: we could remove the `> select option:checked:not(.placeholder), ` selector part as soon as `:open` is baseline.
+	&:has(> select option:checked:not(.placeholder), > select:open) {
 		[id$="-placeholder"],
 		// Hide placeholder option in dropdown for required selects after a selection has been made
```
mfranzke:
```suggestion
		// Hide inserted `option` for placeholder or floating label types in dropdown for `required` selects after a selection has been made or the `select` has been opened
```

packages/components/src/components/select/select.scss
```diff
@@ -44,9 +44,10 @@
 		background-color: colors.$db-adaptive-bg-basic-level-3-default;
 	}
 
-	&:has(> select option:checked:not(.placeholder)) {
+	// TODO: we could remove the `> select option:checked:not(.placeholder), ` selector part as soon as `:open` is baseline.
+	&:has(> select option:checked:not(.placeholder), > select:open) {
```
michaelmkraus:
```suggestion
	&:is(:has(> select option:checked:not(.placeholder)), :has(> select:open)) {
```
We could use `:is()` to prevent browsers without `:open` support (like Safari) from invalidating the rule.

mfranzke:
```suggestion
	&:is(:has(> select option:checked:not(.placeholder)), > select:open) {
```
Great idea ! We could as well get rid of that second `:has` anyhow.

nmerget:
@mfranzke Can you add an example with `showEmptyOption` to test it directly, thank you

mfranzke:
> @mfranzke Can you add an example with `showEmptyOption` to test it directly, thank you

you're welcome

packages/components/src/components/select/select.scss
```diff
@@ -44,10 +44,10 @@
 		background-color: colors.$db-adaptive-bg-basic-level-3-default;
 	}
 
-	&:has(> select option:checked:not(.placeholder)) {
-		[id$="-placeholder"],
-		// Hide placeholder option in dropdown for required selects after a selection has been made
-		// This prevents users from selecting an invalid empty option when the field is required
+	select:is(:has(option:checked:not(.placeholder)), :open) {
+		& + [id$="-placeholder"],
```
mfranzke:
```suggestion
		& ~ [id$="-placeholder"],
```

packages/components/src/components/select/select.scss
```diff
@@ -44,9 +44,9 @@
...
 
-	&:has(> select option:checked:not(.placeholder)) {
-		[id$="-placeholder"],
-		// Hide placeholder option in dropdown for required selects after a selection has been made
+	select:is(:has(option:checked:not(.placeholder)), :open) {
+		& + [id$="-placeholder"],
+		// Hide inserted `option` for placeholder or floating label types in dropdown for `required` selects after a selection has been made or the `select` has been opened
 		// This prevents users from selecting an invalid empty option when the field is required
 		option[data-show-empty-option="false"] {
 			display: none;
```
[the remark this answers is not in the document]
mfranzke: using the `~` instead of `+` selector to solve this.

nmerget (approved):

PR #6538, opened 2026-04-21
fix: align package versions before packing to prevent ENOENT on publish

michaelmkraus:
### Problem
The CI/CD publish job failed with:

```
npm error enoent ENOENT: no such file or directory, open '…/build-outputs/db-ux-mcp-server-4.6.1.tgz'
The newly added @db-ux/mcp-server package was missing from the publish script, so it was never packed into a .tgz — but the pipeline expected it to exist.
```

### Root Cause

| Step | What happens | Problem |
|---|---|---|
| `changeset version` | Bumps versions of packages with changesets | `mcp-server` was new and had no changeset → version stayed at `4.6.0` |
| `npm pack` (RELEASE) | Version was **not** set (only ran for `PRE_RELEASE`) | `npm pack` produced `db-ux-mcp-server-4.6.0.tgz` |
| `npm publish` | Looked for `db-ux-mcp-server-4.6.1.tgz` (hardcoded from `VALID_SEMVER_VERSION`) | **ENOENT** — filename mismatch |

## Types of changes

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/fix-pipeline-mcp-publish>

### DIFF
--- a/.changeset/fallback-icon-base64.md
+++ b/.changeset/fallback-icon-base64.md
@@ -1,5 +1,0 @@
----
-"@db-ux/core-foundations": patch
----
-
-refactor: replaced external fallback icon font as inline base64 to ensure it loads with CSS even when the font file cannot be fetched, preventing layout shifts.
--- a/.changeset/fix-icon-inline-size.md
+++ b/.changeset/fix-icon-inline-size.md
@@ -1,5 +1,0 @@
----
-"@db-ux/core-foundations": patch
----
-
-fix: ensure that the icon fallback is displayed as a rectangle
--- a/output/angular/CHANGELOG.md
+++ b/output/angular/CHANGELOG.md
@@ -1,5 +1,9 @@
 # @db-ux/ngx-core-components
 
+## 4.10.2
+
+_version bump_
+
 ## 4.10.1
 
 ### Patch Changes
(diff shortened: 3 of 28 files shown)


scripts/github/publish-npm.js
```diff
@@ -39,40 +39,51 @@ const packages = [
...
+const STANDALONE_PACKAGES = new Set([
+	'core-foundations',
+	'agent-cli',
+	'core-migration',
+	'core-stylelint',
+	'core-eslint-plugin',
+	'core-postcss-plugin',
+	'core-vite-plugin',
+	'mcp-server'
+]);
```
mfranzke: I'm struggling whether we should maintain this as a static list in which we might still forget to add new entries. Those packages would be possible to get determined programmatically out of the package.json contents out of all packages.

nmerget:
Right now we ship all packages with the same version, to make sure that a user knows they belong to each other, even when there is no change. 

If we want to separate the versions, we need to use some sort of peerDependency. Imagine the mcp-server is on version 1.2.3 and our core packages on 5.4.3. If we change the documentation and the components in 6.0.0 we also need to change the mcp so it become 2.0.0 and it has to know which range of core packages it serves.

PR #6573, opened 2026-04-28
refactor: allow text and children to be used together in components

nmerget:
## Proposed changes

Previously, components used `<Show when={props.text} else={props.children}>` which made text and children mutually exclusive.
Now both can be rendered simultaneously.

## Types of changes

## Further comments

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/refactor-text-children-components>


mfranzke:
@nmerget I'm struggling whether we should still label this as a BREAKING CHANGE. In general nobody should have used both ways to define contents in parallel at the moment, but it might still lead to unexpected results.

nmerget:
> @nmerget I'm struggling whether we should still label this as a BREAKING CHANGE. In general nobody should have used both ways to define contents in parallel at the moment, but it might still lead to unexpected results.

I thought it is a breaking change as well. But you could also argue that it is a feature ^.^

.changeset/text-and-children-together.md
```diff
@@ -0,0 +1,9 @@
+---
+"@db-ux/core-components": minor
```
mfranzke:
```suggestion
```

We don't need this entry, as this PR is purely about JS template rendering, isn't it ?

packages/components/src/components/tooltip/tooltip.lite.tsx
```diff
@@ -171,6 +172,7 @@ export default function DBTooltip(props: DBTooltipProps) {
 			onClick={(event: ClickEvent<HTMLElement>) =>
 				state.handleClick(event)
 			}>
+			<Show when={props.text}>{props.text}</Show>
```
mfranzke, line 175: @nmerget shouldn't we mention the existance of the `text` property in this component? Or do we want to keep this filechange as a "bugfix" ?

packages/components/src/components/tag/tag.lite.tsx
```diff
@@ -52,9 +52,8 @@ export default function DBTag(props: DBTagProps) {
 			data-overflow={getBooleanAsString(props.overflow)}>
 			<Slot name="content" />
 
-			{props.children}
-
 			<Show when={props.text}>{props.text}</Show>
+			{props.children}
```
mfranzke, line 55: @nmerget changing the order might be relevant to tell the users in our changesets, shouldn't we ?

packages/components/src/components/button/button.lite.tsx
```diff
@@ -51,9 +51,8 @@ export default function DBButton(props: DBButtonProps) {
 			name={props.name}
 			form={props.form}
 			value={props.value}>
-			<Show when={props.text} else={props.children}>
-				{props.text}
-			</Show>
+			<Show when={props.text}>{props.text}</Show>
+			{props.children}
```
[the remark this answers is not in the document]
nmerget: Text and Children should not be used together. I'll update the comment for the text property to make this clear

packages/components/src/components/tag/tag.lite.tsx
nmerget: update the changeset

packages/components/src/components/tooltip/tooltip.lite.tsx
nmerget: add it to changeset

packages/components/src/components/navigation-item/navigation-item.lite.tsx
```diff
@@ -115,9 +115,8 @@ export default function DBNavigationItem(props: DBNavigationItemProps) {
 			data-wrap={getBooleanAsString(props.wrap)}
 			aria-disabled={getBooleanAsString(props.disabled)}>
 			<Show when={!state.hasSubNavigation}>
-				<Show when={props.text} else={props.children}>
-					{props.text}
-				</Show>
+				<Show when={props.text}>{props.text}</Show>
```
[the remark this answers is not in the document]
nmerget: reverted it

packages/components/src/components/brand/brand.lite.tsx
```diff
@@ -26,9 +26,8 @@ export default function DBBrand(props: DBBrandProps) {
 			data-show-icon={getBooleanAsString(props.showIcon)}
 			id={props.id ?? props.propOverrides?.id}
 			class={cls('db-brand', props.className)}>
-			<Show when={props.text} else={props.children}>
-				{props.text}
-			</Show>
+			<Show when={props.text}>{props.text}</Show>
+			{props.children}
```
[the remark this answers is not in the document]
nmerget: updated

michaelmkraus:
We have the same pattern (`when={props.text} else={props.children}`) also in further components.

- `navigation-item.lite.tsx` (:118, :134)
- `custom-select-list-item.lite.tsx` (:104)

Do we need to adapt them too?

nmerget:
> We have the same pattern (`when={props.text} else={props.children}`) also in further components.
> 
> * `navigation-item.lite.tsx` (:118, :134)
> * `custom-select-list-item.lite.tsx` (:104)
> 
> Do we need to adapt them too?

`navigation-item` was on purpose: https://github.com/db-ux-design-system/core-web/pull/6573#discussion_r3177879520

`custom-select-list-item.lite.tsx` is missing yes

.changeset/text-and-children-together.md
```diff
@@ -0,0 +1,14 @@
...
+"@db-ux/wc-core-components": minor
+---
+
+refactor: allow `text` property and components children to be used together in components. Previously, `text` and children were mutually exclusive.
+
+feat: add `text` property to components that previously only supported children for their label, e.g. `DBTooltip`.
+
+refactor: `label` and `children` can now be used together in `DBRadio`, `DBCheckbox`, and `DBSwitch`. Previously they were mutually exclusive.
+
+refactor: changed render order of `children` and `text` in `DBTag` to align with design (children rendered before text).
```
mfranzke: these would end up like this in a changelog. Does changesets actually correctly format this as a list, or would it end of crashing our changelogs ?

.changeset/text-and-children-together.md
```diff
@@ -0,0 +1,14 @@
...
+"@db-ux/react-core-components": minor
+"@db-ux/v-core-components": minor
+"@db-ux/wc-core-components": minor
+---
+
+refactor: allow `text` property and components children to be used together in components. Previously, `text` and children were mutually exclusive.
+
+feat: add `text` property to components that previously only supported children for their label, e.g. `DBTooltip`.
+
+refactor: `label` and `children` can now be used together in `DBRadio`, `DBCheckbox`, and `DBSwitch`. Previously they were mutually exclusive.
```
mfranzke:
```suggestion
refactor(`DBRadio`, `DBCheckbox`, `DBSwitch`): `label` and `children` can now be used together. Previously they were mutually exclusive.
```

mention the components in scope.

.changeset/text-and-children-together.md
```diff
@@ -0,0 +1,14 @@
+---
+"@db-ux/ngx-core-components": minor
+"@db-ux/react-core-components": minor
+"@db-ux/v-core-components": minor
+"@db-ux/wc-core-components": minor
+---
+
+refactor: allow `text` property and components children to be used together in components. Previously, `text` and children were mutually exclusive.
```
mfranzke:
```suggestion
refactor: allow `text` property and components children to be used in parallel in components. Previously, `text` and children were mutually exclusive.
```

.changeset/text-and-children-together.md
```diff
@@ -0,0 +1,14 @@
...
+
+feat: add `text` property to components that previously only supported children for their label, e.g. `DBTooltip`.
```
mfranzke:
```suggestion
feat: add `text` property to components that previously only supported children for their label, like e.g. `DBTooltip`.
```

.changeset/text-and-children-together.md
```diff
@@ -0,0 +1,14 @@
+---
+"@db-ux/ngx-core-components": minor
+"@db-ux/react-core-components": minor
+"@db-ux/v-core-components": minor
+"@db-ux/wc-core-components": minor
+---
+
+feat: add `text` property to components that previously only supported children for their label, like e.g. `DBTooltip`.
+
+refactor: allow `text` property and components children to be used together in components. Previously, `text` and children were mutually exclusive.
```
mfranzke:
```suggestion
refactor: allow `text` property and components children to be used in parallel in components. Previously, `text` and children were mutually exclusive.
```

.changeset/text-and-children-together.md
```diff
@@ -0,0 +1,14 @@
...
+
+refactor(`DBRadio`, `DBCheckbox`, `DBSwitch`): `label` and `children` can now be used together. Previously they were mutually exclusive.
+
+refactor(`DBTag`): changed render order of `children` and `text` to align with design (children rendered before text).
```
mfranzke:
```suggestion
```

The order, that has been changed in the meantime, seems to have been reverted lately. In that case we should remove this changesets entry.

.changeset/text-and-children-together.md
```diff
@@ -0,0 +1,14 @@
...
+"@db-ux/react-core-components": minor
+"@db-ux/v-core-components": minor
+"@db-ux/wc-core-components": minor
+---
+
+feat: add `text` property to components that previously only supported children for their label, like e.g. `DBTooltip`.
+
+refactor: allow `text` property and components children to be used together in components. Previously, `text` and children were mutually exclusive.
+
+refactor(`DBRadio`, `DBCheckbox`, `DBSwitch`): `label` and `children` can now be used together. Previously they were mutually exclusive.
```
mfranzke, line 12: these would end up like this in a changelog. Does changesets actually correctly format this as a list, or would it end of crashing our changelogs ?

.changeset/text-and-children-together.md
nmerget: Good catch, it was missing somehow. I updated the tag.lite.tsx

.changeset/text-and-children-together.md
nmerget: image

mfranzke: I could have tested this myself. thanks.

.changeset/text-and-children-together.md
```diff
@@ -0,0 +1,14 @@
...
+"@db-ux/wc-core-components": minor
+---
+
+feat: add `text` property to components that previously only supported children for their label, like e.g. `DBTooltip`.
+
+refactor: allow `text` property and components children to be used in parallel in components. Previously, `text` and children were mutually exclusive.
+
+refactor(`DBRadio`, `DBCheckbox`, `DBSwitch`): `label` and `children` can now be used together. Previously they were mutually exclusive.
+
+refactor(`DBTag`): changed render order of `children` and `text` to align with design (children rendered before text).
```
mfranzke:
```suggestion
fix(`DBTag`): changed render order of `children` and `text` to align with design (children rendered before text).
```

mfranzke (approved):

PR #6583, opened 2026-04-29
fix(dialog): missing selector

### LINKED ISSUE #6580: Dialog - missing CSS

- [ ] All
- [ ] HTML
- [x] React
- [ ] Angular
- [ ] Vue
- [ ] Web components
- [ ] Power Apps

### Reproduction case

After migrating `@db-ux/react-core-components` from v4.5.4 to v4.6.1, we noticed our dialogs to be wrongly positioned. After inspecting the issue, it seems the following CSS has been removed from the version 6.4.1:

Image

I've checked tag notes of published versions and the migration folder https://github.com/db-ux-design-system/core-web/tree/main/packages/migration but I couldn't find any mentions of removing CSS parts responsible for modals.

Was this mentioned somewhere else maybe?

### Expected Behaviour

Changing element's position from `fixed` to `absolute` is considered a breaking change, and it should be clearly documented in order to prevent broken UIs.

### Screenshots

_No response_

### Browser version

None

_No response_

None

_No response_

_No response_

mfranzke:
## Proposed changes

Revert "fix(drawer): resolve mobile view transform issue for drawer" (https://github.com/db-ux-design-system/core-web/pull/6460) / commit 17629755c0c249808c0bb69686f167677d4181ee, as we should make a change like this with the next BREAKING CHANGE release (even though that the usage of the `dialog` selector and implications might not have been intended, it's still unexpected and disturbing for our users). This might even also get combined with a DBDialog component if it fits timewise.

resolves https://github.com/db-ux-design-system/core-web/issues/6580

## Types of changes

## Further comments

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/6580-dialog---missing-css>


nmerget:
> Revert "fix(drawer): resolve mobile view transform issue for drawer" (https://github.com/db-ux-design-system/core-web/pull/6460) / commit https://github.com/db-ux-design-system/core-web/commit/17629755c0c249808c0bb69686f167677d4181ee, as we should make a change like this with the next BREAKING CHANGE release (even though that the usage of the dialog selector and implications might not have been intended, it's still unexpected and disturbing for our users). This might even also get combined with a DBDialog component if it fits timewise.

I disagree. It was a bug in the first place that we used a global selector like `dialog` to style our DBDrawer.

We have a clear way to detect what a breaking change is. When the API of a component changes, not for internal styles. Even then the styles we can test didn't change at all.

mfranzke:
> > Revert "fix(drawer): resolve mobile view transform issue for drawer" (#6460) / commit [1762975](https://github.com/db-ux-design-system/core-web/commit/17629755c0c249808c0bb69686f167677d4181ee), as we should make a change like this with the next BREAKING CHANGE release (even though that the usage of the dialog selector and implications might not have been intended, it's still unexpected and disturbing for our users). This might even also get combined with a DBDialog component if it fits timewise.
> 
> I disagree. It was a bug in the first place that we used a global selector like `dialog` to style our DBDrawer.
> 
> We have a clear way to detect what a breaking change is. When the API of a component changes, not for internal styles. Even then the styles we can test didn't change at all.

All of this is correct. But the differentiation of patch, minor and major releases doesn't relate to who caused the bug or did some unintended or incorrect implementation in the first place, but how we'd like to fix it and by that the implications to the users.

These styles aren't internal at all, but global styles. Even if they would just be internal styles, our rules only relate to the release category (patch, minor, major), but doesn't imply that we wouldn't mention a change in our changelog.

It's not even clear whether we have to "fix" this, as we provide other relevant global styles as well already, but we actually need to inform our users either on necessary changes or impacts. This didn't happen in the current case, as reported by two users, so there's no way around to revert this and do it right afterwards (which includes a clear communication, documentation/issue description, etc.).

PR #6620, opened 2026-05-05
chore: selector simplification

mfranzke:
## Proposed changes

We're targeting a hyphen separated value here, so we don't need to select for `*`, but only [`|`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/Attribute_selectors#attrvalue_3); expected values:

- `corner-top-left`
- `corner-top-right`
- `corner-center-left`
- `corner-center-right`
- `corner-bottom-left`
- `corner-bottom-right`

## Types of changes

## Further comments

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/chore-selector-simplification>

### DIFF
--- a/packages/components/src/components/badge/badge.scss
+++ b/packages/components/src/components/badge/badge.scss
@@ -21,19 +21,19 @@
 		inset-inline-start: 0;
 	}
 
-	&[data-placement*="top"] {
+	&[data-placement|="top"] {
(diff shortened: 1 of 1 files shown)


mfranzke:
I didn't remember that selector correctly. We won't proceed with this.

PR #6722, opened 2026-05-18
fix: publish for releases

nmerget:
## Proposed changes

fix the `publishpackages` job, that was missing a `!cancelled()` condition part.

final release pipeline test-run: https://github.com/db-ux-design-system/core-web/actions/runs/26092684046

> [!WARNING]
> We need to release this PR with release 4.9.0, as someone has already merged some 4.9.0 parts into main branch, and these code changes would elsewhere get released as patch release 4.8.1 instead even already.

## Types of changes

## Further comments

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/fix-publish>


.github/actions/pnpm-cache/action.yml
```diff
@@ -9,7 +9,9 @@ runs:
   using: "composite"
   steps:
     - name: 🆙 Setup pnpm
-      uses: pnpm/action-setup@v4
+      uses: pnpm/action-setup@v6
+      with:
+        version: 11
```
mfranzke:
```suggestion
        run_install: false
```

adapted this by other steps. Defining the version here is not necessary, compare to:

> Optional when there is a [`packageManager` field in the `package.json`](https://nodejs.org/api/corepack.html).

source: https://github.com/pnpm/action-setup#version

.github/actions/pnpm-cache/action.yml
```diff
@@ -9,7 +9,7 @@ runs:
   using: "composite"
   steps:
     - name: 🆙 Setup pnpm
-      uses: pnpm/action-setup@v4
+      uses: pnpm/action-setup@v6
```
mfranzke:
```suggestion
      uses: pnpm/action-setup@v4
```

do we actually need this update for this fix? elsewhere it would be part of https://github.com/db-ux-design-system/core-web/pull/6711 already.

nmerget: That's what I try to find out

.github/workflows/default.yml
```diff
@@ -543,7 +544,7 @@ jobs:
     uses: ./.github/workflows/03-preview-url-pr-description.yml
 
   publishpackages:
-    if: ${{ github.event_name == 'release' }}
+    if: ${{ !cancelled() && github.event_name == 'release' }}
```
mfranzke: @nmerget now this publish would also happen if any of the previous steps would fail (un)expectedly, [compare to the current flacky screen-reader tests and the resulting `checks-done`](https://github.com/db-ux-design-system/core-web/actions/runs/26051296796/attempts/1). This has been implicitly checked for previously without the `!cancelled()`. And don't get me wrong, integrating `!cancelled()` is the way to go, but don't we need to additionally check for failures and in that case prevent some of our steps to run like this one ?

nmerget: `!cancelled()` just says we should run it if it was not canceled by any user or something else.

The `needs` are still there so `checks-done` has to be valid.

mfranzke: > Use `jobs.<job_id>.needs` to identify any jobs that must complete **successfully** before this job will run. It can be a string or array of strings. If a job fails or is skipped, all jobs that need it are skipped **unless the jobs use a conditional expression** that causes the job to continue

That assumption was actually correct and we've integrated an explicit check for the depending workflow steps.

.github/workflows/03-publish-packages.yml
mfranzke:
```suggestion
    name: Publish (or dry-run) latest package versions to npmjs.com registry
```

we should adapt this text with the newest change (doing a dry-run) and correct GitHub packages to npmjs.com

PR #6725, opened 2026-05-19
fix: flacky screen-reader tests by retrying

mfranzke:
## Proposed changes

Currently, we are experiencing unreliable screen-reader tests, with at least one of those steps failing on the first attempt. After rerunning the test (up to two more times in extreme situations), the problem disappears and the expected result is reported. The downside of this approach is that mismatches might be expected due to changes, but unexpected failures of regular jobs outnumber them by far.

If it succeeds, we'd like to proceed with the native approach: https://github.com/db-ux-design-system/core-web/pull/6724

## Types of changes

## Further comments

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/fix-flacky-screen-reader-tests-by-retrying>

### DIFF
--- a/.github/workflows/02-e2e-screen-reader.yml
+++ b/.github/workflows/02-e2e-screen-reader.yml
@@ -53,11 +53,14 @@ jobs:
           path: build-showcases/react-showcase
 
       - name: 👩‍🔬 Test showcase with Playwright 🎭
-        shell: bash
         env:
           showcase: react-showcase
-        run: |
-          pnpm --filter react-showcase run test-sr:${{ env.OS }} --shard=${{ matrix.shardIndex }}/${{ matrix.shardTotal }}
+        uses: nick-fields/retry@v4
+        with:
+          timeout_seconds: 15
+          max_attempts: 3
+          shell: bash
+          command: pnpm --filter react-showcase run test-sr:${{ env.OS }} --shard=${{ matrix.shardIndex }}/${{ matrix.shardTotal }}
 
       - name: 🆙 Upload test results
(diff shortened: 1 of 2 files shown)


.github/workflows/02-e2e-screen-reader.yml
```diff
@@ -53,11 +53,14 @@ jobs:
           path: build-showcases/react-showcase
 
       - name: 👩‍🔬 Test showcase with Playwright 🎭
-        shell: bash
         env:
           showcase: react-showcase
-        run: |
-          pnpm --filter react-showcase run test-sr:${{ env.OS }} --shard=${{ matrix.shardIndex }}/${{ matrix.shardTotal }}
+        uses: nick-fields/retry@v4
```
nmerget, line 58: We shouldn't add such a action. Playwright has already a retry. We should use for main `pnpm --filter react-showcase run test-sr:${{ env.OS }} --shard=${{ matrix.shardIndex }}/${{ matrix.shardTotal }} --retries=5` and for release `pnpm --filter react-showcase run test-sr:${{ env.OS }} --shard=${{ matrix.shardIndex }}/${{ matrix.shardTotal }} --retries=10`. For a PR the retry already is set to 2.

mfranzke: Sounds good to me. AI even already came up with the same approach in the other PR.

mfranzke:
https://github.com/db-ux-design-system/core-web/pull/6724 looks promising, so we don't proceed with this one.

PR #6737, opened 2026-05-19
refactor: update accessibility-checker integration to use local engine

nmerget:
## Proposed changes

Some [tests](https://github.com/db-ux-design-system/core-web/actions/runs/26096796761/job/76749857968) fail because of a mismatch between playwright page and IBM a11y checkers selenium api

## Types of changes

## Further comments

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/fix-ibm-a11y-checker>


mfranzke:
@nmerget Is this supposed to be a temporary fix?

nmerget:
> @nmerget Is this supposed to be a temporary fix?

No it should fix the issue. I need to validate if it works. But right now the checker is broken.

So what it does is that we skip the boilerplate (which used selemium - which caused the issue) and uses the IBM script directly. Benefits: no external call against a CDN, and it is way faster because it runs in playwright directly.

.github/workflows/02-e2e-showcases.yml
```diff
@@ -104,8 +104,7 @@ jobs:
       - name: 👩🔬 Test a11y-checker with Playwright 🎭
         shell: bash
         if: |
-          github.event.pull_request != null &&
-          inputs.test-ally == 'true'
```
mfranzke, line 108: @nmerget have you removed this condition part intentionally, or only temporarily for testing purpose? I wouldn't have expected a change to this condition based on the PRs description (and the other code changes in this PR).

mfranzke (dismissed):

.github/workflows/02-e2e-showcases.yml
nmerget: No it is intentional.

a) That the tests realy runs
b) Because we don'T spawn selenium anymore, it'S fast enought to run in every pipeline

.github/CODEOWNERS
```diff
@@ -1,3 +1,3 @@
 # Global Repository owners
 # see https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners for details of syntax
-# * @mfranzke @nmerget @michaelmkraus @bruno-sch
+# * @mfranzke @nmerget @michaelmkraus
```
mfranzke:
```suggestion
# * @mfranzke @nmerget @michaelmkraus @bruno-sch 
```

how is this related to updating an accessibility-checker integration?

.config/cspell.config.ts
```diff
@@ -45,6 +45,7 @@ export default defineConfig({
 		'packages/mcp-server/src/manifest.json',
 		'storybooks/angular-storybook/src/components/documentation.json',
 		'packages/foundations/src/all-icons.ts',
+		'.github/CODEOWNERS',
```
mfranzke:
```suggestion
```

this file doesn't produce a warning or error, neither in the pipeline nor on localhost.

nmerget: It does

mfranzke: no

mfranzke: let's check

mfranzke (approved):

PR #6749, opened 2026-05-20
fix: resolve `ERR_UNSUPPORTED_DIR_IMPORT` in strict ESM environments

### LINKED ISSUE #6720: @db-ux/react-core-components@4.8.0: ESM build uses directory imports without explicit file extensions

- [ ] All
- [ ] HTML
- [x] React
- [ ] Angular
- [ ] Vue
- [ ] Web components
- [ ] Power Apps

### Description:
 
The ESM build of @db-ux/react-core-components@4.8.0 uses relative imports without explicit file extensions in dist/index.js as well as in all component files under dist/components/:
 
// dist/index.js
export * from './components/accordion';       // ❌ missing extension
export * from './components/accordion/model'; // ❌ missing extension
 
According to the Node.js ESM specification, explicit file extensions are mandatory for relative imports in ES modules. The correct form would be:
 
export * from './components/accordion/index.js'; // ✅
export * from './components/accordion/model.js'; // ✅
 
### Impact:
 
When using Vitest 4, all tests that import components from this package fail:
 
Error: Directory import '.../node_modules/@db-ux/react-core-components/dist/components/accordion'
is not supported resolving ES modules imported from .../dist/index.js
ERR_UNSUPPORTED_DIR_IMPORT
 
Vitest's ModuleRunner (based on vite/module-runner) loads ESM packages strictly according to the specification and rejects directory imports. The issue occurs with all pool types (forks, vmForks), as Vitest consistently uses its own resolver.
 
### Workaround (our side):
 
In vitest.config.ts, setting ssr.noExternal forces Vite to run the package through its transform pipeline, which resolves directory imports internally:
ssr: {
  noExternal: ["@db-ux/react-core-components"],
}

### Expected Behaviour

The above workaround should not be neccessary.
 
### Request:
 
Please update the ESM build so that all relative imports in dist/ include explicit .js file extensions — in compliance with the ESM specification and the "type": "module" declared in the package's package.json.

### Screenshots

_No response_

### Browser version

Edge

_No response_

DB Infra GO

UJK / C2S

michaelmkraus:
## Proposed changes

The ESM build of `@db-ux/react-core-components` previously generated relative imports without explicit file extensions (e.g., `export * from './components/accordion';`). According to the Node.js ESM specification, explicit extensions are mandatory. This caused tools with strict ESM resolution, such as Vitest 4, to fail with an `ERR_UNSUPPORTED_DIR_IMPORT` error.

Added `tsc-esm-fix` as a post-build step for the React output. The build script now automatically appends the correct `.js` and `/index.js` extensions to all relative imports inside the `dist/` directory after `tsc` finishes.

resolves #6720 

## Types of changes

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/fix-react-esm-import-extensions>


nmerget:
@michaelmkraus I wouldn't recommend to use thise package. It has 80k downloads and was published 2 years ago.

In fact we need to change it for react and vue output.

I asked the AI and recommend to use `vite + tsc`:

````
Yes — you can handle it with a bundler, and that’s a common way to avoid tsc-esm-fix.
 Use tsc only for types (.d.ts)
 Use a bundler for JS emit (so imports are ESM-valid)
 Keep React and peer deps external
For this package, esbuild or Vite/Rollup both work:
esbuild: fast, simple; good if you’re okay with a bundler-driven JS output.
Vite (Rollup): better when you need library-style control (preserveModules, multiple entrypoints, stable output structure).
Practical pattern
tsc --emitDeclarationOnly for declarations.
Bundler builds dist/**/*.js with valid ESM imports/extensions.
Remove fix-esm (tsc-esm-fix) from scripts/devDeps.
Caveat for your repo
If you need the current “one file per source module” structure under dist/components/**, prefer Rollup/Vite with preserveModules. Pure esbuild can do multi-entry builds, but Rollup is usually cleaner for library packaging layout.
If you want, I can draft the exact output/react/package.json script changes plus a minimal vite.config.ts or esbuild script tailored to your current folder structure.
````

File: output/react/package.json [hunk 1]
```diff
@@ -37,12 +38,14 @@
...
     "@types/react": "19.2.14",
+    "@vitejs/plugin-react": "4.5.0",
     "cpr": "3.0.1",
     "npm-run-all2": "8.0.4",
     "playwright": "1.60.0",
     "react": "19.2.6",
     "react-dom": "19.2.6",
-    "typescript": "5.9.3"
+    "typescript": "5.9.3",
+    "vite": "^6.4.0"
```
nmerget: We should avoid `^` and align it with the vite version of the other packages

output/react/package.json
```diff
@@ -17,7 +17,9 @@
     "dist/"
   ],
   "scripts": {
-    "build": "npm-run-all tsc --parallel mv:*",
+    "build": "npm-run-all --parallel build:types build:js --serial mv:*",
```
nmerget:
```suggestion
    "build": "npm-run-all --parallel build:types build:js --parallel mv:*",
```

Can't we use `--parallel` for the move as well?

nmerget:
@michaelmkraus we should make a pre-release and test it in the examples, and with vitest

michaelmkraus:
Pre-Release: [v4.10.0-esm](https://www.npmjs.com/package/@db-ux/core-components/v/4.10.0-esm-94516f3)

nmerget:
I don't like to use such an outdated dependency. We need to use the same for vue and stencil outputs as well and we should fix it during Mitosis generation:


## Summary: ESM Extension Fix via Mitosis `build.post` Plugin

The key insight is that `build.post` receives **both** `componentFiles` and `nonComponentFiles` — meaning everything Mitosis outputs, including `index.ts`, `model.ts`, `utils/`, etc.

This allows writing a `build.post` plugin that appends `.js` extensions to all relative imports in generated files.

### The resulting flow:

1. Mitosis generates code → `build.post` fixes all imports with `.js` extensions
2. `tsc` sees correct imports → emits correct `.d.ts` files with extensions
3. `tsc` also emits `.js` directly (no more `emitDeclarationOnly` needed)
4. Vite, Rollup, `tsc-esm-fix` — all become unnecessary

### Implementation plan:

1. **New plugin** `esm-extensions.cjs` with `build.post` hook
2. **Integrate plugin** into the React (and Vue + Stencil?) Mitosis config
3. **Adjust `tsconfig.json`**: remove `emitDeclarationOnly`, upgrade `module`/`moduleResolution`
4. **Clean up `package.json`**: remove Vite, `tsc-esm-fix`, `@vitejs/plugin-react` from devDeps and scripts
5. **Delete `vite.config.ts`**

michaelmkraus:
Implemented a lighter fix without `vite` / `tsc-esm-fix` by normalizing ESM paths in `packages/components/scripts/post-build/react.ts`.

### Changes
- Added post-build normalization for relative imports/exports:
  - directory imports -> `/index.js`
  - file imports -> `.js`
- Updated `output/react/tsconfig.json` to `moduleResolution: "bundler"` + `module: "preserve"`, and removed `emitDeclarationOnly`
- Simplified `output/react/package.json` build to `tsc` + copy
- Removed `output/react/vite.config.ts`
- Updated `pnpm-lock.yaml`

### Validation
- `pnpm --filter @db-ux/core-components run build-components`
- `pnpm --filter @db-ux/react-core-components run build`
- `pnpm --filter @db-ux/core-components run test`
- `node -e "import('./output/react/dist/index.js')..."` -> `esm-import-ok`

Result: generated `dist` now uses explicit ESM paths and no longer triggers `ERR_UNSUPPORTED_DIR_IMPORT`.

packages/components/scripts/post-build/react.ts
```diff
@@ -89,6 +90,97 @@ const overwriteFragmentMap = (input: string) => {
 		.join('\n');
 };
 
+const collectSourceFiles = (root: string): string[] => {
```
nmerget: We want to avoid adding new `post-build` scripts this was just a workaround. The plugin system from Mitosis should handle everything in the future as I exmplained in [this](https://github.com/db-ux-design-system/core-web/pull/6749#issuecomment-4645928208).

With the plugin system you can easily append every import starting with  a `./` or `..` by `.ts`.

Also the big advantage of the plugin: You can use it for vue and stencil as well. Which has the same problem

output/react/tsconfig.json
```diff
@@ -8,14 +8,13 @@
 		"experimentalDecorators": true,
 		"esModuleInterop": true,
 		"lib": ["dom", "es2022"],
-		"module": "es2015",
-		"moduleResolution": "node",
+		"module": "preserve",
+		"moduleResolution": "bundler",
```
nmerget: If everything works like expected we could use `node16` here instead right?

packages/components/scripts/post-build/react.ts
michaelmkraus: Implemented the Mitosis solution

output/react/tsconfig.json
nmerget: Why did you resolve it @michaelmkraus?

output/react/tsconfig.json
```diff
@@ -8,14 +8,13 @@
 		"experimentalDecorators": true,
 		"esModuleInterop": true,
 		"lib": ["dom", "es2022"],
-		"module": "es2015",
-		"moduleResolution": "node",
+		"module": "preserve",
```
nmerget: **Suggestion: Consider using `"module": "node16"` / `"moduleResolution": "node16"` instead of `"preserve"` / `"bundler"`.**

Since `package.json` already declares `"type": "module"` and the `esm-extensions` plugin guarantees all relative imports have explicit `.js` extensions before `tsc` runs, you could use the stricter `node16` resolution. This would give you compile-time validation that all imports are ESM-compliant (rather than only discovering issues at runtime).

The change would be:
```jsonc
{
    "module": "node16",
    "moduleResolution": "node16",
    "jsx": "react-jsx",  // required: legacy "react" mode isn't compatible with node16 ESM
    "target": "es2022"   // node16 pairs best with es2020+
}
```

The benefit: if the Mitosis plugin ever misses an import, `tsc` will catch it immediately rather than failing silently at runtime. The current `"bundler"` mode is permissive and passes extensionless specifiers through without complaint, which means broken imports can slip through to consumers.

One caveat: this requires the `esm-extensions` plugin to run **before** `tsc` (which is already the case since the plugin runs in Mitosis `build.post`, which fires before the `tsc` script). Just verify the ordering holds.

packages/components/configs/plugins/esm-extensions.cjs
```diff
@@ -0,0 +1,132 @@
...
+			fs.existsSync(path.join(absolute, 'index.ts')) ||
+			fs.existsSync(path.join(absolute, 'index.tsx'))
+		) {
+			return `${importPath}/index.js`;
+		}
+	}
+
+	// Is it a file (.ts or .tsx)?
+	if (fs.existsSync(`${absolute}.ts`) || fs.existsSync(`${absolute}.tsx`)) {
+		return `${importPath}.js`;
```
nmerget, line 45: **Suggestion: Log a warning when a relative import can't be resolved.**

If `resolveEsmPath` falls through to the final `return importPath` (no directory with index, no `.ts`/`.tsx` file found), it silently leaves the import unchanged. This could lead to hard-to-debug runtime errors if a file is missing or the structure changes.

Consider adding a `console.warn`:
```js
// At the end, before the final return:
console.warn(`[esm-extensions] Could not resolve: "${importPath}" from "${fromFile}"`);
return importPath;
```

This way broken imports surface during the build rather than silently propagating to consumers.

packages/components/configs/plugins/esm-extensions.cjs
```diff
@@ -0,0 +1,132 @@
...
+ * @param {string} filePath - Absolute path to the file
+ */
+const fixFileImports = (filePath) => {
+	if (!fs.existsSync(filePath)) return;
+
+	const source = fs.readFileSync(filePath, 'utf-8');
+
+	const fixed = source.replace(
+		/((?:from|import)\s+['"])(\.[^'"]+)(['"])/g,
+		(match, prefix, importPath, suffix) => {
```
nmerget, line 91: **Note: This regex doesn't match side-effect imports.**

The pattern `/((?:from|import)\s+['"])(\.[^'"]+)(['"])/g` correctly handles:
- `import x from './foo'`
- `export { x } from './foo'`

But it won't match side-effect imports like `import './some-module'` (no `from` keyword, and the pattern requires `from` or `import` followed by whitespace then a quote). For this codebase that's likely fine since side-effect imports of relative `.ts` files are rare, but worth documenting as a known limitation.

packages/components/configs/plugins/esm-extensions.cjs
```diff
@@ -0,0 +1,132 @@
...
+			if (!files) return;
+
+			const allOutputFiles = [
+				...(files.componentFiles || []),
+				...(files.nonComponentFiles || [])
+			];
+
+			if (allOutputFiles.length === 0) return;
+
+			// Determine the src directory from the first file's outputDir
```
nmerget: **Minor: The plugin only processes files under `src/`.**

If there are ever output files outside `src/` (e.g., a root-level `index.ts` barrel that re-exports from `./src/...`), they won't get their imports rewritten. Currently this matches the project structure, but adding a comment here noting this assumption would help future maintainers.

```js
// NOTE: Only processes files inside src/. Root-level files (if any) are not covered.
const srcDir = path.resolve(outputDir, 'src');
```

packages/components/scripts/post-build/react.ts
```diff
@@ -137,7 +137,7 @@ export default DB${upperComponentName};`
 					from: 'import * as React from "react";',
 					to:
 						'import * as React from "react";\n ' +
-						'import { filterPassingProps, getRootProps } from "../../utils/react";\n'
+						'import { filterPassingProps, getRootProps } from "../../utils/react.js";\n'
```
nmerget, line 143: **Good: explicit `.js` extension here is correct.**

Since this import is injected by the post-build script (not processed by the `esm-extensions` plugin), the manual `.js` is necessary. Consider adding a short comment explaining *why* it's explicit here, so future maintainers don't accidentally remove it thinking the plugin handles it:

```ts
// Explicit .js extension required — this import is injected post-Mitosis
// and isn't processed by the esm-extensions plugin.
```

nmerget (commented):
Overall this is a solid approach — replacing the Vite build with direct `tsc` + a Mitosis `build.post` plugin for explicit ESM extensions is cleaner and reduces dependencies.

Key suggestions:
1. **Consider `node16` module resolution** instead of `bundler` — gives compile-time validation that all imports are ESM-compliant
2. **Add a warning log** when `resolveEsmPath` can't resolve a relative import (prevents silent runtime failures)
3. **Document the `src/`-only scope** and the side-effect import limitation
4. **Add a comment** on the manual `.js` in `post-build/react.ts` explaining why it's needed

The ESM fix itself is correct and well-structured. Nice work removing the Vite dependency entirely.

.changeset/plain-teams-juggle.md
```diff
@@ -0,0 +1,5 @@
+---
+"@db-ux/react-core-components": patch
```
nmerget:
```suggestion
"@db-ux/react-core-components": patch
"@db-ux/v-core-components": patch
"@db-ux/wc-core-components": patch
```

packages/components/configs/plugins/esm-extensions.cjs
```diff
@@ -0,0 +1,132 @@
...
+			results.push(...collectFiles(fullPath));
+		} else if (
+			/\.(ts|tsx)$/.test(entry.name) &&
+			!EXCLUDED_PATTERNS.test(entry.name)
+		) {
+			results.push(fullPath);
+		}
+	}
+	return results;
+};
```
nmerget: Why do we collect the files here?

`allOutputFiles` should be already all the files.

Furthermore we don't collect the `.vue` files with we do it like this.

We could just filder the `allOutputFiles` with `EXCLUDED_PATTERNS` and won't fetch all files again

nmerget:
**Follow-up: Vue output tsconfig should be updated for consistency**

The `esm-extensions` plugin is now applied to React, Vue, and Stencil — but only React's `tsconfig.json` was updated (`"module": "preserve"` / `"moduleResolution": "bundler"`). Vue still uses the legacy `"module": "esnext"` / `"moduleResolution": "node"`, which means:

- The plugin adds `.js` extensions to Vue output files, but `tsc` doesn't *validate* them — if the plugin ever misses an import, it'll silently pass compilation and only fail at runtime.
- The settings are inconsistent across framework outputs.

**Suggested change for `output/vue/tsconfig.json`:**

```jsonc
"module": "node16",
"moduleResolution": "node16",
"target": "es2022"
```

Or at minimum, match React's current approach:
```jsonc
"module": "preserve",
"moduleResolution": "bundler"
```

Both Vue and React outputs declare `"type": "module"` in their `package.json`, so they're in the same situation.

**Stencil** can be left as-is — its tsconfig is primarily used for type-checking within Stencil's own compiler pipeline, and changing module resolution there could conflict with Stencil internals (`"jsx": "react"` with `jsxFactory: "h"`).

packages/components/configs/plugins/esm-extensions.cjs
```diff
@@ -0,0 +1,132 @@
...
+			const srcDir = path.resolve(outputDir, 'src');
+
+			// Fix ALL .ts/.tsx files in the entire src directory
+			// This covers component barrels, shared utils, and root index.ts
+			for (const filePath of collectFiles(srcDir)) {
+				fixFileImports(filePath);
+			}
+		}
+	}
+});
```
nmerget, line 152: **Simplification: `collectFiles` is unnecessary — `allOutputFiles` already has everything you need.**

The Mitosis config uses `files: 'src/**/*.{lite.tsx,ts}'` which means all `.ts` files (barrels, models, utilities) end up in `nonComponentFiles`, and all `.lite.tsx` components end up in `componentFiles`. Together they cover the entire `src/` directory.

You could replace the recursive filesystem walk with a direct iteration:

```js
post: (targetContext, files) => {
    if (!files) return;

    const allOutputFiles = [
        ...(files.componentFiles || []),
        ...(files.nonComponentFiles || [])
    ];

    for (const file of allOutputFiles) {
        const filePath = path.resolve(file.outputDir, file.outputFilePath);
        fixFileImports(filePath);
    }
}
```

Benefits:
- No need for `collectFiles` or `EXCLUDED_PATTERNS` — Mitosis already excludes showcase/example files via the `exclude` config in `mitosis.config.cjs`
- More predictable: only touches files Mitosis actually wrote, not anything else that might be on disk
- Simpler to reason about and maintain

packages/components/configs/plugins/esm-extensions.cjs
```diff
@@ -0,0 +1,132 @@
...
+ * @param {string} fromFile   - Absolute path of the file containing the import
+ * @returns {string}
+ */
+const resolveEsmPath = (importPath, fromFile) => {
+	if (!importPath.startsWith('.')) {
+		return importPath;
+	}
+
+	if (/\.(js|mjs|cjs|json|css|scss)$/.test(importPath)) {
+		return importPath;
```
nmerget, line 27: **Bug-prone: `.vue` extension is missing from the early-return check.**

Since this plugin also runs for the Vue output, imports like `'./accordion.vue'` hit this code path. They currently work by accident — `.vue` isn't in the skip list, so the function falls through to the filesystem checks, doesn't find `accordion.vue.ts` or `accordion.vue.tsx`, and returns the path unchanged.

This is fragile and does unnecessary `fs.existsSync` calls for every `.vue` import. Add `.vue` to the extension check:

```js
if (/\.(js|mjs|cjs|json|css|scss|vue)$/.test(importPath)) {
    return importPath;
}
```

Without this, if someone ever creates a file named `accordion.vue.ts` (unlikely but possible), the plugin would incorrectly rewrite `'./accordion.vue'` → `'./accordion.vue.js'`.

packages/components/configs/plugins/esm-extensions.cjs
```diff
@@ -0,0 +1,132 @@
...
+		if (
+			fs.existsSync(path.join(absolute, 'index.ts')) ||
+			fs.existsSync(path.join(absolute, 'index.tsx'))
+		) {
+			return `${importPath}/index.js`;
+		}
+	}
+
+	// Is it a file (.ts or .tsx)?
+	if (fs.existsSync(`${absolute}.ts`) || fs.existsSync(`${absolute}.tsx`)) {
```
nmerget, line 44: **Simplification: This entire block can be reduced significantly.**

This project never uses `index.tsx` (none exist anywhere in source or output), and every directory import always has an `index.ts`. The multiple `existsSync` checks are redundant — if Mitosis generated an import pointing at a directory, it has an index file; if it points at a file, it exists as `.ts` or `.tsx` (both compile to `.js`).

Simplified version:

```js
const baseDir = path.dirname(fromFile);
const absolute = path.resolve(baseDir, importPath);

if (fs.statSync(absolute, { throwIfNoEntry: false })?.isDirectory()) {
    return `${importPath}/index.js`;
}

return `${importPath}.js`;
```

This removes 4 `existsSync` calls per import, drops the dead `index.tsx` check, and is easier to reason about. The logic is: directory → `/index.js`, anything else → `.js`.

nmerget:
**CI Fix needed: Patternhub and next-showcase builds fail due to `.js` extensions in raw source**

The `esm-extensions` plugin rewrites imports in `output/react/src/` to use `.js` extensions (e.g. `'../../utils/react.js'`). But the Patternhub and next-showcase consume these files as **raw TypeScript source** via `transpilePackages`. Webpack can't resolve `react.js` because only `react.ts` exists on disk — the `.js` files only exist after `tsc` compiles into `dist/`.

**Quick fix:** Add `resolve.extensionAlias` to the Webpack config in both Next.js configs. This tells Webpack to try `.ts`/`.tsx` when it encounters a `.js` import:

**`showcases/patternhub/next.config.js`:**
```js
webpack: (config) => {
    config.resolve.extensionAlias = {
        '.js': ['.ts', '.tsx', '.js']
    };
    return config;
},
```

**`showcases/next-showcase/next.config.js`:**
```js
webpack: (config) => {
    config.resolve.extensionAlias = {
        '.js': ['.ts', '.tsx', '.js']
    };
    return config;
},
```

This is the standard approach for consuming ESM-style TypeScript sources (with `.js` specifiers) in Webpack/Next.js.

mfranzke:
@codex please review

michaelmkraus:
## Status update: Mitosis approach implemented + open points

The ESM fix now runs entirely through the Mitosis `build.post` plugin (`configs/plugins/esm-extensions.cjs`), active for React, Vue and Stencil. The post-tsc workaround (`fix-esm-imports.cjs`) has been removed.

### The review points
- ✅ No outdated dependency / no post-build workaround anymore
- ✅ Plugin runs for React **and** Vue + Stencil
- ✅ `node16` / `node16` for the React and Vue tsconfig (compile-time validation)
- ✅ `console.warn` on unresolvable imports – including `.vue` detection, so the Vue build no longer produces false warnings
- ✅ Changeset extended with `@db-ux/v-core-components` and `@db-ux/wc-core-components`
- ✅ Unit tests for the plugin (`esm-extensions.spec.ts`)
- ✅ Removed the now-orphaned `@vitejs/plugin-react` devDependency from the React output (leftover from the dropped Vite build; lockfile updated)

### Decision that needs feedback: next-showcase

The proposed `resolve.extensionAlias` works for Patternhub (webpack). However, **next-showcase was running on Turbopack, which does not support `extensionAlias`** (verified: ~325 resolution errors). We therefore switched next-showcase to `--webpack` and changed the db-theme CSS import to the `webpack.css` variant (consistent with how Patternhub already consumes the theme CSS).

Tradeoff: no more Turbopack build speed for next-showcase. The actual root cause is that the showcases consume the **raw** `output/react/src` instead of the built package (`dist`, now valid ESM). Switching to package consumption would remove the need for `extensionAlias`, the bundler switch and the CSS change – but that's a larger change to the dev workflow (HMR on source) and, in my view, a separate follow-up issue.

→ Is the bundler switch OK for you?

### Known remaining workaround

`scripts/post-build/react.ts` injects an import with a hardcoded `.js` extension (it runs after the plugin). Noted in `AGENTS.md` as a follow-up – it should move along when this injection is migrated to a plugin.

packages/components/configs/plugins/esm-extensions.cjs
michaelmkraus: Done. The plugin now logs a warning when falling through.

packages/components/configs/plugins/esm-extensions.cjs
michaelmkraus: The regex does match side-effect imports like import './foo'. Added a JSDoc comment noting this.

packages/components/configs/plugins/esm-extensions.cjs
michaelmkraus: Done. .vue is now included in the regex with a comment.

packages/components/configs/plugins/esm-extensions.cjs
michaelmkraus: Done. It now iterates over `files.componentFiles` + `files.nonComponentFiles` directly..

packages/components/scripts/post-build/react.ts
michaelmkraus: Done. Added a comment that explains that the import is injected after the Mitosis plugin has run.

packages/components/configs/plugins/esm-extensions.cjs
michaelmkraus: The implementation needs the extra checks to handle .vue fallback and the console.warn path. "directory → 
index.js, else → .js" would break Vue imports and suppress the warning.

output/react/tsconfig.json
michaelmkraus: It's implemented now. React tsconfig uses "module": "node16" / "moduleResolution": "node16" with "jsx": "react-jsx" and "target": "es2022".

.changeset/plain-teams-juggle.md
michaelmkraus: Done. Changeset now includes `@db-ux/v-core-components` and `@db-ux/wc-core-components`.

packages/components/configs/plugins/esm-extensions.cjs
michaelmkraus: one. The plugin now uses `files.componentFiles` + `files.nonComponentFiles` directly.

mfranzke:
@codex review

mfranzke:
@nmerget, `codex` is even already fine, I had a quick lock and give a thumbs up, so it's on you for a final approval.

showcases/next-showcase/package.json
```diff
@@ -12,9 +12,9 @@
     "node": ">=18"
   },
   "scripts": {
-    "build": "pnpm run setup:browserslist && next build --turbopack",
+    "build": "pnpm run setup:browserslist && next build --webpack",
     "dev": "npm-run-all setup:browserslist --parallel next:dev open",
-    "next:dev": "next dev --turbopack --port 5050",
```
nmerget, line 17: I tought turbopack can use all webpack related things?

Did you test it with turbopack?

nmerget: Maybe this works: 

``turbopack: {
		resolveExtensions: ['.tsx', '.ts', '.jsx', '.js', '.mjs', '.json']
	},``

nmerget:
## Suggestion: Move `.vue` extension handling into `esm-extensions.cjs`

The `esm-extensions` plugin currently **intentionally leaves** Vue imports without an extension (e.g. `export { default as DBButton } from './button'`), relying on Vite/vue-tsc to resolve `.vue` natively. Then `scripts/post-build/vue.ts` runs a separate `replaceInFileSync` step to add `.vue` back:

```js
replaceInFileSync({
    files: `../../${outputFolder}/vue/src/components/${componentName}/index.ts`,
    from: `./${componentName}`,
    to: `./${componentName}.vue`
});
```

This could be consolidated directly into the `esm-extensions` plugin by changing the Vue resolution block to append `.vue` instead of returning the path untouched:

```js
// Before (current):
if (fs.existsSync(`${absolute}.vue`) || ...) {
    return importPath; // leaves './button' as-is
}

// After (proposed):
if (fs.existsSync(`${absolute}.vue`)) {
    return `${importPath}.vue`; // → './button.vue'
}
if (fs.existsSync(absolute) && fs.statSync(absolute).isDirectory() && 
    fs.existsSync(path.join(absolute, 'index.vue'))) {
    return `${importPath}/index.vue`;
}
```

Benefits:
- Removes the extra post-build `replaceInFileSync` step for `.vue` extensions
- All extension resolution logic lives in one place (`esm-extensions.cjs`)
- Works consistently on all platforms (no reliance on `replace-in-file` string matching)
- The existing test `'leaves a barrel re-export pointing at a .vue file untouched'` would change to `'appends .vue to a barrel re-export pointing at a .vue file'`

showcases/next-showcase/package.json
nmerget: It doesn'T work :/

showcases/next-showcase/styles/global.css
```diff
@@ -1,5 +1,5 @@
 @import "@db-ux/core-components/build/styles/rollup.css";
```
nmerget:
```suggestion
@import "@db-ux/core-components/build/styles/webpack.css";
```

michaelmkraus: Good point, changed.

michaelmkraus:
> ## Suggestion: Move `.vue` extension handling into `esm-extensions.cjs`
> The `esm-extensions` plugin currently **intentionally leaves** Vue imports without an extension (e.g. `export { default as DBButton } from './button'`), relying on Vite/vue-tsc to resolve `.vue` natively. Then `scripts/post-build/vue.ts` runs a separate `replaceInFileSync` step to add `.vue` back:
> 
> ```js
> replaceInFileSync({
>     files: `../../${outputFolder}/vue/src/components/${componentName}/index.ts`,
>     from: `./${componentName}`,
>     to: `./${componentName}.vue`
> });
> ```
> 
> This could be consolidated directly into the `esm-extensions` plugin by changing the Vue resolution block to append `.vue` instead of returning the path untouched:
> 
> ```js
> // Before (current):
> if (fs.existsSync(`${absolute}.vue`) || ...) {
>     return importPath; // leaves './button' as-is
> }
> 
> // After (proposed):
> if (fs.existsSync(`${absolute}.vue`)) {
>     return `${importPath}.vue`; // → './button.vue'
> }
> if (fs.existsSync(absolute) && fs.statSync(absolute).isDirectory() && 
>     fs.existsSync(path.join(absolute, 'index.vue'))) {
>     return `${importPath}/index.vue`;
> }
> ```
> 
> Benefits:
> 
> * Removes the extra post-build `replaceInFileSync` step for `.vue` extensions
> * All extension resolution logic lives in one place (`esm-extensions.cjs`)
> * Works consistently on all platforms (no reliance on `replace-in-file` string matching)
> * The existing test `'leaves a barrel re-export pointing at a .vue file untouched'` would change to `'appends .vue to a barrel re-export pointing at a .vue file'`

The esm-extensions plugin now automatically adds .vue or index.vue to Vue exports. Removed the extra replacement code from vue.ts, so all extension logic is now handled in one single place.

mfranzke (approved):

packages/components/configs/plugins/esm-extensions.spec.ts
```diff
@@ -0,0 +1,137 @@
...
+import os from 'node:os';
+import path from 'node:path';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+// The plugin is CommonJS; import its named exports for unit testing.
+const {
+	resolveEsmPath,
+	fixFileImports
+	// eslint-disable-next-line @typescript-eslint/no-require-imports
+} = require('./esm-extensions.cjs');
```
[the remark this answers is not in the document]
michaelmkraus: I verified it against the actual test run and it doesn't reproduce.
Running the exact command from the comment passes:

`pnpm --filter @db-ux/core-components test → Test Files 5 passed (5), Tests 65 passed (65)`

---

The spec is loaded through Vitest's own transform pipeline (esbuild/vite), not by Node's native ESM loader. In that transform context `require` is available in module scope regardless of the package's `"type": "module"`, so the top-level `require('./esm-extensions.cjs')` resolves fine. The Node ESM semantics this comment is based on don't apply here because Node never loads the file directly.

No functional issue, so I'll leave it as is. If we want stricter ESM hygiene in the test file later, `createRequire(import.meta.url)` would be the cosmetic alternative, but it's not required for the test to pass.

packages/components/configs/plugins/esm-extensions.cjs
```diff
@@ -0,0 +1,155 @@
...
+};
+
+/**
+ * Files matching these patterns are consumed as raw source by the showcases
+ * (Next.js transpilePackages / Turbopack) and are excluded from tsc
+ * compilation. They must NOT receive .js extensions because no compiled .js
+ * counterpart is ever emitted for them.
+ */
+const EXCLUDED_PATTERNS =
+	/\.(showcase|showcase-only|example|arg\.types)\.(ts|tsx)$/;
```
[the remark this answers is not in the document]
michaelmkraus: This doesn't happen in practice. The plugin never processes the `.vue` showcase/example files.

Those files are generated by a separate Mitosis config, `configs/mitosis.showcase.config.cjs`, which uses its own inline `options` and does not load `configs/vue/index.cjs` where the `esm-extensions` plugin is registered. So the plugin simply never runs over `*.example.vue` / `*.showcase.vue`.

Verified against the built output:

- `0` relative `.js` imports in any `.vue` file across the whole Vue output
- `98` `.vue` files still contain extensionless imports like `import { defaultTable } from "./data"` (with `data.ts` sitting right next to them). If the plugin had touched these files, those would have been rewritten to `./data.js` — they weren't.

The `EXCLUDED_PATTERNS` filter only needs to cover `.ts`/`.tsx`, because the *main* config (`mitosis.config.cjs`) pulls in showcase/example sources via its broad `src/**/*.{lite.tsx,ts}` glob — and those are `.ts`/`.tsx`, never `.vue`. So no `.vue` entry is needed here.

PR #6750, opened 2026-05-20
refactor(form components): replace hard `25ch` label width limit with customizable CSS variable

### LINKED ISSUE #6492: fix(input): align `label` and `input` elements width
Image

Beim Label fehlt nur ein Buchstabe und automatisch wird die Overflow-Darstellung verwendet, auch wenn das Inputfeld selbst noch länger ist. Kann man das besser lösen, als ggf. mit css zu überschreiben?

michaelmkraus:
## Proposed changes

The `label` element inside input components was artificially constrained by a hard `25ch` limit. This caused the label text to be truncated with an ellipsis way too early, even if the underlying input field had plenty of space left, resulting in a poor user experience.

Removed the hard character limit and replaced it with `max-inline-size: 100%`. The label now correctly inherits the available width of its wrapper element. Combined with `text-overflow: ellipsis`, the label will now only truncate if it actually exceeds the physical width of the input field.

resolves #6492 

## Types of changes

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/fix-align-label-and-input-elements-width>


packages/components/src/styles/internal/_form-components.scss
```diff
@@ -453,7 +453,7 @@ $input-valid-types:
 		@extend %db-overwrite-font-size-xs;
 
 		padding-block-end: variables.$db-spacing-fixed-xs;
-		max-inline-size: 25ch;
+		max-inline-size: 100%;
```
nmerget: We should add a variable here like:

```suggestion
		max-inline-size: var(--db-label-max-size,100%);
```

This would be a minor change, but we should write it in the changeset. That users can "revert" to the current state by using `--db-label-max-size:25ch;`. And we should add it to the framework spefic changelog as well to inform users.

michaelmkraus: Good point. I changed it.

nmerget (approved):

mfranzke:
@michaelmkraus the title of this PR will be the squashed commits message. Could you please adapt it, as it has changed from a bugfix to a feat ?

PR #6752, opened 2026-05-20
fix(custom-select): update a11y roles to use valid listbox pattern

### LINKED ISSUE #6208: DBCustomSelect list has incorrect a11y role

- [x] All
- [ ] HTML
- [ ] React
- [ ] Angular
- [ ] Vue
- [ ] Web components
- [ ] Power Apps

### Reproduction case

When selecting filters with a screen reader, the screen reader only announces "group" because `role="group"` is set in the code.

Selector: `.db-custom-select-list`

Note: This finding emerged from an accessibility audit.

### Expected Behaviour

There is the [`listbox`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/listbox_role) role which would be correct for this specific use case.
Example code:

```
<div
  role="listbox"
  aria-multiselectable="true"
  ...
```

### Screenshots

_No response_

### Browser version

None

_No response_

None

_No response_

_No response_

michaelmkraus:
## Proposed changes

During an accessibility audit, it was reported that the `DBCustomSelect` list wrapper incorrectly used `role="group"` instead of the required `role="listbox"`. This caused screen readers to announce the dropdown content incorrectly.

Implemented a valid ARIA listbox pattern:
* Updated the list container (`custom-select-list.lite.tsx`) to use `role="listbox"`.
* Dynamically applied `aria-multiselectable="true"` only when `props.multiple` is active.
* Updated all selectable list items (`custom-select-list-item.lite.tsx`) to use `role="option"` with the corresponding `aria-selected` state based on their checked status.
* Ensured that non-selectable group titles are excluded from receiving the `option` role.

resolves db-ux-design-system/core-web#6208 

## Types of changes

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/fix-custom-select-a11y-role>


packages/components/src/components/custom-select-list-item/custom-select-list-item.lite.tsx
```diff
@@ -72,6 +72,12 @@ export default function DBCustomSelectListItem(
 		<li
 			ref={_ref}
 			id={props.id ?? props.propOverrides?.id}
+			role={!props.isGroupTitle ? 'option' : undefined}
+			aria-selected={
+				!props.isGroupTitle
+					? getBooleanAsString(getBoolean(props.checked, 'checked'))
```
nmerget: We need to use an internal _check here as well because the `<input/>` can be used uncontrolled as well.

nmerget:
1.  I think it is not an accessibilty issue in this case, while the aira-snapshot is different the screen-reader (nvda) reads the same. 
2. @mfranzke did the same in a branch a made a comment that it is [much more complex](https://github.com/db-ux-design-system/core-web/pull/6213#issuecomment-3997677173)
3. I agree, if we want to do it with listbox we need to provide those [implementations](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/listbox_role#:~:text=If%20the%20current,key%20for%20this.) in JS as well. Maybe there are some additional concerns we need to check WAI Aria Patterns or other sites maybe the AI can be helpful as well.

@michaelmkraus Do you think we can solve this as a bugfix? Or is this more like a story?

packages/components/src/components/custom-select-list-item/custom-select-list-item.lite.tsx
michaelmkraus: You are right, I adapted it.

michaelmkraus:
> 3. I agree, if we want to do it with listbox we need to provide those [implementations](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/listbox_role#:~:text=If%20the%20current,key%20for%20this.) in JS as well. Maybe there are some additional concerns we need to check WAI Aria Patterns or other sites maybe the AI can be helpful as well.
> 
> @michaelmkraus Do you think we can solve this as a bugfix? Or is this more like a story?

I think this is too big for a bug-ticket. We shouldn't merge this branch yet. I will convert this PR to a draft and create a follow-up story to implement the keyboard navigation. Is this fine for you, @nmerget and @mfranzke?

michaelmkraus:
> > 3. I agree, if we want to do it with listbox we need to provide those [implementations](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/listbox_role#:~:text=If%20the%20current,key%20for%20this.) in JS as well. Maybe there are some additional concerns we need to check WAI Aria Patterns or other sites maybe the AI can be helpful as well.
> > 
> > @michaelmkraus Do you think we can solve this as a bugfix? Or is this more like a story?
> 
> I think this is too big for a bug-ticket. We shouldn't merge this branch yet. I will convert this PR to a draft and create a follow-up story to implement the keyboard navigation. Is this fine for you, @nmerget and @mfranzke?

[Follow-up story](https://github.com/db-ux-design-system/core-team/issues/2569)

michaelmkraus:
Created follow-up story

PR #6754, opened 2026-05-20
Fix notification gap no icon

michaelmkraus:
## Proposed changes

When using the `DBNotification` component with `showIcon={false}`, an unnecessary empty space (gap) appeared to the left of the text. Although the icon itself was hidden via `display: none`, the underlying CSS Grid template still reserved a `min-content` column for it, triggering the grid gap.

Updated the grid template selectors in `notification-grid-default.scss`. By appending `:not([data-show-icon="false"])` to the `[data-icon]` and `[data-semantic]` selectors, the component now falls back to a simpler grid layout without the icon column when the icon is disabled. This completely removes the unwanted left spacing and aligns the text perfectly.

resolves #6429 

## Types of changes

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/fix-notification-gap-no-icon>


.changeset/large-lights-grow.md
```diff
@@ -0,0 +1,5 @@
+---
+"@db-ux/core-components": patch
+---
+
+fix: remove empty grid gap when icon is hidden
```
nmerget:
```suggestion
---
"@db-ux/core-components": patch
"@db-ux/ngx-core-components": patch
"@db-ux/react-core-components": patch
"@db-ux/v-core-components": patch
"@db-ux/wc-core-components": patch
---

fix(notification): remove empty grid gap when icon is hidden
```

We should add also the other framworks here to nofity the users that there was a change

nmerget (approved):

PR #6872, opened 2026-06-02
feat(kiro): Create Prettier format on save hook

mfranzke:
## Proposed changes

Adds a Prettier hook to format code on save.

### Code Quality

- [x] I have reviewed my own code (self-review)

### Validation

### General

- [x] The PR title follows the conventional commits format (e.g. `feat:`, `fix:`, `chore:`)

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/feat-kiro-Create-Prettier-format-on-save-hook>

### DIFF
--- a/.kiro/hooks/prettier-format.kiro.hook
+++ b/.kiro/hooks/prettier-format.kiro.hook
@@ -1,0 +1,16 @@
+{
+  "enabled": true,
+  "name": "Prettier Format on Save",
+  "description": "Runs prettier --write on source code files when they are saved to automatically format code.",
+  "version": "1",
+  "when": {
+    "type": "fileEdited",
+    "patterns": [
+      "*"
+    ]
+  },
+  "then": {
+    "type": "runCommand",
+    "command": "prettier --write $KIRO_FILE_PATH"
+  }
+}
--- a/.kiro/hooks/prettier-format.kiro.hook
+++ b/.kiro/hooks/prettier-format.kiro.hook
@@ -1,16 +1,14 @@
 {
-  "enabled": true,
-  "name": "Prettier Format on Save",
-  "description": "Runs prettier --write on source code files when they are saved to automatically format code.",
-  "version": "1",
-  "when": {
-    "type": "fileEdited",
-    "patterns": [
-      "*"
-    ]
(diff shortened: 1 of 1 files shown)


nmerget:
Against hooks in the repo:

Tested it — extremely annoying in practice. fileEdit triggers on every single change, so in a batch of 10 edits, Kiro waits for the hook command 10 times.
The hooks are too "dumb" to be useful in a real workflow.
Redundant: Husky + Prettier + CI/CD already cover this. Kiro runs Prettier on commit anyway.
If the AI detects code style issues, it can fix them once at the end — no need for a hook on every file change.
Suggestion: Hooks belong as a personal preference on your own machine (~/.kiro/hooks/), not in the repo. Having a .kiro folder in the repo is still valuable — just without hooks.

mfranzke:
> Against hooks in the repo:
> 
> Tested it — extremely annoying in practice. fileEdit triggers on every single change, so in a batch of 10 edits, Kiro waits for the hook command 10 times. The hooks are too "dumb" to be useful in a real workflow. Redundant: Husky + Prettier + CI/CD already cover this. Kiro runs Prettier on commit anyway. If the AI detects code style issues, it can fix them once at the end — no need for a hook on every file change. Suggestion: Hooks belong as a personal preference on your own machine (~/.kiro/hooks/), not in the repo. Having a .kiro folder in the repo is still valuable — just without hooks.

what's the cost of running `prettier --write CHANGED_FILE`, if this command takes around a small amount of milliseconds?
Plus that this isn't actually not a replacement/not redundant to any other hook like the Git hooks by `husky`, it's actually a hook to prettify a file changed by Kiro exactly in that moment, so that Kiro doesn't have to do this itself, and you could easily commit on the changes after a review in a perfectly prettified file previous to the checkin without any further prettifying needed.
And we could move that simple file operation based on that hook, which that mechanism is meant for, instead of adding this operation to an instruction/rule/whatever (no need to add it to context, as it's a simple file operation based on a trigger).

But I'll try it out first of all.

mfranzke:
we'll investigate on further AI output and than decide on whether it's needed.

PR #6876, opened 2026-06-03
fix: make custom input icons render cross-browser

michaelmkraus:
## Proposed changes

Fixes cross-browser rendering of custom icons (calendar, clock) for `DBInput` components with `date`, `datetime-local`, and `time` input types. Previously, these icons were only visible in Chromium-based browsers due to a `@supports selector(::-webkit-calendar-picker-indicator)` restriction.

* Moved `date`, `datetime-local`, and `time` icon definitions out of the `@supports` block so custom icons render in all browsers (Chromium, Firefox, Safari).
* Kept `month` and `week` inside the `@supports` block, since Firefox doesn't support these input types and renders them as plain text fields — showing a calendar icon would be misleading.
* Added `pointer-events: none` on the `::after` pseudo-element to ensure custom icons don't intercept clicks meant for native picker triggers.
* Added `z-index: 1` on `::-webkit-calendar-picker-indicator` and `::-webkit-search-cancel-button` so the invisible native click target reliably sits above the custom icon layer.

resolves #5077

### Code Quality

- [ ] No hardcoded values, magic numbers, or debug code left in

### Validation

### General

- [ ] The PR title follows the conventional commits format (e.g. `feat:`, `fix:`, `chore:`), as in [Git Commit Conventions](docs/conventions.md#git-commits-conventions)

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/fix-input-icons-cross-browser>


nmerget (commented):
## Code Review

Good fix overall — moving `date`, `datetime-local`, and `time` out of the `@supports` block is the right call for cross-browser rendering. The reasoning for keeping `month`/`week` gated behind `@supports selector(::-webkit-calendar-picker-indicator)` is sound and well-documented in the comments.

### ✅ What looks good

- **Clear rationale in comments** — the distinction between universal types vs. WebKit-only types is explained directly in the code.
- **`pointer-events: none` on `::after`** — correct approach to prevent the icon layer from stealing clicks from the native picker.
- **`z-index: 1` on native indicators** — ensures the invisible native trigger sits above the icon pseudo-element for proper click handling.
- **Snapshot update** — Firefox screenshot updated to reflect the new visible icons.

### ⚠️ Items to address

1. **Missing `padding-inline-end` for `input[type="time"]`** — `date` and `datetime-local` get explicit padding via `--db-form-component-padding-inline-end`, but `time` does not. On `main`, `time` was inside the `@supports` block but also lacked its own padding rule (it relied on the general `$icon-padding` from the trailing-icon selectors above). Is the clock icon rendering without text overlap in Firefox? Please verify visually.

2. **Duplicate selector block** — Lines 70–74 and 78–80 both target `&:has(input[type="date"]), &:has(input[type="datetime-local"])`. These can be combined into a single rule block:

   ```scss
   &:has(input[type="date"]),
   &:has(input[type="datetime-local"]) {
       --db-form-component-padding-inline-end: calc(
           1lh + #{variables.$db-spacing-fixed-xs}
       );
       @include icons.set-icon("calendar", "after");
   }
   ```

3. **Broad `&::after { pointer-events: none }` scope** — This applies to *all* `.db-input` variants, including search. The search input uses `::after` for the cross/cancel icon which toggles opacity on hover/focus. If the actual click target is the native `::-webkit-search-cancel-button` on the `<input>` element (not the `::after` pseudo), it should be fine — but please confirm across browsers. If it causes issues, scope it to date/time types only:

   ```scss
   &:has(input[type="date"], input[type="datetime-local"], input[type="time"])::after {
       pointer-events: none;
   }
   ```

4. **Missing changeset** — This is a bug fix changing rendering behavior. Add a `patch` changeset for `@db-ux/core-components` (and the framework wrapper packages per the repo conventions).

### 💡 Suggestion

Add a brief inline comment near `z-index: 1` explaining it's needed so the native invisible picker trigger sits above the custom `::after` icon layer. Future maintainers will appreciate the context.

nmerget:
@michaelmkraus As I said I think it is not possible to do it in firefox:

https://github.com/db-ux-design-system/core-web/pull/6876/changes#diff-6567f35b24db4a154c6ed59c37e8232af76cdda973bdd8e06ae7cf84e5f0165c

Look at the difference, the firefox calender is still present and the other icon as well. 

Also the extra padding for the icons in chrome looked better before

michaelmkraus:
> @michaelmkraus As I said I think it is not possible to do it in firefox:
> 
> https://github.com/db-ux-design-system/core-web/pull/6876/changes#diff-6567f35b24db4a154c6ed59c37e8232af76cdda973bdd8e06ae7cf84e5f0165c
> 
> Look at the difference, the firefox calender is still present and the other icon as well.
> 
> Also the extra padding for the icons in chrome looked better before

I wrote it [in the ticket](https://github.com/db-ux-design-system/core-web/issues/5077#issuecomment-4610471518) some days ago:



I've taken a deeper look into the cross-browser issue with the date picker. It is technically not possible to hide or override the native calendar icon in Firefox via CSS.

According to the [MDN documentation](https://www.google.com/search?q=https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date%23styling), the date picker's internal UI components are rendered entirely by the browser or the OS and are encapsulated from the CSS tree. While Chrome and Safari still give access via proprietary hacks (::-webkit-calendar-picker-indicator), there are [no corresponding Mozilla extensions](https://developer.mozilla.org/en-US/docs/Web/CSS/Mozilla_Extensions). If we force our design system to display a custom icon in Firefox, we will create broken, duplicate icons and messed-up paddings (as you can see when you run the branch of this PR with my changes).

image

Therefore, we must accept that native date pickers in Firefox will keep their native system UI, or we will have to build a completely custom, JS-based date picker component in the medium term...

.changeset/fix-input-icons-cross-browser.md
```diff
@@ -0,0 +1,9 @@
+---
+"@db-ux/core-components": patch
+"@db-ux/ngx-core-components": patch
+"@db-ux/react-core-components": patch
+"@db-ux/wc-core-components": patch
+"@db-ux/v-core-components": patch
+---
+
+Fix DBInput date/time custom icon behavior across browsers and keep native picker/cancel hit targets clickable.
```
mfranzke:
```suggestion
fix(DBInput): date/time custom icon behavior across browsers and keep native picker/cancel hit targets clickable.
```

nmerget:
> > @michaelmkraus As I said I think it is not possible to do it in firefox:
> > https://github.com/db-ux-design-system/core-web/pull/6876/changes#diff-6567f35b24db4a154c6ed59c37e8232af76cdda973bdd8e06ae7cf84e5f0165c
> > Look at the difference, the firefox calender is still present and the other icon as well.
> > Also the extra padding for the icons in chrome looked better before
> 
> I wrote it [in the ticket](https://github.com/db-ux-design-system/core-web/issues/5077#issuecomment-4610471518) some days ago:
> 
> I've taken a deeper look into the cross-browser issue with the date picker. It is technically not possible to hide or override the native calendar icon in Firefox via CSS.
> 
> According to the [MDN documentation](https://www.google.com/search?q=https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date%23styling), the date picker's internal UI components are rendered entirely by the browser or the OS and are encapsulated from the CSS tree. While Chrome and Safari still give access via proprietary hacks (::-webkit-calendar-picker-indicator), there are [no corresponding Mozilla extensions](https://developer.mozilla.org/en-US/docs/Web/CSS/Mozilla_Extensions). If we force our design system to display a custom icon in Firefox, we will create broken, duplicate icons and messed-up paddings (as you can see when you run the branch of this PR with my changes).
> 
> image
> Therefore, we must accept that native date pickers in Firefox will keep their native system UI, or we will have to build a completely custom, JS-based date picker component in the medium term...

So we'll close the PR and the issue, right? It is not possible

mfranzke:
we close this PR, as we won't proceed with this solution.

PR #6877, opened 2026-06-03
refactor: Added `kiroignore`

mfranzke:
## Proposed changes

We don't want to have those files being put into context for Kiro. This should be a starting point, most likely we identify further files in the future.

I haven't managed to set this up as workspace settings automatically, so every Kiro user would need to add this ignore file to their settings individually:
image

### Code Quality

### Validation

### General

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/refactor-add-node_modules-to--kiroignore>


.kiroignore
```diff
@@ -0,0 +1,8 @@
+# Dependencies
+node_modules/
```
d-koppenhagen, line 2: Sure? My experience with Kiro is, that especially when running into problem with specific packages you rely on, Kiro (depending on the model of course) is quite good in digging into the node_modules to see how libs handle things.

mfranzke: @d-koppenhagen Yes, it's actually a recommendation from an internal workshop I participated in. Debugging dependencies sounds like an edge case to me, but it is still relevant. In that case, you might want to specifically target this folder. However, 99% of the time, you won't need anything from this folder. Otherwise, it would bloat your context when debugging problems or developing components.

nmerget: That's why I hate workshops... Did those workshop people ever used the product themselfs in a real environment?

I agree to @d-koppenhagen. I never had an issue that Kiro uses node_modules directly. But in fact it uses it if e.g. some TypeScript types are outdated. This is very helpful because some documentation on a website is often outdated and the real dependency is way better to read.

mfranzke:
@codex please review

.kiroignore
mfranzke: it's not about issues, it's about bloatet context. An alternative would be just to use `.gitignore` as an ignore setting, which is even also mentioned in the Kiro Spec: https://kiro.dev/docs/editor/kiroignore/#setting-options

Chattys conclusion:

For Kiro specifically, the official guidance leans toward **excluding `node_modules` by default**. The purpose of `.kiroignore` is not just security; it's also to keep context focused and avoid wasting tokens on large, generated, or third-party code. The Kiro documentation explicitly lists "large files or build artifacts" as a reason to use `.kiroignore`, and examples include excluding build outputs and other non-essential project content. ([kiro.dev](https://kiro.dev/docs/editor/kiroignore/))

A typical setup would be:

```gitignore
# Dependencies
node_modules/

# Build output
dist/
build/
.next/

# Generated files
coverage/
.cache/
```

### Why exclude `node_modules`?

In most cases:

* The code is not part of your application.
* It can contain hundreds of thousands of files.
* It dilutes the context window.
* It makes indexing and retrieval slower.
* The agent can usually reason about a library from its public API, types, package.json metadata, or documentation without reading its full source. ([kiro.dev](https://kiro.dev/docs/editor/kiroignore/))

### What about debugging dependencies?

Your colleague is also correct.

There are situations where you want Kiro to inspect dependency code:

* Debugging a bug inside a library.
* Understanding an undocumented behavior.
* Investigating a transpilation or bundling issue.
* Reading type definitions generated by a package.
* Contributing a fix upstream.

However, these are usually **exceptional cases**, not the default mode of working.

### Recommended approach

Rather than keeping all of `node_modules` available all the time, a good workflow is:

**Default:**

```gitignore
node_modules/
```

**When needed:**

1. Temporarily comment out the ignore rule.
2. Use explicit file references to the specific dependency files you want Kiro to inspect.
3. Re-enable the ignore afterwards.

Or, if the package is particularly important to your project, selectively unignore it:

```gitignore
node_modules/
!node_modules/my-critical-library/
```

assuming Kiro's ignore implementation follows standard gitignore semantics, which the docs state it does. ([kiro.dev](https://kiro.dev/docs/editor/kiroignore/))

### Another option: use `.gitignore`

Kiro can be configured to honor both `.gitignore` and `.kiroignore`:

```json
{
  "kiroAgent.agentIgnoreFiles": [
    ".gitignore",
    ".kiroignore"
  ]
}
```

This is actually what many teams do:

* `.gitignore` handles generated artifacts and dependencies.
* `.kiroignore` adds AI-specific exclusions such as secrets, internal documents, large datasets, etc. ([kiro.dev](https://kiro.dev/docs/editor/kiroignore/))

### My recommendation for a Node.js project

I would typically have:

```gitignore
# .kiroignore

node_modules/
dist/
build/
.next/
coverage/

.env
.env.*
!.env.example

*.log
```

and only remove or narrow the `node_modules/` exclusion when actively investigating a dependency.

That gives Kiro the highest signal-to-noise ratio during normal development while still allowing targeted dependency analysis when you actually need it. This is generally consistent with how Kiro describes `.kiroignore`: use it to keep context relevant and avoid large, unnecessary workspace content. ([kiro.dev](https://kiro.dev/docs/editor/kiroignore/))

nmerget: This is a hallucination on your side.

There is no `node_modules` found on their side.

And yes we could exclude the build outputs. And in a normal repo this makes sense. In our case we should ignore the build outputs from showcases not all outputs. Because our outputs are the generated Mitosis files and kiro should check them if there is any error in a showcase. And build outputs are not the same like dependencies.

When I use kiro I never had the problem that kiro reads `node_modules` by accident. There was no need to ignore them for now.

````
{
  "kiroAgent.agentIgnoreFiles": [
    ".gitignore",
    ".kiroignore"
  ]
}
````

^-- This is just an example what'S possbile for exclussion. And in some projects this might work perfectly. But you can't assume that this example .gitignore contains node_modules lol

nmerget: Look at the [Tip](https://kiro.dev/docs/editor/kiroignore/#setting-options:~:text=Tip,your%20project%20evolves.): 

Start with credentials and secrets—these are the highest priority files to protect. You can always expand your patterns as your project evolves.

That's what we should do. If we encounter a problem, we should add it to the kiroignore

mfranzke: again, this is still not about whether it "works" in a project. It's still about bloating the context or not.

mfranzke:
we won't do this on a project level, as it should be everybodys decision of bloating their context in the individual situation or even permantenly or not.

PR #6893, opened 2026-06-03
fix(angular-showcase): disable Vite pre-bundling to fix NG0203 dev server error

michaelmkraus:
## Proposed changes

Fixes the Angular showcase dev server (`pnpm run dev` → angular), which threw
`NG0203: The EnvironmentInjector token injection failed` on bootstrap and rendered a blank page.

**Root cause:** Vite's prebundling combined with pnpm's isolated `node_modules` produced
duplicate `@angular/core` instances at runtime. The Angular esbuild compiler bundled Angular
code inline into `main.js` while Vite simultaneously served pre-bundled chunks for the same
packages. This created two separate `EnvironmentInjector` classes, so dependency injection
failed during `bootstrapApplication`.

**Fix:** Set `prebundle: false` on the `@angular/build:dev-server` target in
`showcases/angular-showcase/angular.json`. esbuild now produces a single self-contained
bundle with no module duplication, so there is exactly one `@angular/core` instance.

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/fix-angular-showcase-ng0203-prebundle>


.changeset/fix-angular-showcase-ng0203-prebundle.md
mfranzke, line 1: I don't think that we need a changeset for a showcase, as this is not part of the released node package.

mfranzke (dismissed):

PR #6901, opened 2026-06-04
fix: tests

mfranzke:
## Proposed changes

Running `pnpm run test` from root resulted in several errors, which might be related to pnpm migration. Nevertheless we won't need `tsx` (and especially not the one initiated via `npx` without the `--no` flag) anymore, as Node could run TS natively.  
And import assertions have been replaced by import attributes. We use `with` instead of `assert` from now on.

Additionally, script tests are kept in pipeline validation (not in `lint-staged` pre-commit hooks). The `scripts/**/*` lint-staged test trigger was removed accordingly.

🔭🐙🐈 Test this branch here: <a href="https://design-system.deutschebahn.com/core-web/review/fix-tests">https://design-system.deutschebahn.com/core-web/review/fix-tests</a>

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/fix-tests>


.config/.lintstagedrc.js
```diff
@@ -10,5 +10,6 @@ export default {
 	'*.{css,scss}': 'stylelint --fix --allow-empty-input --no-validate',
 	'*.{js,ts,tsx,jsx,mjs,cjs}': 'xo --fix',
 	'**/package.json': [() => 'pnpm install', 'pnpm run lint:package-json'],
-	'*': 'cspell --config .config/cspell.config.ts --no-must-find-files --quiet'
+	'*': 'cspell --config .config/cspell.config.ts --no-must-find-files --quiet',
+	'scripts/**/*': [() => 'pnpm --filter scripts test:scripts']
```
nmerget:
```suggestion
	'*': 'cspell --config .config/cspell.config.ts --no-must-find-files --quiet'
```

Shouldn't be part of lintstaged. It's not linting it's testing and it should be part of the pipeline.

mfranzke: @copilot please take care and implement it as part of the pipeline testing.

nmerget: > @copilot please take care and implement it as part of the pipeline testing.

It runs already via the validation inside `scripts/self-healing-orchestrator.ts`

nmerget (approved):

PR #6913, opened 2026-06-05
fix: add propertyName param to all getBooleanAsString calls

### LINKED ISSUE #5494: Setting noText Attribute on db-button removes class

- [ ] All
- [ ] HTML
- [ ] React
- [x] Angular
- [ ] Vue
- [ ] Web components
- [ ] Power Apps

### Reproduction case

Setting the **noText** attribute on a db-button element removes the css class added to the host element.

CSS class is set:
```
<db-button class="myClass"
                 variant="ghost"
                 data-icon="question_mark_circle"
                 (click)="doSomething()">
</db-button>
```

Image

CSS class is missing:
```
<db-button class="myClass"
                 variant="ghost"
                 data-icon="question_mark_circle"
                 noText="noText"
                 (click)="doSomething()">
      </db-button>
```

Image

### Expected Behaviour

The **noText** attribute should not cause the css class to be missing from the db-button element.

### Screenshots

_No response_

### Browser version

None

Sidenote: The css class is still present in the inner <button> element, but without effect because of view encapsulation.

DB InfraGO AG

EBuLa Rückfallebene (IRE)

nmerget:
Ensures every getBooleanAsString call in components passes the property name as the second string parameter for consistent string-based boolean parsing. Also updates the getHideProp utility.

## Proposed changes

resolves #5494

### Code Quality

### Validation

### General

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/fix-get-boolean-as-string-property-name>


.changeset/add-property-name-to-get-boolean-as-string.md
```diff
@@ -0,0 +1,9 @@
+---
+"@db-ux/core-components": patch
+"@db-ux/ngx-core-components": patch
+"@db-ux/react-core-components": patch
+"@db-ux/wc-core-components": patch
+"@db-ux/v-core-components": patch
+---
+
+Add missing `propertyName` parameter to all `getBooleanAsString` calls across components for consistent string-based boolean parsing.
```
mfranzke:
```suggestion
fix: Add missing `propertyName` parameter to all `getBooleanAsString` calls across components for consistent string-based boolean parsing.
```

.changeset/add-property-name-to-get-boolean-as-string.md
```diff
@@ -0,0 +1,9 @@
+---
+"@db-ux/core-components": patch
```
mfranzke:
```suggestion
```

this package isn't needed to get mentioned for this PR, is it ?

packages/components/src/utils/index.ts
```diff
@@ -104,7 +112,11 @@ export const getBoolean = (
 	if (originBool === undefined || originBool === null) return;
 
 	if (typeof originBool === 'string') {
-		return Boolean(propertyName === originBool || originBool === 'true');
+		return Boolean(
+			originBool === '' ||
+			originBool === 'true' ||
+			propertyName === originBool
```
mfranzke:
```suggestion
			propertyName.toLowerCase() === originBool.toLowerCase()
```

shouldn't we check a lowerCase here, as HTML is even also forgiving on the case of the value ?

packages/components/src/utils/index.ts
```diff
@@ -85,13 +85,21 @@ export const delay = (fn: () => void, ms: number) =>
...
+	propertyName?: string
+): any => {
 	if (originBool === undefined || originBool === null) return;
 
 	if (typeof originBool === 'string') {
-		return String(originBool === 'true');
+		return String(
+			originBool === '' ||
+				originBool === 'true' ||
+				propertyName === originBool
```
mfranzke:
```suggestion
				propertyName.toLowerCase() === originBool.toLowerCase()
```

shouldn't we check a lowerCase here, as HTML is even also forgiving on the case of the value ?

packages/components/src/components/accordion-item/accordion-item.lite.tsx
```diff
@@ -86,7 +86,7 @@ export default function DBAccordionItem(props: DBAccordionItemProps) {
 			id={props.id ?? props.propOverrides?.id}
 			class={cls('db-accordion-item', props.className)}>
 			<details
-				aria-disabled={getBooleanAsString(props.disabled)}
+				aria-disabled={getBooleanAsString(props.disabled, 'disabled')}
```
mfranzke, line 89:
```suggestion
				aria-disabled={getBooleanAsString(props.disabled, 'true')}
```

according to the [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-disabled#values), `disabled` isn't a valid value specifically for this `aria-disabled`-HTML-attribute, but only `true` and `false`.

packages/components/src/components/custom-select/custom-select.lite.tsx
```diff
@@ -988,13 +988,17 @@ export default function DBCustomSelect(props: DBCustomSelectProps) {
 					<summary
 						id={state._summaryId}
 						class="db-custom-select-form-field"
-						aria-disabled={getBooleanAsString(props.disabled)}
+						aria-disabled={getBooleanAsString(
+							props.disabled,
+							'disabled'
```
mfranzke, line 993:
```suggestion
							props.disabled,
							'true'
```

according to the [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-disabled#values), `disabled` isn't a valid value specifically for this `aria-disabled`-HTML-attribute, but only `true` and `false`.

packages/components/src/components/custom-select/custom-select.lite.tsx
```diff
@@ -1211,7 +1215,7 @@ export default function DBCustomSelect(props: DBCustomSelectProps) {
 
 			<span
 				class="db-custom-select-placeholder"
-				aria-hidden={getBooleanAsString(true)}
+				aria-hidden={getBooleanAsString(true, 'true')}
```
mfranzke: why don't we set this as try directly, but use `getBooleanAsString` instead ?

packages/components/src/components/link/link.lite.tsx
```diff
@@ -25,13 +25,16 @@ export default function DBLink(props: DBLinkProps) {
 			role={props.role}
 			referrerPolicy={props.referrerpolicy ?? props.referrerPolicy}
 			hrefLang={props.hreflang}
-			aria-disabled={getBooleanAsString(props.disabled)}
+			aria-disabled={getBooleanAsString(props.disabled, 'disabled')}
```
mfranzke, line 28:
```suggestion
			aria-disabled={getBooleanAsString(props.disabled, 'true')}
```

according to the [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-disabled#values), `disabled` isn't a valid value specifically for this `aria-disabled`-HTML-attribute, but only `true` and `false`.

packages/components/src/components/navigation-item/navigation-item.lite.tsx
```diff
@@ -110,10 +110,10 @@ export default function DBNavigationItem(props: DBNavigationItemProps) {
 			class={cls('db-navigation-item', props.className)}
 			data-width={props.width}
 			data-icon={props.icon}
-			data-show-icon={getBooleanAsString(props.showIcon)}
+			data-show-icon={getBooleanAsString(props.showIcon, 'showIcon')}
 			data-active={props.active}
-			data-wrap={getBooleanAsString(props.wrap)}
-			aria-disabled={getBooleanAsString(props.disabled)}>
+			data-wrap={getBooleanAsString(props.wrap, 'wrap')}
+			aria-disabled={getBooleanAsString(props.disabled, 'disabled')}>
```
mfranzke, line 116:
```suggestion
			aria-disabled={getBooleanAsString(props.disabled, 'true')}>
```

according to the [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-disabled#values), `disabled` isn't a valid value specifically for this `aria-disabled`-HTML-attribute, but only `true` and `false`.

packages/components/src/components/custom-select/custom-select.lite.tsx
nmerget: It's our own `disabled` property and doesn't reflect `aria-selected` directly...

packages/components/src/components/accordion-item/accordion-item.lite.tsx
nmerget: It's our own `disabled` property and doesn't reflect `aria-selected` directly...

packages/components/src/components/link/link.lite.tsx
nmerget: It's our own `disabled` property and doesn't reflect `aria-selected` directly...

packages/components/src/components/navigation-item/navigation-item.lite.tsx
nmerget: It's our own `disabled` property and doesn't reflect `aria-selected` directly...

mfranzke:
@codex please review this PR.

packages/components/src/components/custom-select/custom-select.lite.tsx
nmerget: fixed

nmerget (commented):
Solid fix. The `propertyName` parameter ensures that when frameworks pass string attribute values (e.g. `noText="noText"` in Angular templates), they're correctly interpreted as `true` via case-insensitive comparison. The companion test suite is comprehensive and documents the expected behavior well.

A few observations:

🟡 [important] In `button.lite.tsx` and other components, the pattern changed from:
```tsx
data-show-icon={getBooleanAsString(props.showIconLeading ?? props.showIcon)}
```
to:
```tsx
data-show-icon={getBooleanAsString(props.showIconLeading, 'showIconLeading') || getBooleanAsString(props.showIcon, 'showIcon')}
```
This subtly changes behavior: if `showIconLeading` is explicitly `false`, the old code would pass `false` to the function (since `false ?? 'showIcon'` → `false`). The new code would evaluate `"false" || getBooleanAsString(props.showIcon, 'showIcon')`, and since `"false"` is a truthy string, it would short-circuit and return `"false"`. This happens to work correctly but relies on the fact that `getBooleanAsString(false)` returns the string `"false"` which is truthy in JS. Worth a comment to clarify this is intentional.

🟢 [nit] In `custom-select.lite.tsx`:
```tsx
aria-hidden={getBooleanAsString(true)}
```
was changed to `aria-hidden="true"` — good simplification since it's always a static literal.

🟢 [nit] The `toBool` helper change from `value === 'true'` to `value !== 'false'` makes the helper treat any non-"false" string as truthy. This aligns with HTML boolean attribute semantics (presence = true) but is a subtle semantic shift worth documenting.

Overall this is a well-scoped fix with good test coverage. The changeset is correctly structured for a patch bump across all framework packages.

.changeset/add-property-name-to-get-boolean-as-string.md
mfranzke: @nmerget how is this resolved ?

.changeset/add-property-name-to-get-boolean-as-string.md
```diff
@@ -0,0 +1,8 @@
+---
+"@db-ux/ngx-core-components": patch
```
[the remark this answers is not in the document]
mfranzke: that's a false positive here.

mfranzke (approved):

PR #6914, opened 2026-06-05
fix: icon inline size

nmerget:
## Proposed changes

If an icon-font is not loading, the icon name will destroy a layout causing layout-shifts

image

### Code Quality

### Validation

### General

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/fix-icon-inline-size>


packages/foundations/scss/icons/_icon-helpers.scss
```diff
@@ -162,6 +162,7 @@ $default-icon-font-size: var(--db-icon-font-size, #{$default-icon-size-rem});
 	vertical-align: var(--db-icon-vertical-align, middle);
 	/* stylelint-disable-next-line db-ux/use-sizing */
 	block-size: $default-icon-font-size;
```
mfranzke, line 164:
```suggestion
```

@nmerget ~as I don't know how to reproduce those "all cases", could we get rid of (or in other words, replace inline-size by) block-size, as we still have aspect-ratio, so does that aspect-ratio work in that direction ?~ to me it looks like that we could get rid of `block-size` than.

nmerget: The icon has to be square. Which is the normal case, but when the font is not loaded the `:before` just shows the content which is a text. For the text `aspect-ratio` is not applied correctly, `inline-size` does. 

There is only one line extra but it makes sure that the layout won't break

.changeset/fix-icon-inline-size.md
```diff
@@ -0,0 +1,5 @@
+---
+"@db-ux/core-foundations": patch
+---
+
+fix: Re-add `inline-size` to icon styles because `aspect-ratio: 1` alone does not work reliably in all cases.
```
mfranzke:
```suggestion
fix: ensure that the icon fallback is displayed as a rectangle
```

that was what this PR does "internally", but we should tell our users why we do it or in other words what it does fix for them.

nmerget: I didn't replace it. I add `inline-size` to have both

packages/foundations/scss/icons/_icon-helpers.scss
mfranzke: but than we could remove `aspect-ratio`.

.changeset/fix-icon-inline-size.md
mfranzke: that's fine, but still the message needs to tell the users what is changing to them, not to us or our code.

packages/foundations/scss/icons/_icon-helpers.scss
nmerget: I think this was the fix for iOS added from @d-koppenhagen.

He added `min-inline-size` and `max-inline-size` but you said he should use `aspect-ratio`. So we need to keep both

d-koppenhagen: Yes, I remember, both needed since omitting one of them, caused issues with Safari Browser

packages/foundations/scss/icons/_icon-helpers.scss
```diff
@@ -162,6 +162,7 @@ $default-icon-font-size: var(--db-icon-font-size, #{$default-icon-size-rem});
...
+	inline-size: $default-icon-font-size;
```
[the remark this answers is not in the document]
mfranzke: it's fine.

mfranzke (approved):

PR #7029, opened 2026-06-15
fix: "fix tooltip listener cleanup" review

mfranzke:
## Proposed changes

The original cleanup inferred ownership by checking `parent.getAttribute('aria-describedby') === state._id`. The problem: when a trigger declaratively sets `aria-describedby="tooltip-01"` itself (the documented React usage), and the tooltip uses that same id, the equality check passes and the attribute gets removed. Since the parent prop never changed, React won't reapply it after the child unmounts, so the trigger silently loses its accessible name/description.

The new approach captures the parent's attribute value before the tooltip writes to it:

- On attach, `_previousAttributeValue` records what was already there (the trigger's declarative value, or `undefined` if none) and `_attachedAttribute` records which attribute (`aria-labelledby` vs `aria-describedby`).
- On cleanup, if the current value still matches what this tooltip wrote, it restores `_previousAttributeValue` when one existed, otherwise removes the attribute. The id-equality guard is retained so a value set by a different tooltip afterward isn't clobbered.

### Code Quality

### Validation

### General

### DIFF
--- a/packages/components/src/components/tooltip/model.ts
+++ b/packages/components/src/components/tooltip/model.ts
@@ -40,6 +40,8 @@ export type DBTooltipProps = DBTooltipDefaultProps &
 export type DBTooltipDefaultState = {
 	getParent: () => HTMLElement;
 	_attachedParent?: HTMLElement;
+	_attachedAttribute?: string;
+	_previousAttributeValue?: string;
 	_boundListeners?: {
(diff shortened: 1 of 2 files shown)


mfranzke:
@codex review

mfranzke:
@codex review

mfranzke:
we won't proceed with this.

PR #7054, opened 2026-06-17
fix: DBInput and DBTextarea are `readonly` when `readOnly` property has value `false`

### LINKED ISSUE #7051: DBInput breaks when `readOnly=false`

- [ ] All
- [ ] HTML
- [ ] React
- [x] Angular
- [ ] Vue
- [ ] Web components
- [ ] Power Apps

### Reproduction case

Short:

```html
<db-input [readonly]="false" />
```

But this is especially hindersome because Angular's `formField` directive automatically sets the readOnly field under the hood:

```ts
import { Component, signal } from '@angular/core';
import { form, FormField } from '@angular/forms/signals';
import { DBInput } from '@db-ux/ngx-core-components';

@Component({
  selector: 'app-repro',
  imports: [FormField, DBInput],
  template: `
    <db-input [formField]="$any(form.text)" />
  `
})
export class C extends Component {
  readonly model = signal<{ text: string }>({ text: '' })
  readonly form = form(this.model)
}
```

### Expected Behaviour

The element is not readonly and I can type into it.

### Screenshots

_No response_

### Browser version

None

The prop is treated as boolean attribute, so `readonly="false"` is added to the HTML. But readonly is controlled by attribute presence, so the attribute would need to be removed altogether when `false` is passed.

Not sure about the other Mitosis builders, this might also affect other framework packages.

It works if I go into the generated Angular component template and change

```html
      [attr.readOnly]="getBoolean(readOnly(), 'readOnly') || getBoolean(readonly(), 'readonly')"
```

to

```html
      [attr.readOnly]="getBoolean(readOnly(), 'readOnly') || getBoolean(readonly(), 'readonly') || null"
```

DV Fernverkehr

_No response_

michaelmkraus:
Fixes `DBInput` and `DBTextarea` becoming read-only when `readOnly={false}` is passed. This is especially problematic in Angular, where the `formField` directive sets `readOnly` to `false` under the hood, making the field unusable.
The Mitosis binding resolved to a real boolean `false`. In Angular, `[attr.readOnly]="...false"` renders `readonly="false"` into the DOM. Since `readonly` is a boolean HTML attribute, its mere presence makes the field read-only — the value is irrelevant.

## Proposed changes

Appended `|| undefined` to the `readOnly` binding in both Mitosis source files. When neither `readOnly` nor `readonly` is `truthy`, the expression now resolves to `undefined`, so Angular (and React) omit the attribute entirely. The `true` case is unchanged. `undefined` was chosen over the `null` suggested in the issue because it cleanly signals "omit attribute" across all framework targets (React, Angular, Vue, Stencil).

resolves #7051

### Code Quality

### Validation

### General

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/fix-input-readonly-false>


packages/components/src/components/input/input.lite.tsx
mfranzke, line 1: I'm struggling whether we need a more global solution, as `disabled` has the same problem: `disabled="false"` would even also disable the field. So deploying this fix now would be perfectly fine, but the question is whether it would be possible to provide a more generic refactoring in an acceptable amount of time or whether we would need to create a story in the backlog for this.

michaelmkraus: I checked the generated output: `disabled`/`required` on `input`/`textarea` use Angular property bindings (`[disabled]=`), so `false` is handled correctly there — only attribute bindings (`[attr.X]="getBoolean(...)"`) render the literal `false`. Besides `readOnly`, the same pattern affects `navigation-item` `[attr.disabled]`, plus `input` `[attr.multiple]`/`[attr.autoFocus]` and `tab-item` `[attr.checked]`.

A clean generic fix would be to make `getBoolean` return `undefined` instead of `false`, which would cover all `[attr.*]` bindings at once. But since `getBoolean` is used very broadly (including in logic contexts), that change needs dedicated testing. I'd merge this targeted fix for the reported issue and tracking the generic refactor in a separate story.

.changeset/fix-readonly-false-input-textarea.md
```diff
@@ -0,0 +1,11 @@
+---
+"@db-ux/core-components": patch
```
mfranzke:
```suggestion
```

mfranzke (approved):

packages/components/src/components/input/input.lite.tsx
mfranzke: I've created an issue as a follow up for this: https://github.com/db-ux-design-system/core-web/issues/7055

PR #7134, opened 2026-06-26
feat(angular): add signal forms support (#6418)

d-koppenhagen:
Implements Angular Signal Forms strategy for reactive form components. Adds validation bridge, version detection utilities, and post-build script updates for Angular signal model/output generation. Updates Angular showcase form components to demonstrate signal forms usage. Includes ADR-06 documenting the technical decision and comprehensive tests.

## Proposed changes

resolves #6418

### Code Quality

### Validation

### General

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/6418-signal-forms>


.changeset/angular-signal-forms-support.md
```diff
@@ -0,0 +1,9 @@
+---
+"@db-ux/core-components": minor
+"@db-ux/ngx-core-components": minor
+"@db-ux/react-core-components": minor
+"@db-ux/wc-core-components": minor
+"@db-ux/v-core-components": minor
```
mfranzke:
```suggestion
"@db-ux/ngx-core-components": minor
```

To my understanding, we only need to mention the Angular package here.

mfranzke:
@codex review

.changeset/angular-signal-forms-support.md
d-koppenhagen: Okay I want sure about this since I thought always all versions should be aligned but this ist probably part of the release process. Btw. can you release a version for checking it out in a real repo?

mfranzke: > Okay I want sure about this since I thought always all versions should be aligned but this ist probably part of the release process. Btw. can you release a version for checking it out in a real repo?

we could do a pre-release.

mfranzke: > > Okay I want sure about this since I thought always all versions should be aligned but this ist probably part of the release process. Btw. can you release a version for checking it out in a real repo?
> 
> we could do a pre-release.

~https://github.com/db-ux-design-system/core-web/releases/tag/v4.13.1-angular-signal-forms – the exact pre-release version number will be published after https://github.com/db-ux-design-system/core-web/actions/runs/28266541715 was successful.~ https://github.com/db-ux-design-system/core-web/actions/runs/28266541715 failed

packages/components/scripts/post-build/angular.ts
```diff
@@ -177,6 +184,24 @@ export default (tmp?: boolean) => {
 				component.config.angular.controlValueAccessor, // value / checked / ...
 				component.config.angular.controlValueAccessorRequired // Radio needs a value
 			);
+
+			// Signal Forms value alias for components using 'values' (e.g. DBCustomSelect)
+			if (component.config.angular.signalFormsValueAlias) {
+				// Inject value alias and sync effect after duck-typing fields
```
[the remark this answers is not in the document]
d-koppenhagen: Out of scope: This would be a bigger refactoring and should be done in a separate PR when needed.

mfranzke:
@codex review

.config/cspellignorewords.txt
```diff
@@ -228,3 +228,5 @@ fbca
 unreviewed
 notext
 prebundle
+
+unprovide
```
d-koppenhagen:
```suggestion
unprovide
```

jonaskuske:
Hi, noticed that in the latest release validity is no longer configurable but seems to always be controlled by the Signal Form, `validation="no-validation"` is ignored

And the native inputs no longer reflect validation state and thus also lack the browser-provided `validationMessage` - used to show e.g. "Der Wert muss größer oder gleich 2 sein." and now just shows "TODO: add an invalidMessage"

packages/components/scripts/post-build/angular.ts
michaelmkraus, line 1: `AGENTS.md` marks `scripts/post-build/` as deprecated. This PR adds ~276 lines of new transformation logic here. Can we implement the Signal Forms bridge as a Mitosis plugin instead?

packages/components/scripts/post-build/angular.ts
```diff
@@ -197,5 +250,218 @@ export default (tmp?: boolean) => {
...
+				// Verify the validation bridge was injected successfully
+				const verifyContent = readFileSync(file, 'utf-8');
+				if (
+					!verifyContent.includes('// Signal Forms validation bridge')
+				) {
+					console.warn(
+						`⚠️  Signal Forms validation bridge was NOT injected into ${componentName}. ` +
+							`The handleValidation() format may have changed.`
+					);
+				}
```
michaelmkraus: If the generated `handleValidation()` format changes, injection silently no-ops and the build still succeeds with only a `console.warn`. This risks shipping a build with a broken validation bridge without CI catching it. Please make this a hard failure (`throw`) so the build fails loudly when the bridge can't be injected.

scripts/tests/validation-bridge.spec.ts
```diff
@@ -0,0 +1,290 @@
...
+
+/**
+ * Simulates the validation bridge logic injected into handleValidation().
+ * Returns:
+ * - 'no-validation' if validation="no-validation" suppressed everything
+ * - 'signal-forms' if Signal Forms errors triggered an early return
+ * - 'native-fallback' if _valid was reset because native validation disagrees
+ * - 'continue' if normal validation flow continues
+ */
+function simulateValidationBridge(component: {
```
michaelmkraus, line 25: This suite tests a hand-copied re-implementation (`simulateValidationBridge`) rather than the actually injected code, and injection success is only surfaced via `console.warn`, not asserted. The test can stay green while the real generated output drifts/breaks. Could we assert against the generated Angular output (e.g. snapshot the injected `handleValidation()` after build-outputs, or a behavioral test on the compiled component)?

showcases/angular-showcase/src/app/components/form/form.component.ts
```diff
@@ -27,158 +39,135 @@ import {
...
 
-	handleChange(event: Event | void) {
+			// Debounce search input by 300ms
+			debounce(path.search, 300);
+
+			// City is disabled until a country is selected
+			disabled(path.city, ({ valueOf }) => valueOf(path.country) === '');
+
+			// Set to true to test readonly behaviour
+			readonly(path.email, () => false);
```
michaelmkraus: `readonly(path.email, () => false)` is a permanent no-op ("set to true to test"). Either wire it to a real condition to actually demonstrate readonly, or remove it.

showcases/angular-showcase/src/app/components/form/inputs/inputs.component.ts
```diff
@@ -1,18 +1,30 @@
...
+	imports: [
+		WrapperComponent,
+		DBInput,
+		FormsModule,
+		ReactiveFormsModule,
+		FormField
+	],
+	// NO_ERRORS_SCHEMA required for Angular 21 [formField] directive template type-checking.
+	// Remove when Angular 22 is the minimum supported version.
+	schemas: [NO_ERRORS_SCHEMA],
```
michaelmkraus: `NO_ERRORS_SCHEMA` disables _all_ template type-checking for this component, not just the `[formField]` conflict — real template typos (props/elements) will be silenced too. Since it's applied across all form sub-showcases, could we scope it more narrowly?

packages/components/scripts/post-build/angular.ts
```diff
@@ -197,5 +250,218 @@ export default (tmp?: boolean) => {
...
+      this._validMessage.set(DEFAULT_VALID_MESSAGE);
+      this._invalidMessage.set('');
+    }
+
+    // If Signal Forms says "valid" but native validation disagrees, reset _valid
+    // so native validation can take over via CSS :user-invalid selectors
+    if (this._valid() === 'valid' && this._ref()?.nativeElement && !this._ref()?.nativeElement?.validity?.valid) {
+      this._valid.set(undefined);
+      this._validMessage.set('');
+    }
```
michaelmkraus: The native-validity fallback reads `this._ref()?.nativeElement?.validity?.valid`, but for `DBCustomSelect` `_ref` is the outer `<div>` (native validity lives on selectRef, the hidden `<select>`). A `<div>` has no validity, so `this._ref()?.nativeElement?.validity` is `undefined` and `!undefined` is always `true`. As soon as Signal Forms sends `errors=[]` and `_valid` transitions to 'valid', this branch immediately resets `_valid`/`_validity` back to `undefined`, so custom-select can never actually show the Signal Forms "valid" state.

Two options: 
(a) read native validity from the correct element for components that have their own `selectRef`/`_validity` (use `selectRef` instead of `_ref`), or 
(b) only run the fallback when a native validity actually exists, e.g. guard on `this._ref()?.nativeElement?.validity && !this._ref()?.nativeElement?.validity?.valid`.

Same for line [410](https://github.com/db-ux-design-system/core-web/pull/7134/changes#diff-b8dfc3192b4fcae2da0da90c9130a452a0d25cc282f66b6bf26a8ad46e95874fR410)

packages/components/configs/plugins/angular/signal-forms.cjs
```diff
@@ -0,0 +1,315 @@
...
+	code = code.replace(
+		`[attr.aria-invalid]="validation() === 'invalid' ? 'true' : undefined"`,
+		`[attr.aria-invalid]="validation() !== 'no-validation' && (_valid() ?? validation()) === 'invalid' ? 'true' : undefined"`
+	);
+
+	// Connect _valid to data-custom-validity
+	code = code.replace(
+		`[attr.data-custom-validity]="validation()"`,
+		`[attr.data-custom-validity]="validation() === 'no-validation' ? 'no-validation' : (_valid() ?? validation())"`
+	);
```
michaelmkraus, line 186: `code.replace(...)` silently no-ops when the `from` string isn't found. Only the `handleValidation()` marker is verified via throw — the `aria-invalid`, `data-custom-validity` and `[attr.pattern]` replacements are not. 
Suggestion: after each critical replacement, assert the expected substring exists and throw otherwise.

packages/components/configs/plugins/angular/signal-forms.cjs
```diff
@@ -0,0 +1,315 @@
...
+		});
+
+		/** @internal Sync values → value (CVA/user interaction writes to values) */
+		private _syncValuesToValue = effect(() => {
+			const vals = this.values();
+			if (this._syncing) return;
+			this._syncing = true;
+			this.value.set(vals ?? []);
+			this._syncing = false;
+		});`
```
michaelmkraus, line 385: The `_syncing` flag can't prevent circular triggering: Angular `effect()` callbacks run asynchronously via the scheduler, but the flag is set to true and back to false synchronously within the same callback. By the time the other effect runs, `_syncing` is already false, so the guard never blocks. Convergence only works because signal equality stops re-emission. 
Suggestion: 
we could drop the two-effect bridge in favor of a linkedSignal/computed mapping between value and values, or make the guard span a microtask. At minimum, fix the misleading comment.

showcases/angular-showcase/src/app/app.component.ts
```diff
@@ -40,8 +39,12 @@ import {
 		DBButton,
 		SecondaryActionDirective,
 		NavigationDirective,
-		MetaNavigationDirective
+		MetaNavigationDirective,
+		FormField
 	],
+	// NO_ERRORS_SCHEMA required for Angular 21 [formField] directive template type-checking.
+	// Remove when Angular 22 is the minimum supported version.
+	schemas: [NO_ERRORS_SCHEMA],
```
michaelmkraus: schemas: `[NO_ERRORS_SCHEMA]` disables all template type-checking for these components, not just the `[formField]` binding — real template typos would be masked.

packages/components/configs/plugins/angular/signal-forms.cjs
```diff
@@ -0,0 +1,315 @@
...
+	}
+
+	// Inject hidden/errors inputs + @HostBinding before ngAfterViewInit
+	// Find the first ngAfterViewInit() occurrence (the actual method definition)
+	code = code.replace(
+		/(\n\s*ngAfterViewInit\(\))/,
+		`
+		/** Signal Forms optional fields (Duck-Typing compatibility) */
+		hidden = input<boolean>(false);
+		errors = input<any>(undefined);
```
michaelmkraus:
```suggestion
		errors = input<ValidationError[] | undefined>(undefined);
```

mfranzke:
@codex review

packages/components/configs/plugins/angular/signal-forms.cjs
```diff
@@ -0,0 +1,434 @@
...
+      // Signal Forms: re-run validation when errors or validation prop changes externally
+      effect(
+        () => {
+          this.errors();
+          this.validation();
+          this.handleValidation();
+        },
+        Number(VERSION.major) < 19
+          ? ({ allowSignalWrites: true } as any)
+          : undefined
```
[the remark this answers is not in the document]
d-koppenhagen: Not applicable. VERSION is already imported in every generated Angular component by the base Mitosis Angular generator — it's needed for the existing allowSignalWrites version guards that Mitosis emits into all effect()-using components.

packages/components/configs/plugins/angular/signal-forms.cjs
```diff
@@ -0,0 +1,434 @@
...
+      }
+      return;
+    }
+
+    // Signal Forms validation bridge: errors InputSignal has priority
+    const signalFormErrors = this.errors();
+    if (Array.isArray(signalFormErrors) && signalFormErrors.length > 0) {
+      this._descByIds.set(this._invalidMessageId());
+      this._invalidMessage.set(
+        this.invalidMessage() || signalFormErrors[0].message || this._ref()?.nativeElement?.validationMessage || DEFAULT_INVALID_MESSAGE
```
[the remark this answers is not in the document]
d-koppenhagen: Already addressed. The native-validity fallback guard reads this._ref()?.nativeElement?.validity && !this._ref()?.nativeElement?.validity?.valid — for DBCustomSelect, _ref is the outer <div> which has no validity property (undefined), so the entire condition short-circuits to false and _valid is never reset. The "valid" state from Signal Forms is preserved correctly for custom-select.

packages/components/configs/plugins/angular/signal-forms.cjs
```diff
@@ -0,0 +1,434 @@
...
+ * Injects `this.propagateTouched()` and a Signal Forms `touch` output emission
+ * into handleBlur() so both legacy CVA and Signal Forms are notified that the
+ * control was touched. Without this, Signal Forms' `touched()` signal never
+ * becomes true.
+ */
+function injectPropagateTouched(code) {
+	// Match handleBlur method and inject propagateTouched + touch.emit() call at the start
+	code = code.replace(
+		/handleBlur\(event:[^)]+\)\s*\{\n/,
+		`handleBlur(event: any) {\n    this.propagateTouched();\n    this.touch.emit();\n`
```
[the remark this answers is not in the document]
d-koppenhagen: Correct observation — DBCustomSelect has no handleBlur so the injection is a no-op for that component. This is a known limitation documented in the code. Touch state for custom-select is currently propagated through CVA's onTouched callback when the dropdown closes. Full Signal Forms touched() support for custom-select would require changes to the component source itself (emitting on dropdown close), which is out of scope for this PR. Added a NOTE comment documenting this.

docs/adr/adr-06-angular-signal-forms-strategy.md
```diff
@@ -0,0 +1,125 @@
...
+This approach was chosen because:
+
+1. Angular 21's `ngModel` and `formControlName` still require a CVA provider for Custom Components — removing it breaks Template-Driven and Reactive Forms
+2. Angular 22 correctly prioritizes Signal Forms' `[formField]` over CVA when both are present
+3. No runtime version detection needed — simpler, more predictable implementation
+4. CVA and Signal Forms coexist without conflict
+
+## Problem description and context
+
+The library must simultaneously support the **last three Angular major versions** (20, 21, 22):
```
mfranzke:
```suggestion
The library must simultaneously support the **last three Angular major versions** ([Active and two LTS](https://angular.dev/reference/releases#actively-supported-versions): currently 20, 21, 22):
```

nitpick: Maybe this is a little bit clearer.

d-koppenhagen: Listing "currently" with the explicit version numbers isn't a good idea IMO, sind we have regular 6 month releases and I am quite sure we will miss to update this comment et least once. So maybe your suggestion but without the explicit versions

mfranzke: I'm fine with the versions, as I do expect that the ADRs reflect a decision from a specific point in time and that they won't change (any change would be reflected in a new ADR, wouldn't it?). We haven't had that case and I don't know if there's a best practice.

packages/components/configs/plugins/angular/signal-forms.cjs
```diff
@@ -0,0 +1,455 @@
...
+	}
+
+	// Inject hidden/errors inputs + @HostBinding before ngAfterViewInit
+	// Find the first ngAfterViewInit() occurrence (the actual method definition)
+	code = code.replace(
+		/(\n\s*ngAfterViewInit\(\))/,
+		`
+		/** Signal Forms optional fields (Duck-Typing compatibility) */
+		hidden = input(false, { transform: booleanAttribute });
+		errors = input<readonly {message?: string}[] | undefined>(undefined);
```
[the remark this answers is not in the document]
d-koppenhagen: That is a valid point, but it was designed that way intentionally. The `errors` input is a duck-typing field, it accepts errors from the frameworks Signal Forms. The child field is provided at runtime, but the input type does not need to cover the full interface, as it is used solely for display purposes (the message). It has been kept intentionally minimalist.

showcases/angular-showcase/src/app/components/form/form.component.ts
```diff
@@ -27,170 +39,140 @@ import {
 	selector: 'app-form',
 	templateUrl: './form.component.html',
 	imports: [
-		FormsModule,
-		ReactiveFormsModule,
+		JsonPipe,
+		FormField,
```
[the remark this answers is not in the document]
d-koppenhagen: Not applicable: FormField is imported in the component's imports array, so Angular resolves all template bindings correctly. The showcase builds without `NO_ERRORS_SCHEMA`. This workaround is only needed when a component is not declared/imported.

showcases/angular-showcase/src/app/components/form/form.component.html
```diff
@@ -1,379 +1,195 @@
...
+		<db-input
+			type="password"
+			[formField]="userForm.password"
+			[validation]="fieldValidation(userForm.password)"
+			label="Password"
+		></db-input>
+		<db-input
+			type="number"
+			[formField]="userForm.age"
+			[validation]="fieldValidation(userForm.age)"
```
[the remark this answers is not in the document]
d-koppenhagen: Signal Forms manages the value through its own signal binding, not via raw `event.target.value`. The numeric type is preserved by the framework. No conversion needed in the showcase.

showcases/angular-showcase/src/app/components/form/form.component.ts
```diff
@@ -27,170 +39,140 @@ import {
...
+			debounce(path.search, 300);
+
+			// City is disabled until a country is selected
+			disabled(path.city, ({ valueOf }) => valueOf(path.country) === '');
+
+			// Email becomes readonly once username is filled (demonstrates readonly)
+			readonly(
+				path.email,
+				({ valueOf }) => valueOf(path.username) !== ''
+			);
```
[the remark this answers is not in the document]
d-koppenhagen: Good catch: fixed. The readonly condition now requires both username and email to be filled, so the required email field can't get locked while empty.

packages/components/configs/plugins/angular/signal-forms.cjs
```diff
@@ -0,0 +1,455 @@
...
+				return;
+			}
+			this._syncSource = 'value';
+			if (v !== undefined) {
+				this.values.set(Array.isArray(v) ? v : v ? [v] : []);
+			} else {
+				// Reset: undefined clears the selection and resets the sync marker
+				// so subsequent user interactions via 'values' are not blocked.
+				this.values.set([]);
+			}
```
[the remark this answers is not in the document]
d-koppenhagen: Fixed. Added a `_valueAliasActive` guard so the sync effect only fires when the value alias is actively driven by Signal Forms. Initial undefined from unbound state no longer clears existing values selections.

mfranzke (approved):

showcases/angular-showcase/src/app/components/form/form.component.html
```diff
@@ -1,379 +1,195 @@
...
+		></db-input>
+		<db-input
+			type="password"
+			[formField]="userForm.password"
+			[validation]="fieldValidation(userForm.password)"
+			label="Password"
+		></db-input>
+		<db-input
+			type="number"
+			[formField]="userForm.age"
```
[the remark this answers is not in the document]
mfranzke: extracted to https://github.com/db-ux-design-system/core-web/issues/7276

showcases/angular-showcase/src/app/components/form/form.component.ts
```diff
@@ -27,170 +39,143 @@ import {
...
+			// City is disabled until a country is selected
+			disabled(path.city, ({ valueOf }) => valueOf(path.country) === '');
+
+			// Email becomes readonly once both username AND email are filled
+			// (prevents locking an empty required field)
+			readonly(
+				path.email,
+				({ valueOf }) =>
+					valueOf(path.username) !== '' && valueOf(path.email) !== ''
+			);
```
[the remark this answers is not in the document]
mfranzke: https://github.com/db-ux-design-system/core-web/issues/7277

PR #7173, opened 2026-07-01
fix: keys need to be unique

mfranzke:
## Proposed changes

We can't use nullish coalescing here, but need to use logical OR, as `option.id || option.value || uuid()` might return `''`, which might be the first parts `option.id` value if no `id` is being set.

Currently this leads to a warning on the console:
> [browser] Encountered two children with the same key, `select-optgroup-`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version. 

image

And `.value` might (in edge cases) not be unique. That for added `.label` as well.

### Code Quality

### Validation

### General

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/fix-unique-keys>

### DIFF
--- a/.changeset/lovely-lizards-appear.md
+++ b/.changeset/lovely-lizards-appear.md
@@ -2,6 +2,7 @@
 "@db-ux/react-core-components": patch
 "@db-ux/wc-core-components": patch
 "@db-ux/v-core-components": patch
+"@db-ux/ngx-core-components": patch
 ---
 
 fix: `key` special attributes need to be unique
--- a/.changeset/lovely-lizards-appear.md
+++ b/.changeset/lovely-lizards-appear.md
@@ -1,0 +1,5 @@
+---
+"@db-ux/react-core-components": patch
+---
+
+fix: `key` special attributes need to be unique
--- a/.changeset/lovely-lizards-appear.md
+++ b/.changeset/lovely-lizards-appear.md
@@ -1,5 +1,7 @@
 ---
 "@db-ux/react-core-components": patch
+"@db-ux/wc-core-components": patch
+"@db-ux/v-core-components": patch
 ---
 
 fix: `key` special attributes need to be unique
--- a/.changeset/lovely-lizards-appear.md
+++ b/.changeset/lovely-lizards-appear.md
@@ -2,7 +2,6 @@
 "@db-ux/react-core-components": patch
 "@db-ux/wc-core-components": patch
 "@db-ux/v-core-components": patch
-"@db-ux/ngx-core-components": patch
 ---
 
 fix: `key` special attributes need to be unique
--- a/packages/components/src/utils/index.ts
+++ b/packages/components/src/utils/index.ts
@@ -210,7 +210,7 @@ export const getOptionKey = (
 	option: { id?: string; value?: string | number | string[] | undefined },
 	prefix: string
 ) => {
-	const key = option.id || option.value || uuid();
+	const key = option.id || uuid();
 	return `${prefix}${key}`;
 };
 
--- a/packages/components/src/utils/index.ts
+++ b/packages/components/src/utils/index.ts
@@ -208,10 +208,16 @@ export const getSearchInput = (element: HTMLElement): HTMLInputElement | null =>
 
 export const getOptionKey = (
 	option: { id?: string; value?: string | number | string[] | undefined },
-	prefix: string
+	prefix: string,
+	index: number
 ) => {
-	const key = option.id || uuid();
-	return `${prefix}${key}`;
+	// Prefer a stable identifier (id, then value). uuid() is only a last
+	// resort: it returns a new value on every call, which would defeat stable
+	// keys in React and break trackBy in Angular (nodes recreated every cycle).
+	// The index is always appended to guarantee uniqueness even when multiple
+	// options share the same value.
+	const key = option.id || option.value || uuid();
+	return `${prefix}${key}-${index}`;
 };
 
 export const isKeyboardEvent = <T>(
--- a/packages/components/src/utils/index.ts
+++ b/packages/components/src/utils/index.ts
@@ -208,16 +208,17 @@ export const getSearchInput = (element: HTMLElement): HTMLInputElement | null =>
 
 export const getOptionKey = (
 	option: { id?: string; value?: string | number | string[] | undefined },
-	prefix: string,
-	index: number
+	prefix: string
 ) => {
-	// Prefer a stable identifier (id, then value). uuid() is only a last
-	// resort: it returns a new value on every call, which would defeat stable
-	// keys in React and break trackBy in Angular (nodes recreated every cycle).
-	// The index is always appended to guarantee uniqueness even when multiple
-	// options share the same value.
-	const key = option.id || option.value || uuid();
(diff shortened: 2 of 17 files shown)


mfranzke:
@codex review

packages/components/src/utils/index.ts
```diff
@@ -210,7 +210,7 @@ export const getOptionKey = (
 	option: { id?: string; value?: string | number | string[] | undefined },
 	prefix: string
 ) => {
-	const key = option.id ?? option.value ?? uuid();
+	const key = option.id || option.value || uuid();
```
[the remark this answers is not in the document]
mfranzke: this is kind of a false positive, as this generated return value is only used on the special attribute `key`.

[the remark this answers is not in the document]
mfranzke: added a changeset

mfranzke:
```suggestion
	const key = option.id || uuid();
```

.value might (in edge cases) not be unique.

.changeset/lovely-lizards-appear.md
```diff
@@ -0,0 +1,7 @@
+---
+"@db-ux/react-core-components": patch
+"@db-ux/wc-core-components": patch
+"@db-ux/v-core-components": patch
```
michaelmkraus, line 4: Shouldn't we add `@db-ux/ngx-core-components` and reconsider whether `v/wc` should be listed? Right now Angular (which is really affected) is the one omitted.

packages/components/src/utils/index.ts
```diff
@@ -210,7 +210,7 @@ export const getOptionKey = (
 	option: { id?: string; value?: string | number | string[] | undefined },
 	prefix: string
 ) => {
-	const key = option.id ?? option.value ?? uuid();
+	const key = option.id || uuid();
```
michaelmkraus: `uuid()` returns a new value on every call, so options without an `id` get a fresh key on each render — defeating stable keys in React and breaking `trackBy` in Angular (nodes recreated every cycle). Dropping the value fallback widens this.

.changeset/lovely-lizards-appear.md
mfranzke: @michaelmkraus i thought that keys is JSX specific (and by that only React, but even also Vue as well ad Stencil (WC) would be affected. And Angular uses [`trackBy` / `track`](https://angular.dev/api/core/TrackByFunction) instead.

packages/components/src/utils/index.ts
mfranzke: That's right, but the initial cause was actually the method being used in an optgroup …
We need to handle this (most likely by a uuid generated and stored in state).

.changeset/lovely-lizards-appear.md
mfranzke: added another changeset specifically for Angular

packages/components/src/utils/index.ts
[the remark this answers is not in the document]
mfranzke: outdated

packages/components/src/utils/index.ts
```diff
@@ -210,8 +210,15 @@ export const getOptionKey = (
...
 	prefix: string
 ) => {
-	const key = option.id ?? option.value ?? uuid();
-	return `${prefix}${key}`;
+	// If the consumer didn't provide an id, generate one and persist it on
+	// the option object itself. Because the object is passed by reference,
+	// the generated id survives re-renders, reorders, and filtering — making
+	// the key both stable and unique across all frameworks.
+	if (!option.id) {
+		option.id = uuid();
```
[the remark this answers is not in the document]
mfranzke: - **No mutation**: frozen objects, Immer proxies, NgRx selectors, Redux strict-mode state all work fine since we never write to the option object
- **Still stable**: `WeakMap` keys off object identity, so the same object reference always resolves to the same generated UUID across re-renders, reorders, and filtering
- **Still unique**: `uuid()` guarantees no collisions
- **No memory leaks**: `WeakMap` automatically releases entries when the option object is garbage collected (e.g. when the consumer replaces the options array)
- **Signature unchanged**: still just (`option`, `prefix`), so call sites stay clean

packages/components/src/utils/index.ts
mfranzke: implemented this by WeakMap

mfranzke:
@codex review

packages/components/src/utils/index.ts
```diff
@@ -206,11 +206,30 @@ export const stringPropVisible = (
...
+	}
+
+	// Otherwise look up or generate a stable key for this object reference.
+	// Using a WeakMap avoids mutating the consumer's option object (which
+	// could be frozen or part of immutable state) while still persisting
+	// the generated key across re-renders, reorders, and filtering.
+	let key = optionKeyMap.get(option);
+	if (!key) {
+		key = uuid();
+		optionKeyMap.set(option, key);
```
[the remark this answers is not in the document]
mfranzke: added this to the logic

mfranzke:
@codex review

packages/components/src/utils/index.ts
```diff
@@ -206,11 +206,49 @@ export const stringPropVisible = (
...
+	//    a value but users couldn't distinguish identical labels, so in
+	//    practice the combination is unique.
+	const value =
+		option.value !== undefined && option.value !== ''
+			? String(option.value)
+			: '';
+	const label = option.label ?? '';
+
+	if (value || label) {
+		return `${prefix}${value}-${label}`;
```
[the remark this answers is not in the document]
mfranzke: adapted a solution for that.

mfranzke:
@codex review

packages/components/src/utils/index.ts
```diff
@@ -206,11 +206,51 @@ export const stringPropVisible = (
...
 export const getOptionKey = (
-	option: { id?: string; value?: string | number | string[] | undefined },
+	option: {
+		id?: string;
+		label?: string;
+		value?: string | number | string[] | undefined;
+	},
 	prefix: string
 ) => {
-	const key = option.id ?? option.value ?? uuid();
```
nmerget, line 213: Why is this so complex?

This would fix it as well. Instead of checking for null we check for null or an empty string

`const key = option.id || option.value || uuid();`

mfranzke: > Why is this so complex?
> 
> This would fix it as well. Instead of checking for null we check for null or an empty string
> 
> `const key = option.id || option.value || uuid();`

this wouldn't be a fix. On an `optgroup` that doesn't have an `id` defined (the regular case), it would simply fall back to `uuid()`. And using `uuid()` directly doesn't work either as @michaelmkraus already [pointed out](https://github.com/db-ux-design-system/core-web/pull/7173#discussion_r3504703149), so we need a solution like a WeakMap in here – or generate the keys for any (?) other HTML tags than `option` elements differently than `option` elements (which would again be more effort again).

Plus `.value` alone isn't necessarily unique (in edge cases), so we should at least add `.label` (which even also exists for `optgroup` HTML element).

michaelmkraus (approved):

PR #7181, opened 2026-07-02
refactor: remove default `margin-inline` from `[data-icon]` pseudo-elements in favor of `gap`

nmerget:
## Proposed changes

Removes the default `margin-inline` (`var(--db-spacing-fixed-xs)`) from `[data-icon]` / `[data-icon-leading]` / `[data-icon-trailing]` pseudo-elements and replaces them with `gap` on the parent flex containers.

**Why:** The default margin conflicted with the preferred approach of using `gap` on flex/grid containers. Every component using `gap` had to explicitly reset the margin to `0`, which was error-prone and added unnecessary overrides.

### Summary of changes

- **Foundations:** Removed `margin-inline-end` / `margin-inline-start` from `default-icons.scss` and the related SCSS variables (`$default-icon-margin-end`, `$default-icon-margin-start`) and CSS custom properties (`--db-icon-margin-end`, `--db-icon-margin-start`) from `_icon-helpers.scss`
- **Components:** Removed all `--db-icon-margin-end: 0` / `--db-icon-margin-start: 0` overrides from components (brand, checkbox, custom-button, infotext, notification, switch, tab-item, tag) since the margin no longer exists by default
- **Icon passing:** Refactored `_icon-passing.scss` to use a `$gap` parameter instead of relying on icon margin variables
- **Components with explicit spacing needs** (link, navigation-item, navigation, custom-select-list-item): Added explicit `margin-inline-start` / `margin-inline-end` or `gap` where spacing between icon and text is still required
- **`is-icon-text-replace` mixin:** Changed from `font-size: 0 !important` to `text-indent: -9999px` for better accessibility (screen readers can still access text content)
- **Button components:** Added `gap: var(--db-spacing-fixed-xs)` to buttons with text
- **Migration guide:** Created `docs/migration/v4.x.x-to-v5.0.0.md` with detailed migration instructions and added it to the README migration list

### Code Quality

### Validation

### General

🔭🐙🐈 Test this branch here:

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/refactor-data-icon-margin>


mfranzke:
@codex review

docs/migration/v4.x.x-to-v5.0.0.md
```diff
@@ -0,0 +1,70 @@
...
+    ```
+
+### Removed CSS custom properties
+
+The following CSS custom properties are no longer functional:
+
+- `--db-icon-margin-end` — previously controlled `margin-inline-end` on `::before`
+- `--db-icon-margin-start` — previously controlled `margin-inline-start` on `::after`
+
+If you were using these custom properties to adjust icon spacing, replace them with direct `margin-inline-end` / `margin-inline-start` on the respective pseudo-element, or use `gap` on the parent.
```
mfranzke, line 47: > replace them with direct `margin-inline-end` / `margin-inline-start` on the respective pseudo-element,

If this is one of our suggestions, why don't we keep those CSS Custom Properties, as this would be a Styling API to expose these specific CSS Custom Properties?

docs/migration/v4.x.x-to-v5.0.0.md
```diff
@@ -0,0 +1,70 @@
...
+
+### Removed SCSS variables
+
+- `icons.$default-icon-margin-end`
+- `icons.$default-icon-margin-start`
+
+If you referenced these in `calc()` expressions (e.g., for padding calculations), remove them and adjust your calculations accordingly.
+
+## Changed `is-icon-text-replace` mixin to use `text-indent` instead of `font-size: 0`
```
mfranzke:
```suggestion
## Changed `is-icon-text-replace` SASS mixin to use `text-indent` instead of `font-size: 0`
```

__snapshots__/badge/showcase/chromium/DBBadge-should-match-screenshot-1/DBBadge-should-match-screenshot.png
mfranzke, line 1: the read dots in the badges are cut off:
https://design-system.deutschebahn.com/core-web/review/refactor-data-icon-margin/components/feedback/badge/overview

nmerget: Good catch. The AI is sometimes bad. It adds `overflow:hidden` which is not required but leads to different problems..

fixed it

docs/migration/v4.x.x-to-v5.0.0.md
nmerget: Good point. We deliberately removed these CSS Custom Properties (`--db-icon-margin-end` / `--db-icon-margin-start`) because they created a conflicting API: every component using `gap` had to override them to `0`, which was error-prone. The replacement approach is:

- **Internal components** use `gap` on the flex container (preferred) or direct `margin-inline-start`/`margin-inline-end` on the pseudo-element where `gap` doesn't apply (e.g. wrapped links).
- **External consumers** who need custom icon spacing can set `margin-inline-end` / `margin-inline-start` directly on `::before` / `::after`, or use `gap` on their parent container.

Re-exposing them as a "Styling API" would re-introduce the same conflict. If we want a public API for icon spacing, it should probably be a new, opt-in custom property that doesn't have a default value baked into the global icon styles. But that feels like a separate enhancement rather than something to block this refactor on. What do you think?

nmerget:
Thanks for the review @chatgpt-codex-connector! Here's the status of each point:

**1. `is-icon-text-replace` width/clip concerns** — Already addressed in the latest commits: `overflow: hidden` was removed (fixes badge clipping), and `> * { text-indent: 0 }` ensures children render correctly. Width is not a concern because icon-only buttons have explicit `inline-size` set.

**2. Table-row z-index stacking** — Already fixed separately in #7323 (merged into main and this branch). No diff to `table-row.scss` exists in this PR.

**3. Wrapped link icon spacing** — Fixed in `5d0a290`: replaced `gap` with direct `margin-inline-start` on `::after`, which works for both `inline-flex` and `inline-block` (wrapped) cases.

**4. Small button gap** — Fixed in `5d0a290`: added `gap: variables.$db-spacing-fixed-2xs` to the existing `data-size="small"` + `:not([data-no-text="true"])` selector to preserve the smaller spacing.

**5. `overflow: hidden` clipping badges/tooltips** — Already resolved by removing `overflow: hidden` from the mixin in `73e4e66`.

mfranzke:
> **1. `is-icon-text-replace` width/clip concerns** — Already addressed in the latest commits: `overflow: hidden` was removed (fixes badge clipping), and `> * { text-indent: 0 }` ensures children render correctly. Width is not a concern because icon-only buttons have explicit `inline-size` set.

This seems to work fine for flexbox elements, but regular elements now do not have these dimensions and the text would get displayed (easiest to reproduce by deactivating the `display: flex` CSS declarations).

Probably we would need to refactor the `is-icon-text-replace` in general e.g. by not depending on any CSS hiding the text, but not rendering this text in the first place and using `aria-label`-HTML-attribute instead (for interactive elements, not generic ones).

Is it necessary to refactor `is-icon-text-replace` in the context of this PR that should remove the default `margin-inline`?

mfranzke:
> **5. `overflow: hidden` clipping badges/tooltips** — Already resolved by removing `overflow: hidden` from the mixin in `73e4e66`.

But now the phrase `menu` is clearly shown on the [screenshot](https://github.com/db-ux-design-system/core-web/blob/9171ca2c8ec6dc63c53184a66d14f15a75c455ce/__snapshots__/header/showcase/firefox/DBHeader-should-match-screenshot-1/DBHeader-should-match-screenshot.png) in Firefox on the [header / behavior example](https://design-system.deutschebahn.com/core-web/review/refactor-data-icon-margin/components/layout/header/overview).

docs/migration/v4.x.x-to-v5.0.0.md
mfranzke: > Re-exposing them as a "Styling API" would re-introduce the same conflict. If we want a public API for icon spacing, it should probably be a new, opt-in custom property that doesn't have a default value baked into the global icon styles. But that feels like a separate enhancement rather than something to block this refactor on. What do you think?

Instead of re-exposing it after deleting it now, I would suggest to keep it. Having our users selecting into our HTML structure is exactly why we provide some CSS Custom Properties as Styling API, so abstract our structure. And I could completely agree with not providing a default value, can't identify why this would be necessary.

packages/components/src/styles/internal/_icon-passing.scss
```diff
@@ -50,8 +49,8 @@ $interactive-button-selectors: "label,button,a";
 		> :is(#{$interactive-button-selectors}),
 		> * > :is(#{$interactive-button-selectors}) {
 			padding-inline-end: calc(
-				#{$inline-end} + #{icons.$default-icon-margin-end} +
-					var(--db-icon-font-size)
+				#{$inline-end} + #{variables.$db-spacing-fixed-xs} +
+					var(--db-icon-font-size, 1lh)
```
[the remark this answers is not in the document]
nmerget: `noText` Buttons should just use `icon` not `iconTrailing`

packages/foundations/scss/icons/_icon-helpers.scss
```diff
@@ -91,16 +78,20 @@ $default-icon-font-size: var(--db-icon-font-size, #{$default-icon-size-rem});
...
+	text-indent: -200vw;
+
+	> * {
+		text-indent: 0;
+	}
 
 	@if $position == "before" {
 		&::before {
-			--db-icon-margin-end: 0;
+			text-indent: 0;
```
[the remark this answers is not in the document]
nmerget: fixed

packages/foundations/scss/icons/_icon-helpers.scss
```diff
@@ -91,16 +78,20 @@ $default-icon-font-size: var(--db-icon-font-size, #{$default-icon-size-rem});
 
 // SCSS mixin for elements that text should get hidden in favour of only displaying the included icon
 @mixin is-icon-text-replace($position: "before") {
-	font-size: 0 !important;
+	text-indent: -200vw;
+
+	> * {
+		text-indent: 0;
```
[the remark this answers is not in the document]
nmerget: fixed

.changeset/remove-icon-margin-inline.md
```diff
@@ -0,0 +1,10 @@
+---
+"@db-ux/core-foundations": major
+"@db-ux/core-components": major
+"@db-ux/ngx-core-components": major
+"@db-ux/react-core-components": major
+"@db-ux/wc-core-components": major
+"@db-ux/v-core-components": major
+---
+
+refactor: remove default `margin-inline` from `[data-icon]` pseudo-elements in favor of `gap`; change `is-icon-text-replace` from `font-size: 0` to `text-indent: -100lvh` for better accessibility
```
mfranzke:
```suggestion
refactor: remove default `margin-inline` from `[data-icon]` pseudo-elements in favor of `gap`
```

docs/migration/v4.x.x-to-v5.0.0.md
mfranzke: We have decided to remove it, cut through the old knot and only reintegrate it if necessary.

docs/migration/v4.x.x-to-v5.0.0.md
```diff
@@ -0,0 +1,70 @@
...
+
+| Before                                  | After                                     |
+| --------------------------------------- | ----------------------------------------- |
+| `font-size: 0 !important`               | `text-indent: -100lvh; overflow: hidden;` |
+| `::before { --db-icon-margin-end: 0 }`  | `::before { text-indent: 0 }`             |
+| `::after { --db-icon-margin-start: 0 }` | `::after { text-indent: 0 }`              |
+
+### How to migrate
+
+If you used the `is-icon-text-replace` mixin or the `.is-icon-text-replace` class, this change is automatic. However, if you were relying on `font-size: 0` behavior in combination with other styles (e.g., using the zero font-size to collapse layout), review your layout after upgrading.
```
mfranzke:
```suggestion
```

Those are obsolete now

mfranzke (dismissed):

mfranzke (approved):

packages/components/src/styles/internal/_button-components.scss
```diff
@@ -76,6 +76,12 @@
 		}
 	}
 
+	&:not([data-no-text="true"]) {
+		#{$selector} {
+			gap: variables.$db-spacing-fixed-xs;
```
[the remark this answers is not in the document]
nmerget: This should be like this yes

PR #7216, opened 2026-07-06
feat(changesets): enabling a publish dry-run

mfranzke:
## Current situation

The [current pipeline](https://github.com/db-ux-design-system/core-web/pull/7294/checks?check_run_id=86316432082) is quite of annoying – it's skipping most of the steps and tries a dry run at the end, which doesn't work, as there's no build step.

image

## Proposed changes

Previous to triggering a release we'd like to ensure that there aren't any known issues that we could've found easily by running the pipeline on `main` once. To achieve that, the publish dry-run now exercises the **full publish path** (`scripts/github/publish-npm.js` with `DRY_RUN_ONLY=true`) — including version stamping, dependency rewriting, individual `pnpm pack` per package, registry config, and `pnpm publish --dry-run` — rather than just a shallow `pnpm pack --recursive`.

The dry-run runs early (as soon as `build-outputs` completes) and feeds into `checks-done` as a required gate. On real release events, the actual publish still waits for `checks-done` to succeed before running with provenance.

Additionally, the `checks-done` changeset-release gate now requires:
- `lint` success (previously unchecked on this path)
- `test-screen-reader` success (previously accepted `"skipped"`, hiding upstream failures)
- `publish-dry-run` success

In any case, the user responsible to do a release could still skip the pipeline by overruling with "Merge without waiting for requirements to be met (bypass rules)", if they're 100% confident that the following release pipeline wouldn't fail (and still in that case, a [restart isn't a big deal](https://github.com/db-ux-design-system/core-web/blob/204c578d02efd5e6ec796cffbf08dc3d23ba3580/.changeset/README.md#5-re-start-a-release-optional)).

**Jobs that will run on changeset-release PRs:**

| Job | Purpose |
|-----|---------|
| `init` | Detect changeset release, set outputs |
| `lint` | Validate code quality |
| `build-packages` | Build core packages → artifacts |
| `build-outputs` | Build framework outputs → `db-ux-outputs` artifact |
| `init-playwright` | Set up Playwright (needed for screen-reader) |
| `test-vite-plugin` | Test the vite plugin |
| `test-components` | Component e2e tests |
| `test-foundations` | Foundation e2e tests |
| `build-showcase-react` | Build react showcase (needed for test-showcase-react) |
| `test-showcase-react` | React showcase e2e tests |
| `test-screen-reader` | Screen reader accessibility tests |
| `publish-dry-run` | Full publish dry-run (version stamp, pack, `pnpm publish --dry-run`) |
| `checks-done` | Validates all core results including lint, screen-reader, and publish dry-run |

**Jobs that will be skipped:**

- All storybook builds (`build-storybook-*`)
- All other showcase builds (stencil, angular, angular-ssr, next, vue, nuxt, patternhub)
- All other showcase tests (stencil, angular, vue, patternhub)
- `build-stories`, `build-figma-code-connect`, `test-figma-code-connect`
- `deploy`, `publish-figma-code-connect`, `preview-url-pr-description`
- `publishpackages` (real publish — only runs on `release` events)
- All `regenerate-snapshots-*` (their test prerequisites won't fail from the bot)

### Code Quality

### Validation

### General

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/feat-changesets-enabling-a-dry-run>


.github/workflows/default.yml
michaelmkraus, line 1: The "dry-run" only executes pnpm pack `--recursive --dry-run`, which validates packaging but never exercises the real publish path, npm auth/provenance, registry config, or version computation. Since the stated goal is to catch issues before triggering a release, packing alone gives limited confidence. Could we either document this as an intentional scope limitation, or extend the dry-run to cover more of the publish logic (e.g. npm publish --dry-run against the real script)?

.github/workflows/default.yml
```diff
@@ -424,8 +438,33 @@ jobs:
...
+            echo "resultTestVitePlugin: $resultTestVitePlugin"
+            echo "resultTestShowcaseReact: $resultTestShowcaseReact"
+            echo "resultTestScreenReader: $resultTestScreenReader"
+
+            if [[ $resultBuildOutputs == "success" ]] && \
+               [[ $resultTestPackages == "success" ]] && \
+               [[ $resultTestFoundations == "success" ]] && \
+               [[ $resultTestVitePlugin == "success" ]] && \
+               [[ $resultTestShowcaseReact == "success" ]] && \
+               [[ $resultTestScreenReader == "success" || $resultTestScreenReader == "skipped" ]]; then
```
michaelmkraus: The changeset-release path intentionally enables test-screen-reader, but here a skipped result is still accepted as passing. If an upstream dependency fails, which has no `cancel-on-failure-*` guard — test-screen-reader is skipped and checks-done still goes green, so the screen-reader check never actually ran. 
Suggestion: require success on the changeset path (drop the `|| == "skipped"`), or add a `cancel-on-failure-init-playwright` job so a failed setup cancels the run instead of silently passing.

.github/workflows/default.yml
mfranzke: @michaelmkraus ~~this was an intentional decision to do a build and do a dry run of the publish. What do you mean by `npm publish --dry-run`? A dry run of `pnpm publish` should be performed by https://github.com/db-ux-design-system/core-web/blob/main/scripts/github/publish-npm.js#L134~~
this one is being prevented by `if: ${{ github.event_name == 'release' }}` condition, so we need to extend that.

.github/workflows/default.yml
```diff
@@ -111,7 +108,7 @@ jobs:
 
   test-vite-plugin:
     uses: ./.github/workflows/02-vite-plugin-test.yml
-    needs: [build-outputs]
+    needs: [build-outputs, init]
```
michaelmkraus: `test-vite-plugin` gains an `init` dependency, but this job has no if condition referencing `needs.init.outputs.*`. The `init` need is redundant here since `build-outputs` already transitively depends on `init`. Suggest reverting to `needs: [build-outputs]` for clarity.

mfranzke:
@codex review

.github/workflows/default.yml
```diff
@@ -424,8 +438,33 @@ jobs:
...
+            echo "resultTestShowcaseReact: $resultTestShowcaseReact"
+            echo "resultTestScreenReader: $resultTestScreenReader"
+
+            if [[ $resultBuildOutputs == "success" ]] && \
+               [[ $resultTestPackages == "success" ]] && \
+               [[ $resultTestFoundations == "success" ]] && \
+               [[ $resultTestVitePlugin == "success" ]] && \
+               [[ $resultTestShowcaseReact == "success" ]] && \
+               [[ $resultTestScreenReader == "success" || $resultTestScreenReader == "skipped" ]]; then
+              echo "🎉 All changeset release checks passed."
```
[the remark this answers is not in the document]
mfranzke: This feedback is stale — it references the old `pnpm pack --recursive --filter "@db-ux/*"` step at line 45, which no longer exists. The earlier fix already replaced that shallow dry-run with a single step that always runs `publish-npm.js` (with `DRY_RUN_ONLY=true` for non-release events). That script operates on the `build-outputs` directory — the same artifacts the real release uses — so the concern is already addressed.

No changes needed. The current implementation already exercises the real publish path (version stamping, dependency rewriting, packing from `build-outputs`, and `pnpm publish --dry-run`) on every non-release run.

mfranzke:
@codex review

mfranzke:
@codex review

.github/workflows/default.yml
```diff
@@ -572,8 +622,18 @@ jobs:
     needs: [deploy, init]
     uses: ./.github/workflows/03-preview-url-pr-description.yml
 
+  publish-dry-run:
```
michaelmkraus, line 625: This dry-run executes PR-controlled code with release-grade permissions. The reusable workflow grants `id-token: write` and `contents: write`, inherits secrets, checks out the PR revision, and then executes `scripts/github/publish-npm.js`. The current CI log confirms that every `pnpm publish --dry-run` invocation successfully requests an npm OIDC token.

A same-repository PR could therefore modify this script to omit `--dry-run` and publish through the trusted-publisher identity before the change is merged. Please separate the PR dry-run into a least-privilege job (`contents: read`, no `id-token`, and no inherited release secrets) and reserve the OIDC-enabled publish job for protected release events or an appropriately protected GitHub environment.

.github/workflows/default.yml
```diff
@@ -572,8 +622,18 @@ jobs:
...
+    if: ${{ !cancelled() && github.event_name != 'release' }}
+    uses: ./.github/workflows/03-publish-packages.yml
+    needs: [init, build-outputs]
+    secrets: inherit
+    with:
+      release: ${{ needs.init.outputs.release }}
+      preRelease: ${{ needs.init.outputs.preRelease }}
+      version: ${{ needs.init.outputs.version }}
```
michaelmkraus: `needs.init.outputs.version` is only populated when `00-init.yml` runs its `publish` job, which happens exclusively for release events. This job runs exclusively for non-release events, so the value is always empty here.

As a result, `publish-npm.js` silently substitutes `0.0.0-local`, rewrites every package and internal dependency to that fabricated version, and uses the `next` tag. The current CI log confirms this behavior even though the generated package manifests contain version `5.1.3`. Consequently, the changeset gate does not rehearse the prospective release version or channel and can pass while version-dependent behavior fails during the real release.

Please derive and pass the intended version/channel from the changeset release manifests or another explicit preflight input, and avoid accepting the local fallback for this CI gate.

.github/workflows/default.yml
mfranzke: @michaelmkraus I don't understand how a separate workflow should bring more security to this whole topic. If a same repository PR would be possible to get opened, any code could get changed. I only could've identified the these four different solutions for preventing this, but all of those are beyond the changes of this PR (in other words, this attack vendor would even be possible today without this change):

1. **Using `pull_request_target`**: To ensure that a workflow is only get executed by it's default's branch (`main`) state and not the one out of a PR, we could use `pull_request_target` instead of `pull_request` event. Most secure, but complicated regarding any changes for our workflow code, or just running that `publish --dry-run` in a separate workflow, but that would still depend on a build, and we couldn't use the results out of it directly in the `default` job (even though that it would fail the pipeline, which might be even already enough).
2. **Branch rulesets**: GitHub's newer rulesets can require specific workflows to pass and restrict who can bypass.
3. **npm environment protection** — bind the OIDC token to a GitHub environment (e.g. `npm-publish`) that is restricted to the `main` branch only. The dry-run job on a PR branch wouldn't satisfy the environment's branch constraint, so the OIDC token request would fail even if the workflow has `id-token: write`.
4. Using **staged npmjs.com deployments**.

mfranzke:
this topic seems to be a little bit too complicated and we're running most tests in the pipelines previously already, plus a re-run is easy to do. So we'll skip further work on this topic for the moment.

PR #7257, opened 2026-07-08
fix(screen-reader test): switch test interactions

mfranzke:
## Proposed changes

Updated the test URL to point to the `disabled` switch page (as mentioned in the test description), removed non-existing (on these example pages) third switch and corrected the comments.

### Currently: `checked` state

image

### Changed to: `disabled` (as mentioned in the test description)

image

The now included word "unavailable" results out of the interaction with a `disabled` switch:
`input[type="checkbox"][role="switch"]:disabled`

Snapshots have been tested with pipeline https://github.com/db-ux-design-system/core-web/actions/runs/29038593909

### Code Quality

### Validation

### General

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/fix-screen-reader-test-switch-test-interactions>


showcases/screen-reader/tests/switch.spec.ts
```diff
@@ -8,24 +8,20 @@ test.describe('DBSwitch', () => {
 		title: 'default',
 		description:
 			'should toggle switches, should not toggle disabled switch',
-		url: './#/03/switch?page=checked',
+		url: './#/03/switch?page=disabled',
```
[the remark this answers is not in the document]
mfranzke: https://github.com/db-ux-design-system/core-web/actions/runs/28948530595 is even already running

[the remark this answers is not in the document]
mfranzke: updated the snapshot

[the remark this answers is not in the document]
mfranzke: we'll update all VoiceOver snapshots with https://github.com/db-ux-design-system/core-web/pull/6930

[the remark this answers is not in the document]
mfranzke: fixed that replacement

mfranzke:
@codex review

showcases/screen-reader/tests/switch.spec.ts
```diff
@@ -8,37 +8,35 @@ test.describe('DBSwitch', () => {
...
```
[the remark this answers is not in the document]
mfranzke: we'll update all VoiceOver snapshots with https://github.com/db-ux-design-system/core-web/pull/6930

showcases/screen-reader/tests/switch.spec.ts
```diff
@@ -8,37 +8,34 @@ test.describe('DBSwitch', () => {
...
-				 * We avoid this by replacing the generated log files
+				 * NVDA sometimes announces "blank" instead of the disabled switch,
+				 * so we normalize it to the expected announcement.
 				 */
 				await generateSnapshot(nvda, retry, (phraseLog) =>
 					phraseLog.map((log) =>
 						log
 							// NVDA sometimes shows "blank"
-							.replace('blank', 'switch, off, True')
+							.replace('blank', 'switch, unavailable, off, True')
```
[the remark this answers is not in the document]
mfranzke: corrected this in the snapshot. But I assume that both `unavailable` and `off` are still at the wrong place in the snapshot, so we wait for the pipeline results.

nmerget:
Why do we do this now?

We should consider having one example for the guidpup tests, which cover most states/problems of a component. 
With this implementation we don't check anymore if the `on` switch works properly for screen-readers

showcases/screen-reader/tests/switch.spec.ts
```diff
@@ -8,37 +8,34 @@ test.describe('DBSwitch', () => {
 		title: 'default',
 		description:
 			'should toggle switches, should not toggle disabled switch',
-		url: './#/03/switch?page=checked',
+		url: './#/03/switch?page=disabled',
 		async testFn(voiceOver, nvda, page) {
```
michaelmkraus, line 12: the page parameter is unused and can be removed to keep the signature clean.
```suggestion
		async testFn(voiceOver, nvda) {
```

showcases/screen-reader/tests/switch.spec.ts
```diff
@@ -8,37 +8,34 @@ test.describe('DBSwitch', () => {
 		title: 'default',
 		description:
 			'should toggle switches, should not toggle disabled switch',
-		url: './#/03/switch?page=checked',
+		url: './#/03/switch?page=disabled',
```
michaelmkraus, line 11: By moving from `page=checked` to `page=disabled`, no switch starts in the `on` state, so we lose coverage for how an `enabled`/`checked` switch is announced to screen readers. Consider a single example page that covers the relevant states (`checked` and `disabled`).

showcases/screen-reader/__snapshots__/windows/chromium/DBSwitch-default-1.txt
```diff
@@ -1 +1 @@
-["switch, off, (Default) False","","switch, on, True","off","switch, off, True"]
\ No newline at end of file
+["switch, off, (Default) False","","switch, unavailable, off, True","off","switch, off, True"]
```
michaelmkraus: The normalization target (`switch, unavailable, off, True`) doesn't match the committed snapshot, which stores `switch`, `unavailable`, `on`, `True` for the disabled switch. The `on` came from the merge with main; on the `disabled` page switch 2 is unchecked, so `off` is correct. Note the screen-reader job was skipped in CI, so this snapshot wasn't actually validated by an NVDA run. Please re-run the screen-reader test on the `disabled` page to regenerate the snapshot.

showcases/screen-reader/tests/switch.spec.ts
mfranzke: extracted this into https://github.com/db-ux-design-system/core-web/pull/7278, as one other file is even also involved.

showcases/screen-reader/tests/switch.spec.ts
mfranzke: The `on` state itself is mainly the browsers default announcement. We still have it included as a result out of an interaction, so I think it's covered. Adding another example might be overhead.

showcases/screen-reader/__snapshots__/windows/chromium/DBSwitch-default-1.txt
mfranzke: updated the snapshot.

showcases/screen-reader/tests/switch.spec.ts
nmerget: > The `on` state itself is mainly the browsers default announcement. We still have it included as a result out of an interaction, so I think it's covered. Adding another example might be overhead.

But this is the case for `disabled` as well. I would recommend to test both somehow. In this case we don't need an additional example. We could just use `act()` on the first switch which shows the interaction even better. 

If we start to refactor the guidepup tests, we should consider to align the tests somehow. It's expensive to start the tests so we should have a clear way of testing. This on/off, disabled, etc. can be applied to all form-components. So please make a plan and add this to the sprint instead of doing such a one-shot without clear need.

mfranzke: > We could just use `act()` on the first switch which shows the interaction even better.

@nmerget That's exactly how it's currently included:
https://github.com/db-ux-design-system/core-web/blob/dd20b72b92950d56d798af14aee6bec83f4e9b0e/showcases/screen-reader/tests/switch.spec.ts#L15

mfranzke: > So please make a plan and add this to the sprint instead of doing such a one-shot without clear need.

@nmerget The need is to fix the tests regarding their current setup. Please proceed by creating a story like you suggested.

mfranzke:
> Why do we do this now?
> 
> We should consider having one example for the guidpup tests, which cover most states/problems of a component. With this implementation we don't check anymore if the `on` switch works properly for screen-readers

@nmerget we do this, because the test description doesn't match the tests URL, and that for the test doesn't meet the expectations.
With the current test setup of having distinct examples, you could either test the initially default `on` or `disabled` variants. To fix the current situation would either mean changing the description or the URL. I've chosen the latter, as we still have the interaction with the first element included, so we actually **do** have the `on` state included in this test, but haven't had the result out of an interaction with a `disabled` element, which is mentioned in the test description.

So this both fixes the mismatch, as well as covers more states/problems of a component.

michaelmkraus (approved):

PR #7438, opened 2026-07-27
fix: support Angular RouterLink on DBLink

### LINKED ISSUE #7369: DBLink is inaccessible when used with routerLink

- [ ] All
- [ ] HTML
- [ ] React
- [x] Angular
- [ ] Vue
- [ ] Web components
- [ ] Power Apps

### Reproduction case

```html
<db-link routerLink="help">Help</db-link>
```

renders

```html
<db-link routerlink="help" tabindex="0">
  <a class="db-link" tabindex="0" data-show-icon-trailing="true" data-content="internal">Test</a>
</db-link>
```

Here, the `<a>` is only focusable because it has an explicit `tabindex`, but still isn't clickable (and has no accessible `role`) due to the missing `href`, so pressing Enter/Space has no effect.

If I do `<db-link routerLink="help" href="#">Help</db-link>` then the `<a>` is valid and can be activated through Enter/Space, but it will skip the routerLink and regularly navigate to the `href`, because the routerLink listener sits on the parent `<db-link>`.

### Expected Behaviour

`<db-link>` can be used for in-app navigations

### Screenshots

_No response_

### Browser version

None

The `routerLink` directive is [built in a way where it handles `a`, `area` and Custom Elements](https://github.com/angular/angular/blob/f8c9e9d736b95fc5ec64841419aeb534361bdede/packages/router/src/directives/router_link.ts#L385-L400) with a `href` attribute, but not Angular components with a `href` input, so it never sets the `href`. I originally thought this meant taking a `<a>` as child was the only way to make this work...

```html
<db-link>
  <a link routerLink="help">Help</a>
</db-link>
```

...but I've since found out that you can make it work by injecting RouterLink (optional) and if it's present, read the `href` via the `urlTree` property (which has a `toString()` that returns a relative link) and prevent the default navigation on click (the event then bubbles to the parent `<db-link>` element where the `routerLink` listener handles it):

```ts
@Component({
  selector: 'db-link',
  template: `
    <a [href]="routerLink?.urlTree ?? href()" (click)="void (routerLink && $event.preventDefault())"><ng-content/></a>
  `
})
export class DbLink extends Component {
  readonly routerLink = inject(RouterLink, { optional: true })
  readonly href = input<string>()
}
```

DB Fernverkehr

_No response_

jonaskuske:
## Proposed changes

resolves 

### Code Quality

### Validation

> ℹ️ manually tested in Edge, Firefox and WebKit

### General

___

AI disclosure: built with GPT 5.6 Sol


jonaskuske:
The previous handling did not respect `Shift+Click`, `Ctrl+Click` etc., added now by checking the various conditions that should not result in a router navigation and calling `stopPropagation` so the event doesn't reach the RouterLink handler on the host.

A shorter alternative would be:

```ts
if (this.routerLink) {
  // @ts-ignore
  this.routerLink.isAnchorElement = true
}
```

in the constructor, this enables the built-in handling of `preventDefault()` and `Shift/Ctrl/...+Click` in the RouterLink's click handler.

However `isAnchorElement` is internal and read-only (hence the `@ts-ignore`), so I've opted for the manual condition check, but happy to change if you prefer this shorter solution.

PR #7449, opened 2026-07-28
fix: support Angular RouterLink on DBLink

### LINKED ISSUE #7369: DBLink is inaccessible when used with routerLink

mfranzke:
## Proposed changes

resolves https://github.com/db-ux-design-system/core-web/issues/7369 (follow up to the fork PR https://github.com/db-ux-design-system/core-web/pull/7438; kudos to @jonaskuske for that)

### Code Quality

### Validation

> ℹ️ manually tested in Edge, Firefox and WebKit

### General

___

AI disclosure: built with GPT 5.6 Sol

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/fix-routerlink-on-db-link>


mfranzke:
@codex review

packages/components/configs/plugins/angular/router-link.cjs
```diff
@@ -0,0 +1,49 @@
...
+ * This mirrors the onClick handler of RouterLink:
+ * https://github.com/angular/angular/blob/5ad823139758b4d3a8a021d378b008c3457f8689/packages/router/src/directives/router_link.ts#L479
+ */
+const transformRouterLink = (code, componentName) => {
+	if (componentName !== 'DBLink') return code;
+
+	const replacements = [
+		[
+			'} from "@angular/core";',
+			'inject, } from "@angular/core";\nimport { RouterLink } from "@angular/router";'
```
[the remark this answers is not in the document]
jonaskuske: `@db-ux/ngx-core-components` currently list no peer deps at all even though it already depends on `@angular/common` and others, so this is unrelated to the changes introduced in this PR. Closing

jonaskuske:
@codex review

mfranzke:
@codex review

packages/components/configs/plugins/angular/router-link.cjs
```diff
@@ -0,0 +1,49 @@
...
+			'[attr.href]="computeRouterHref(routerLink?.urlTree) ?? href()"\n    (click)="handleClick($event)"'
+		],
+		[
+			'  constructor() {}',
+			'  readonly routerLink = inject(RouterLink, { optional: true, self: true });\n  readonly locationStrategy = inject(LocationStrategy, { optional: true });\n  readonly router = inject(Router, { optional: true });\n\n  handleClick(e: MouseEvent) {\n    if (!this.routerLink || this.routerLink.urlTree === null) return;\n    if (e.button !== 0 || e.ctrlKey || e.shiftKey || e.altKey || e.metaKey || (this.target() && this.target() !== "_self")) {\n      e.stopPropagation();\n    } else {\n      e.preventDefault();\n    }\n  }\n\n  computeRouterHref(urlTree) {\n    return urlTree != null && this.locationStrategy && this.router ? (this.locationStrategy?.prepareExternalUrl(this.router.serializeUrl(urlTree)) ?? "") : null;\n  }\n\n  constructor() {}'
+		]
+	];
+
+	for (const [from, to] of replacements) {
+		if (!code.includes(from)) {
```
milan-w, line 30: I don't see how you ensure that this transformation is only applied when a db-link contains a routerLink attr. Which might very well be, because I am not deeply familiar with Mitosis.  
But if there is no check, this part prevents using db-link for all other kind of links, which is probably not what you want.

jonaskuske: The transformation is always applied, but the code is a noop when no routerLink is used.

`inject(RouterLink, {self: true, optional: true })` will not inject anything if there is no RouterLink directive present on the component instance, so `computeRouterHref` will return `null` and the component uses the `href` prop instead, and the click handler also returns immediately if `!this.routerLink`

nmerget (approved):

PR #7538, opened 2026-08-06
fix(eslint-plugin): rules don't handle Angular template expressions as element content

### LINKED ISSUE #7117: Lint rules don't handle Angular template expressions as element content

- [ ] All
- [ ] HTML
- [ ] React
- [x] Angular
- [ ] Vue
- [ ] Web components
- [ ] Power Apps

### Reproduction case

*db-ux/text-or-children-required* doesn't allow template expressions:

```html
<db-link routerLink="/account">{{ something }}</db-link>
<!-- db-link must have either a text property or children content eslint(db-ux/text-or-children-required) -->
```

Due to the structural directive `*dbNavigationContent`, this also is an expression rather than a simple child item:

```html
<db-navigation-item backButtonText="Back">
  <a *dbNavigationContent routerLink="/overview">Home</a>
</db-navigation-item>
<!-- db-navigation-item must have either a text property or children content eslint(db-ux/text-or-children-required) -->
```

*db-ux/select-requires-options* has the same problem:

```html
<db-select label="Something">
@for (opt in ["a", "b"]; track opt) {
  <option [value]="opt">{{ opt }}</option>
}
</db-select>
<!-- DBSelect must have either an options property or <option> children eslint(db-ux/select-requires-options) -->
```

### Expected Behaviour

ESLint sees that there are children and doesn't raise an error.

### Screenshots

_No response_

### Browser version

None

_No response_

DB Fernverkehr

_No response_

mfranzke:
### Root cause

The Angular `hasChildren` checks in both rules only recognized `Text`, `Element`, and `Element$1` AST node types. The Angular template parser uses different node types for template expressions:
- `BoundText` — for `{{ interpolation }}`
- `Template` — for structural directives like `*dbNavigationContent`, `*ngIf`, `*ngFor`
- `ForLoopBlock` — for `@for` control flow
- `IfBlock` — for `@if` control flow
- `SwitchBlock` — for `@switch` control flow
- `DeferredBlock` — for `@defer` blocks

### Fix

Added these 6 node types to the children detection logic in both rules:
- `text-or-children-required` — Angular handler's `hasChildren` check
- `select-requires-options` — the shared `hasOptionChildren` function

### Possible followup

Note that `select-requires-options` is a slightly more permissive fix: any of these template expression types counts as "has options", even a `BoundText` interpolation which couldn't actually produce `<option>` elements. This matches the spirit of the rule (if you have dynamic content, we trust you know what you're doing) and avoids false positives. If you'd prefer stricter checking (only `ForLoopBlock`/`IfBlock`/`Template` for select), let me know.

resolves https://github.com/db-ux-design-system/core-web/issues/7117

### Code Quality

- [ ] No hardcoded values, magic numbers, or debug code left in

### Validation

### General

- [ ] The PR title follows the conventional commits format (e.g. `feat:`, `fix:`, `chore:`), as in [Git Commit Conventions](docs/conventions.md#git-commits-conventions)

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/7117-lint-rules-dont-handle-angular-template-expressions-as-element-content>


mfranzke:
@codex review

packages/eslint-plugin/src/rules/select/select-requires-options.ts
```diff
@@ -20,7 +20,15 @@ function hasOptionChildren(node: any): boolean {
...
-		return false;
+		// Angular template expressions that may produce <option> children
+		return (
+			child.type === 'BoundText' ||
+			child.type === 'Template' ||
+			child.type === 'IfBlock' ||
+			child.type === 'ForLoopBlock' ||
+			child.type === 'SwitchBlock' ||
+			child.type === 'DeferredBlock'
+		);
```
michaelmkraus, line 31: Treating every `BoundText`, `Template`, or control-flow node as an option introduces false negatives. For example, both `<db-select>{{ plainText }}</db-select>` and `<db-select>@if (show) { <div>Not an option</div> }</db-select>` now pass without a diagnostic, although neither can provide an `<option>`. `BoundText` can only render text, while container nodes should be searched for descendant `<option>` elements instead of being considered sufficient themselves. Could we recursively inspect the relevant children/branches and add invalid regression tests for these cases?

packages/eslint-plugin/test/rules/content/text-or-children-required.spec.ts
```diff
@@ -112,7 +112,16 @@ describe('text-or-children-required', () => {
...
+			{ code: '<db-button [text]="myText"></db-button>' },
+			{
+				code: '<db-link routerLink="/account">{{ something }}</db-link>'
+			},
+			{
+				code: '<db-navigation-item backButtonText="Back"><a *dbNavigationContent routerLink="/overview">Home</a></db-navigation-item>'
+			},
+			{
+				code: '<db-button>{{ dynamicLabel }}</db-button>'
+			}
```
michaelmkraus, line 124: This list is still incomplete for Angular template expressions. Angular ICU/i18n content is represented by an `Icu` node, so a valid component such as `<db-button i18n>{count, plural, =0 {None} other {{{count}} items}}</db-button>` still reports `missingContent`. Could we recognize `Icu` as rendered content?

michaelmkraus, line 124: Angular represents `<ng-content>` as a `Content` AST node rather than an Element, so `<db-button><ng-content /></db-button>` still triggers `missingContent` even though the component receives projected child content. Could we include `Content` in the Angular child detection?

mfranzke:
this is way more complex than I could currently contribute.

PR #7633, opened 2026-08-11
feat: heading component (beta)

### LINKED ISSUE #530: 1️⃣ 👁️ Heading (Beta Release)
### Beschreibung

Dieses Epic umfasst alle Aufgaben zur Implementierung, Dokumentation, Qualitätssicherung und Beta-Veröffentlichung des Inhalts **Heading** für das DB UX Design System.
Ziel ist eine konsistente, barrierefreie und dokumentierte Komponente, die DS Nutzenden als Beta zur Verfügung steht.

### Akzeptanzkriterien

Zielzustand → Ready for Beta Release

✅ High Design Quality (Design Perspective)
✅ High Development Quality (Development Perspective)
✅ Understandable (User Perspective)
✅ Usable (User Perspective)

michaelmkraus:
## Proposed changes

Adds the beta Heading component family: `DBHeadingH1` through `DBHeadingH6` plus `DBCustomHeading`.

The six semantic components render fixed native elements (`h1` through `h6`). Their semantic level therefore cannot drift from the generated markup. `DBCustomHeading` is a styling-only `div` wrapper without heading semantics of its own. Consumers place exactly one native heading or `DBHeadingH1`–`DBHeadingH6` in its default slot; optional `startSlot` and `endSlot` content is rendered as a sibling so it does not become part of the accessible heading name.

All seven components share the visual API:

- independent sizes from `3xl` through `3xs`
- `fontWeight` values `black` and `light`
- logical `alignment`
- optional `paragraphSpacing` of exactly `1lh`
- global attribute forwarding and regular children

Without an explicit `size`, the semantic components use the shared headline-level mapping from the foundations package. The mapping is exposed as `fonts.$headlines`, allowing the foundations defaults and Heading styles to use one source of truth without changing the existing emitted foundation CSS.

A new `custom-heading-single-heading` ESLint rule validates that `DBCustomHeading` contains exactly one heading. It supports React, Vue and Angular templates, counts projected start/end-slot headings, handles transparent framework wrappers such as Vue `template` and Angular `ng-container`, and avoids false positives when consumer components or other runtime-only content may provide the heading.

Heading is the first component family with several generated components sharing one source folder, model, stylesheet and test. The component registry and post-build tooling now support that structure and generate the Angular, React, Vue and Web Component outputs consistently.

Patternhub groups all seven custom elements on one Heading page in the declared order H1–H6 followed by Custom Heading. Generated Web Types link every Heading tag to that canonical page. The feature also includes examples, documentation, showcase and e2e coverage, Storybook stories, Figma Code Connect definitions, framework outputs and Web Component metadata.

The showcase and Patternhub navigation use the title `Heading`, because the family has no single `DBHeading` export. This replaces the unpublished earlier branch draft based on one dynamically switched `DBHeading` component.

resolves db-ux-design-system/core#530

## Validation

- `pnpm run lint`
- `pnpm run build`
- `pnpm run build-outputs`
- `pnpm run test`
- React, Vue and Angular ESLint parser cases for static, projected and runtime-only content
- generated Heading Web Types documentation URLs
- generated Patternhub element order
- Storybook and Figma Code Connect generation/tests
- framework-specific showcase and e2e coverage

## Decisions

- `fontWeight` intentionally uses `black` and `light` to match the current Figma Heading API.
- The Axe `heading-order` rule remains disabled in the component-level checks because showcase composition intentionally demonstrates headings independently; heading semantics are covered by dedicated assertions.
- Duplicate host/inner IDs in generated Web Components are tracked separately in db-ux-design-system/core-web#7729.

### Code Quality

### Validation

### General

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/feat-heading>


mfranzke:
@codex review

packages/components/scripts/post-build/components.ts
```diff
@@ -29,6 +29,10 @@ export type Component = {
 };
 
 export const getComponents = (): Component[] => [
+	{
+		name: 'heading'
+	},
```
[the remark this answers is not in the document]
michaelmkraus: Fixed.

mfranzke:
@codex review

mfranzke:
@michaelmkraus for the screenreader-tests it might be beneficial to use the new methods provided by guidepup with it's latest release: https://redirect.github.com/guidepup/guidepup/pull/134

packages/components/configs/plugins/heading-dynamic-root.cjs
```diff
@@ -0,0 +1,145 @@
...
+		'  componentDidLoad() {',
+		`  componentDidRender() {\n    const element = this.rootElement.querySelector("h1, h2, h3, h4, h5, h6") as HTMLHeadingElement | null;\n    this.copyHeadingAttributes(this._activeHeadingElement, element);\n    this._activeHeadingElement = element;\n    this._ref = element;\n    this.enableAttributePassing(element, "db-heading");\n  }\n\n  componentDidLoad() {`,
+		'stencil'
+	);
+	changed = replaceOnce(changed, '@Prop() as:', '@Prop() as!:', 'stencil');
+	changed = replaceOnce(
+		changed,
+		'@Prop() className:',
+		'@Prop({ attribute: "classname" }) className:',
+		'stencil'
```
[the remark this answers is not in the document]
michaelmkraus: I could not reproduce this in the current generated output. 
`heading.tsx:7` documents children as unnamed slot, while the generated property declarations contain the actual component props but no `@Prop()` children. Therefore no spurious children property or attribute is currently published. I would leave the transform unchanged rather than add a replacement for output that is not generated.

packages/components/configs/plugins/heading-attribute-passing.cjs
```diff
@@ -0,0 +1,198 @@
...
+		headingFile.outputDir,
+		path.dirname(headingFile.outputFilePath),
+		'heading.spec.tsx'
+	);
+	let source = fs.readFileSync(sourceFile, 'utf-8');
+	if (targetContext.target === 'vue') {
+		source = source
+			.replace(/\{\/\*/g, '')
+			.replace(/\*\/}/g, '')
+			.replace(/\/\/ VUE:/g, '');
```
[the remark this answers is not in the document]
michaelmkraus: fixed

packages/components/configs/plugins/heading-dynamic-root.cjs
```diff
@@ -0,0 +1,145 @@
...
+    for (const attr of Array.from(from.attributes)) {
+      if (
+        attr.name === "style" ||
+        attr.name.startsWith("aria-") ||
+        (attr.name.startsWith("data-") && !${JSON.stringify(OWN_DATA_ATTRIBUTES)}.includes(attr.name))
+      ) {
+        to.setAttribute(attr.name, attr.value);
+      } else if (attr.name === "class") {
+        const classes = new Set([...(to.getAttribute("class") ?? "").split(/\\s+/), ...attr.value.split(/\\s+/)]);
+        to.setAttribute("class", [...classes].filter(Boolean).join(" "));
```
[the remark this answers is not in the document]
michaelmkraus: fixed

packages/components/src/components/heading/figma/heading.figma.lite.tsx
```diff
@@ -0,0 +1,22 @@
...
+export default function HeadingFigmaLite(props: FigmaHeadingProps) {
+	return (
+		<DBHeading
+			as={props.as}
+			size={props.size}
+			fontWeight={props.fontWeight}
+			alignment={props.alignment}
+			paragraphSpacing={props.paragraphSpacing}
+			startSlot={<Fragment>{props.startSlot}</Fragment>}
+			endSlot={<Fragment>{props.endSlot}</Fragment>}>
```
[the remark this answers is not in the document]
michaelmkraus: fixed.

packages/components/configs/plugins/heading-attribute-passing.cjs
```diff
@@ -0,0 +1,198 @@
...
+			hooksImport,
+			'import { useRef, forwardRef, HTMLAttributes } from "react";'
+		)
+		.replace(
+			functionDeclaration,
+			'function DBHeadingFn(props: Omit<HTMLAttributes<HTMLHeadingElement | any>, keyof DBHeadingProps> & DBHeadingProps, component: any) {'
+		)
+		.replace(
+			refDeclaration,
+			'const _ref = component || useRef<HTMLHeadingElement | any>(component);'
```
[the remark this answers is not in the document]
michaelmkraus: fixed

packages/components/src/components/heading/docs/HTML.md
```diff
@@ -0,0 +1,48 @@
+## HTML
+
+> **Beta:** `DBHeading` is available as a beta component. Its API and visual design may change before the stable release.
+
+For installation and configuration of the CSS, see [`@db-ux/core-components`](https://www.npmjs.com/package/@db-ux/core-components). For the custom element, see [`@db-ux/wc-core-components`](https://www.npmjs.com/package/@db-ux/wc-core-components).
+
+### Native HTML and CSS class
+
+Use the `db-heading` class on a native heading. Choose the element from the document hierarchy, not from its appearance. Omitting `data-size` applies the default mapping: `h1`/`xl`, `h2`/`lg`, `h3`/`md`, `h4`/`sm`, `h5`/`xs`, and `h6`/`2xs`.
```
mfranzke:
```suggestion
Use the `db-heading` class on a native heading. Choose the element from the document hierarchy, not from its appearance. Omitting `data-size` applies the default mapping: `<h1>`/`xl`, `<h2>`/`lg`, `<h3>`/`md`, `<h4>`/`sm`, `<h5>`/`xs`, and `<h6>`/`2xs`.
```

nitpick: maybe the mapping get's a little clearer

packages/components/src/components/heading/docs/HTML.md
```diff
@@ -0,0 +1,48 @@
...
+	data-paragraph-spacing="true"
+	data-density="expressive"
+>
+	Responsive heading
+</h2>
+```
+
+### Web Component
+
+The Web Component uses required `as` for explicit semantics. Untyped usage defensively falls back to `h1` when `as` is missing or invalid.
```
mfranzke:
```suggestion
The Web Component uses required `as` for explicit semantics. Untyped usage defensively falls back to `h2` when `as` is missing or invalid.
```

wouldn't `h2`probably be better here? An `h1` HTML element should be unique on a page anyhow, and `h2` might be better fitting. But there might be other reasons why you've decided for this.

packages/components/src/components/heading/examples/_heading.arg.types.ts
```diff
@@ -0,0 +1,20 @@
...
+export const StorybookHeadingArgTypes: Record<string, InputType> = {
+	as: {
+		control: 'select',
+		options: ['h1', 'h2', 'h3', 'h4', 'h5', 'h6']
+	},
+	size: {
+		control: 'select',
+		options: ['3xl', '2xl', 'xl', 'lg', 'md', 'sm', 'xs', '2xs', '3xs']
+	},
+	fontWeight: { control: 'select', options: ['black', 'light'] },
```
mfranzke, line 8: `black` sounds a little bit strange as an opposite value to `light` – wouldn't [`bold`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/font-weight) be a better fit (even if this is most likely adapted out of Figma and the underlaying font directly, it might be worth to discuss, as this doesn't need to be brand related in a white-label based Design System) ?

packages/components/src/components/heading/heading.lite.tsx
```diff
@@ -0,0 +1,125 @@
...
+					data-size={props.size}
+					data-font-weight={props.fontWeight}
+					data-alignment={props.alignment}
+					data-paragraph-spacing={getBooleanAsString(
+						props.paragraphSpacing,
+						'paragraphSpacing'
+					)}>
+					<Slot name="startSlot" />
+					{props.children}
+					<Slot name="endSlot" />
```
nmerget: This will not work.

If I pass a `<a>copy<a>` into the endSlot a screen-reader user will get `My Headline link "copy"` as an example.

We should create a `<div>` around this component and move all [data-attributes] and class to the <div>. Then we just need the `id` on h1,h2 etc. and the props.children. 

Also we can control the <div> with a flex-box to align all items in the center

packages/components/configs/plugins/angular/heading-content-projection.cjs
nmerget, line 1: I didn't tought about this...

We can't use multiple props.children inside a component...

This might be a overkill for such a plugin, I think it is possible to to use dynamic tags for Angular, React and Vue directlly with mitosis. But we need to test it, maybe we can get rid of this.

Here's an [example](https://github.com/BuilderIO/mitosis/blob/main/packages/core/src/__tests__/data/blocks/render-block.raw.tsx#L233). Don't now if it is working

packages/components/src/components/heading/examples/_heading.arg.types.ts
michaelmkraus: These are the values that are used in Figma. We should discuss this with the others.

packages/components/src/components/heading/docs/HTML.md
michaelmkraus: Since `as` is required, this fallback only covers invalid or untyped runtime usage. None of `h1` or `h2` can reliably preserve an unknown document hierarchy, so I would keep the current documented and tested `h1` fallback.

showcases/angular-showcase/src/app/utils/navigation-item.ts
```diff
@@ -80,6 +82,11 @@ export const NAVIGATION_ITEMS: NavItem[] = [
 		path: '04',
 		label: '04 Data-Display',
 		subNavigation: getSortedNavigationItems([
+			{
+				path: '04/heading',
+				label: 'Heading',
+				component: HeadingComponent
```
nmerget:
```suggestion
				component: HeadingShowcase
```

We should point to the generated HeadingShowcase directly

packages/components/src/components/heading/heading.lite.tsx
michaelmkraus: The current assumption was that `startSlot` and `endSlot` are part of the heading’s phrasing content: textual content contributes to its accessible name, while decorative content must use `aria-hidden="true"`. We added examples to the showcases. Interactive content was intended only for explicitly reviewed cases, such as the permalink example.

Moving all classes, `data-*,` and `aria-*` attributes to a wrapper would break the native heading’s styling, attribute forwarding, and ref contract. We should first clarify whether these slots should remain part of the heading or represent separate actions. If actions are required, this needs a dedicated wrapper/action-slot design rather than moving heading attributes to a generic `div`.

packages/components/configs/plugins/angular/heading-content-projection.cjs
michaelmkraus: I tested the `<state.tag>` pattern with Mitosis `0.14.0` and Angular signal configuration. React and Vue generate native dynamic elements, but Angular generates `*ngComponentOutlet="tag()"`, with `tag()` returning a string such as `"h1"`. Since `ngComponentOutlet` expects an Angular component type rather than a native tag name, the generated Angular output does not provide a working native dynamic element. For now, we therefore still need the static heading roots and shared `ng-template` projection workaround.

showcases/angular-showcase/src/app/utils/navigation-item.ts
michaelmkraus: `HeadingComponent` intentionally wraps the `HeadingShowcase`. It also provides the Angular-specific hidden fixture used by `heading-dynamic-root.spec.ts` to change the `as` input from `h1` to `h6` at runtime and verify that forwarded attributes survive the native root replacement. Pointing the route directly to `HeadingShowcase` would remove this fixture and break that regression test. I would therefore keep the wrapper unless we move the fixture to a dedicated test route.

mfranzke:
@codex review

packages/components/src/components/heading/heading.spec.tsx
```diff
@@ -0,0 +1,453 @@
...
+		await expect(component).toHaveAccessibleName('Main content and more');
+	});
+
+	test('has the expected custom ARIA heading-level snapshot', async ({
+		mount
+	}, testInfo) => {
+		const component = await mount(customSemanticHeadings);
+		expect(await component.ariaSnapshot()).toMatchSnapshot(
+			`${testInfo.testId}.yaml`
+		);
```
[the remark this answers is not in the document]
michaelmkraus: Fixed. `__snapshots__/heading/component/` now has baselines for both suites in all three projects.

The cause was broader than the two DBCustomHeading assertions: our snapshotPathTemplate has no placeholder for the name passed to `toMatchSnapshot()`, so the path comes purely from the describe and test titles. Renaming the describe from `DBHeading` to `Heading` invalidated the static baselines too. Also removed 18 orphaned `DBHeading-*` files.

packages/components/src/components/heading/figma/heading.figma.lite.tsx
```diff
@@ -0,0 +1,71 @@
...
+import DBHeadingH5 from '../heading-h5.lite';
+import DBHeadingH6 from '../heading-h6.lite';
+import { FigmaHeadingProps, headings } from './heading.figma';
+
+useMetadata({ figma: headings });
+
+export default function HeadingFigmaLite(props: FigmaHeadingProps) {
+	return (
+		<>
+			<Show when={props.level === 'h1'}>
```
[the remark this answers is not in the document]
michaelmkraus: Confirmed, deferring on purpose. The Figma library for Heading is being restructured, so the `As` property and node IDs this mapping relies on are not final yet.

Note: the `parse.spec.ts.snap` suites pass because they record whatever is generated — they do not prove the snippet is usable. The new fail-fast in the multi-component Figma injection is scoped to `validationMessage/conditionalProp/nestedInstancesToArray` entries, since an enum prop that only drives `Show` conditions legitimately has no injection target.

packages/components/src/components/heading/custom-heading.lite.tsx
```diff
@@ -0,0 +1,27 @@
...
+export default function DBCustomHeading(props: DBCustomHeadingProps) {
+	const _ref = useRef<HTMLDivElement | any>(null);
+
+	return (
+		<div
+			ref={_ref}
+			id={props.id ?? props.propOverrides?.id}
+			class={cls('db-heading', props.className)}
+			role="heading"
+			aria-level={props.semanticLevel ?? 2}
```
nmerget: Why do we need this here?

The CustomHeading should look like this:

``<DBCustomHeading><h1 id="headline1">MyHeadline</h1><a href="#headline1">Link to headline</a></DBCustomHeading>``

The purpose is that the user can control thei own children and just get the stylings from the wrapping <div>.

There is not need to set the role="heading" and therefore we don'T need all the plugins to move any aria-* or other props to the children

packages/components/src/components/heading/heading.scss
```diff
@@ -0,0 +1,65 @@
...
+		@extend %db-overwrite-headline-size-sm;
+	}
+
+	&:not([data-size]):where(h5, [role="heading"][aria-level="5"]) {
+		@extend %db-overwrite-headline-size-xs;
+	}
+
+	&:not([data-size]):where(h6, [role="heading"][aria-level="6"]) {
+		@extend %db-overwrite-headline-size-2xs;
+	}
```
nmerget: We already define this inside the `packages/foundations/scss/defaults/default-fonts.scss` globally, do we need this again?

packages/components/src/components/heading/heading.scss
```diff
@@ -0,0 +1,65 @@
+@use "@db-ux/core-foundations/build/styles/colors";
+@use "@db-ux/core-foundations/build/styles/fonts";
+
+.db-heading {
```
nmerget:
```suggestion
%heading-placeholder{
```

We should create a placeholder and use it for the `.db-heading` and have another `.db-custom-heading` which applies the styles to the children. Also we need to add a flex layout to the .db-custom-heading with some gap to showcase the `LinkButton` example

showcases/angular-showcase/src/app/utils/navigation-item.ts
nmerget: We don't need the heading.component.ts, please use the HeadingShowcase instead

packages/components/src/components/heading/examples/permalink.example.lite.tsx
```diff
@@ -0,0 +1,44 @@
...
+				id="custom-installation"
+				class="heading-permalink"
+				aria-label="Custom installation">
+				<span>Custom installation</span>
+				<a
+					class="db-link heading-permalink-link"
+					href="#custom-installation"
+					aria-label="Direct link to Custom installation">
+					<span aria-hidden="true">#</span>
+				</a>
```
nmerget: Let's provide a `DBCustomButton` here as a ghost button with an `link_chain` icon

nmerget:
A nice addition would be to add a eslint rule that `DBCustomHeading` needs one h1-6, and there shouldn't be more than one h1-6

packages/components/src/components/heading/examples/permalink.example.lite.tsx
michaelmkraus: I'll use the `DBCustomButton` with `link_chain`.

packages/components/src/components/heading/custom-heading.lite.tsx
michaelmkraus: `DBCustomHeading` is now a layout-only wrapper following the `DBCustomButton` pattern: a plain `div.db-custom-heading`, no `role`, no `aria-level`, `semanticLevel` removed. The consumer nests the native heading plus siblings, exactly as in your example.

This also removed all the per-target plugin code: `transformCustomHeadingAttributePassing`, the React `filterPassingProps` exception, the Vue `inheritAttrs: false` handling, the Angular `input.required` and the Stencil `!`. Angular and Stencil need no heading-specific transform at all now.

packages/components/src/components/heading/heading.scss
michaelmkraus: The `[role="heading"][aria-level="N"]` selectors are gone with the ARIA heading, so only the native `h1-h6` mapping remains.

That part has to stay: `default-fonts.scss` is an opt-in import consumers may strip while the component styles must work standalone, and `%db-overwrite-headline-size-*` also sets the icon custom properties on the element itself, whereas `default-fonts.scss` only sets them on `::before/::after`.

packages/components/src/components/heading/heading.scss
michaelmkraus: `%heading-base` now holds the shared declarations, and `.db-custom-heading` is a flex row with `gap: $db-spacing-fixed-sm` that applies the default level mapping to a direct child `h1-h6`, so a bare heading needs no `db-heading` class.

packages/components/src/components/heading/figma/heading.figma.lite.tsx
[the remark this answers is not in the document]
michaelmkraus: Adapted Figma Code Connect.

packages/components/src/components/heading/examples/_heading.arg.types.ts
mfranzke: @michaelmkraus what was the outcome of that discussion ?

mfranzke:
@codex review

packages/components/configs/plugins/heading-attribute-passing.cjs
```diff
@@ -0,0 +1,212 @@
...
+	const openingTag = getOpeningTag(code, heading.tag, componentName, 'vue');
+	const refIndent = openingTag.match(/\n(\s*)ref="_ref"/)?.[1];
+	const classBinding = openingTag.match(/:class="([^"]*className[^"]*)"/);
+	if (refIndent === undefined || classBinding?.length !== 2) {
+		fail(componentName, 'vue', `unexpected ${heading.tag} root attributes`);
+	}
+	const changed = replaceMarker(
+		openingTag,
+		'ref="_ref"',
+		`ref="_ref"\n${refIndent}v-bind="$attrs"`,
```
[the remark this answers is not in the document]
michaelmkraus: Not reproducible. I probed an undeclared `@click` against Vue 3.5.41: our current pattern, automatic fallthrough only, and `inheritAttrs`: false all fire the handler exactly once, with identical rendered attributes.

`mergeProps` only concatenates `on*` keys when `existing !== incoming`. Both sides read the same function reference from the same `$attrs`, so the listener is stored once. Double invocation would need two different references, which cannot happen here.

packages/components/src/components/heading/heading-h1.lite.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+useMetadata({});
+
+export default function DBHeadingH1(props: DBHeadingH1Props) {
+	const _ref = useRef<HTMLHeadingElement | any>(null);
+
+	return (
+		<h1
+			ref={_ref}
+			id={props.id ?? props.propOverrides?.id}
```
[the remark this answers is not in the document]
michaelmkraus: Confirmed, but not heading specific and not introduced here: id is a prop rendered on the inner element in 42 components and `enableAttributePassing` never moves it off the host. I probed 10 components in the built Web Components — all duplicate the id, `getElementById` always resolves to the host.

The impact is worse than described. With 3000px of spacer around the heading:

```
display:inline    scrollY 0 -> 3028   (fragment navigation works)
display:contents  scrollY 0 -> 0      (nothing happens)
```

Angular bakes `:host { display: contents; }` into all 356 components, so it is always affected; WC as soon as `wc-workarounds.css` is loaded. React and Vue are fine.

It does break the new permalink example in the Angular and WC showcases, so worth fixing — but in the shared attribute-passing helper, which touches every component and its snapshots. Opening a separate issue instead of widening this PR.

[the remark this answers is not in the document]
michaelmkraus: Ticket [is here](https://github.com/db-ux-design-system/core-web/issues/7729)

packages/components/src/components/heading/custom-heading.lite.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+export default function DBCustomHeading(props: DBCustomHeadingProps) {
+	const _ref = useRef<HTMLDivElement | any>(null);
+
+	return (
+		<div
+			ref={_ref}
+			id={props.id ?? props.propOverrides?.id}
+			class={cls('db-custom-heading', props.className)}
+			data-alignment={props.alignment}>
```
nmerget:
```suggestion
			data-alignment={props.alignment}
			data-size={props.size}
			data-font-weight={props.fontWeight}
			data-alignment={props.alignment}
			data-paragraph-spacing={getBooleanAsString(
				props.paragraphSpacing,
				'paragraphSpacing'
			)}
			>
```

It is missing the other props

packages/components/src/components/heading/examples/permalink.example.lite.tsx
```diff
@@ -0,0 +1,28 @@
...
+	storybookComponentName: 'DBCustomHeading',
+	storybookComponentNames: ['DBCustomHeading'],
+	storybookNames: ['Heading with a permalink button'],
+	storybookArgTypes: StorybookHeadingArgTypes
+});
+
+export default function HeadingPermalink() {
+	return (
+		<DBCustomHeading class="heading-permalink">
+			<DBHeadingH2 id="installation">Installation</DBHeadingH2>
```
nmerget:
```suggestion
			<h2 id="installation">Installation</h2>
```

As a user I want to throw in a plain `<h*>` tag. So the DBCustomHeader acts just as a wrapper for native html to apply the styling. It's the same concept like <DBButton> and <DBCustomButton>, the <DBButton> always has a pre-defined render with `<button>` if a user just needs the styling they get it via <DBCustomButton>. Therefore, we provide every property from DBButton as well for DBCustomButton

packages/components/configs/plugins/heading-attribute-passing.cjs
nmerget, line 1: Why do we still need this plugins?

packages/components/src/components/heading/heading.scss
```diff
@@ -0,0 +1,118 @@
...
+ * the icon custom properties on the element itself instead of only on
+ * `::before`/`::after`. */
+$level-sizes: (
+	1: "xl",
+	2: "lg",
+	3: "md",
+	4: "sm",
+	5: "xs",
+	6: "2xs"
+);
```
nmerget: We should define this inside the foundations, and reuse it here and inside the /required/default-fonts.scss. There is already a `$headlines` variable we can reuse them here.

packages/components/src/components/heading/heading.scss
```diff
@@ -0,0 +1,118 @@
...
+	> :where(
+		db-heading-h-1,
+		db-heading-h-2,
+		db-heading-h-3,
+		db-heading-h-4,
+		db-heading-h-5,
+		db-heading-h-6
+	) {
+		display: contents;
+	}
```
nmerget: `packages/components/src/styles/wc-workarounds.scss` is using `packages/components/src/styles/internal/_custom-elements.scss`, where you already include those tags. I guess we don't need this here again

packages/components/src/components/heading/examples/children-accessibility.example.lite.tsx
```diff
@@ -0,0 +1,31 @@
...
+	return (
+		<Fragment>
+			<DBHeadingH2>
+				<span aria-hidden="true">* </span>
+				<span>Current disruptions</span>
+				<span aria-hidden="true"> *</span>
+			</DBHeadingH2>
+			<DBCustomHeading>
+				<DBHeadingH2>Current disruptions</DBHeadingH2>
+				<span>3 active</span>
```
nmerget: A DBBage might be a good way to preivew this example

packages/components/src/components/heading/examples/_heading.arg.types.ts
nmerget: @mfranzke look at this table:

https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/font-weight#common_weight_name_mapping

`black` is a common name

mfranzke: > @mfranzke look at this table:
> 
> https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/font-weight#common_weight_name_mapping
> 
> `black` is a common name

that's right. But it doesn't fit that official [OpenType specification](https://learn.microsoft.com/en-us/typography/opentype/spec/os2#usweightclass) mapping:
Our chosen value `700` would be called `bold`, whereas `black` would have a font-weight of `900`.

packages/components/src/components/heading/showcase/heading.showcase.lite.tsx
```diff
@@ -0,0 +1,78 @@
...
+import HeadingParagraphSpacing from '../examples/paragraph-spacing.example.lite';
+// The permalink example is kept but hidden from the showcases for now.
+// import HeadingPermalink from '../examples/permalink.example.lite';
+import HeadingSemanticLevels from '../examples/semantic-levels.example.lite';
+import HeadingSizes from '../examples/sizes.example.lite';
+
+export default function HeadingShowcase(props: PatternhubProps) {
+	return (
+		<ContainerWrapperShowcase
+			title="Heading"
```
mfranzke:
```suggestion
			title="DBHeading"
```

either this is just an inconsistency, or we've decided not to use the `DB` in front of the components in the navigation anymore (but than we need to change it for the others as well to keep it consistent).

packages/components/src/components/heading/heading.scss
```diff
@@ -0,0 +1,138 @@
...
+@use "@db-ux/core-foundations/build/styles/variables";
+@use "@db-ux/core-foundations/build/styles/helpers";
+@use "@db-ux/core-foundations/build/styles/fonts";
+
+/* `fonts.$headlines` is the shared level-to-size mapping, also used by the
+ * foundations' `defaults/default-fonts.scss`. Repeating the mapping here is
+ * intentional: that file is an opt-in import consumers may strip while the
+ * component styles have to work standalone, and `%db-overwrite-headline-size-*`
+ * additionally sets the icon custom properties on the element itself instead of
+ * only on `::before`/`::after`. */
```
nmerget: This comment is wrong at this place right?

packages/components/src/components/heading/heading.scss
```diff
@@ -0,0 +1,138 @@
...
+	/* The size lands on the wrapper as well, so `1lh` for
+	 * `data-paragraph-spacing` resolves from the heading typography instead of
+	 * the surrounding body text. The nested heading needs it explicitly too,
+	 * because the user-agent styles for `h1`-`h6` override inheritance. */
+	@each $tag, $size in fonts.$headlines {
+		&:not([data-size]):has(:where(#{$tag})),
+		&:not([data-size]) > :where(#{$tag}):not(.db-heading) {
+			@extend %db-overwrite-headline-size-#{$size};
+		}
+	}
```
nmerget:
```suggestion
&:not([data-size]){
	@each $tag, $size in fonts.$headlines {
		&:has(#{$tag}){
  			&, #{$tag}{
  				@extend %db-overwrite-headline-size-#{$size};
  			}
		}
	}
}
```

Why do we need the `:where` we select just one tag don't we?

So if we find the tag we apply it to the parent <div> and the headline tag

packages/components/src/components/heading/heading.scss
```diff
@@ -0,0 +1,138 @@
...
+
+	@each $size in fonts.$font-sizes {
+		&[data-size="#{$size}"],
+		&[data-size="#{$size}"] > #{$nested-heading} {
+			@extend %db-overwrite-headline-size-#{$size};
+		}
```
nmerget:
```suggestion
		&[data-size="#{$size}"]{
			&, #{$nested-heading} {
				@extend %db-overwrite-headline-size-#{$size};
			}
		}
```

We should avoid the `>` in this case because of angular/stencil. A user might create a own component like <my-super-heading> which renders a normal `<h*>` but would have the wrapper around it

packages/components/src/components/heading/heading.scss
```diff
@@ -0,0 +1,138 @@
...
+	}
+
+	&[data-font-weight="light"],
+	&[data-font-weight="light"] > #{$nested-heading} {
+		font-weight: 300;
+	}
```
nmerget:
```suggestion
	&[data-font-weight="light"]{
		&, #{$nested-heading} {
			font-weight: 300;
		}
	}
```

We should avoid the > in this case because of angular/stencil. A user might create a own component like which renders a normal <h*> but would have the wrapper around it

packages/components/src/components/heading/heading.scss
```diff
@@ -0,0 +1,138 @@
...
+/* `$child` additionally applies the alignment to a nested heading, which is
+ * required because `%heading-base` sets an explicit `text-align: start` there
+ * that would otherwise block inheritance. */
+@mixin text-alignments($child: null) {
+	@each $alignment, $value in $logical-alignments {
+		@if $child {
+			&[data-alignment="#{$alignment}"],
+			&[data-alignment="#{$alignment}"] > #{$child} {
+				text-align: $value;
+			}
```
nmerget:
```suggestion
			&[data-alignment="#{$alignment}"]{
				&, #{$child} {
					text-align: $value;
				}
			}
```

We should avoid the > in this case because of angular/stencil. A user might create a own component like which renders a normal <h*> but would have the wrapper around it

packages/components/src/components/heading/examples/children-accessibility.example.lite.tsx
michaelkraus: Done, using DBBadge semantic="critical" emphasis="strong".

packages/components/src/components/heading/heading.scss
michaelkraus: Done. The mapping is now in the new _headline-levels.scss and is consumed by both default-fonts.scss and heading.scss as fonts.$headlines.

packages/components/configs/plugins/heading-attribute-passing.cjs
michaelkraus: Still needed, but only for React and Vue now.

packages/components/src/components/heading/examples/permalink.example.lite.tsx
michaelkraus: Done, all examples now use a plain `<h2>`. The wrapper styles it, no class needed.

packages/components/src/components/heading/custom-heading.lite.tsx
michaelkraus: You're right. I checked it: DBCustomButtonProps mirrors DBButton's styling props and drops only what belongs to the native `<button>`. size, fontWeight, paragraphSpacing and alignment are styling, so they belong on the wrapper.

packages/components/src/components/heading/heading.scss
michaelkraus: You're right, removed.

packages/components/src/components/heading/showcase/heading.showcase.lite.tsx
michaelkraus: You're right, adapted it.

michaelmkraus: You're right, adapted it.

packages/components/src/components/heading/examples/children-accessibility.example.lite.tsx
michaelmkraus: Done, using DBBadge semantic="critical" emphasis="strong".

packages/components/src/components/heading/heading.scss
michaelmkraus: You're right, removed.

packages/components/src/components/heading/heading.scss
michaelmkraus: Done. The mapping is now in the new _headline-levels.scss and is consumed by both default-fonts.scss and heading.scss as fonts.$headlines.

packages/components/configs/plugins/heading-attribute-passing.cjs
michaelmkraus: Still needed, but only for React and Vue now.

packages/components/src/components/heading/examples/permalink.example.lite.tsx
michaelmkraus: Done, all examples now use a plain `<h2>`. The wrapper styles it, no class needed.

packages/components/src/components/heading/custom-heading.lite.tsx
michaelmkraus: You're right. I checked it: DBCustomButtonProps mirrors DBButton's styling props and drops only what belongs to the native `<button>`. size, fontWeight, paragraphSpacing and alignment are styling, so they belong on the wrapper.

packages/components/src/components/heading/docs/HTML.md
```diff
@@ -0,0 +1,73 @@
...
+	</div>
+</db-custom-heading>
+```
+
+The equivalent CSS-only markup:
+
+```html
+<div class="db-custom-heading" data-size="xl" data-font-weight="light">
+	<h2 id="installation">Installation</h2>
+	<a href="#installation">Direct link to Installation</a>
```
mfranzke:
```suggestion
	<div class="db-custom-button" data-icon="link_chain" data-no-text="true" data-variant="ghost">
		<a href="#installation">Direct link to Installation</a>
	</div>
```

packages/components/src/components/heading/heading.scss
michaelmkraus: You're right, it was left over from before the mapping moved into the foundations.

packages/components/src/components/heading/heading.scss
michaelmkraus: Restructured as you suggested, but the `:where()` has to stay — dropping it broke a test.

packages/components/src/components/heading/heading.scss
michaelmkraus: Done, exactly as suggested

packages/components/src/components/heading/heading.scss
michaelmkraus: Done, applied as suggested.

packages/components/src/components/heading/heading.scss
michaelmkraus: Done, applie as suggested.

packages/components/src/components/heading/docs/HTML.md
michaelmkraus: Done.

packages/components/src/components/heading/index.ts
```diff
@@ -0,0 +1,7 @@
+export { default as DBCustomHeading } from './custom-heading';
+export { default as DBHeadingH1 } from './heading-h1';
+export { default as DBHeadingH2 } from './heading-h2';
+export { default as DBHeadingH3 } from './heading-h3';
+export { default as DBHeadingH4 } from './heading-h4';
+export { default as DBHeadingH5 } from './heading-h5';
+export { default as DBHeadingH6 } from './heading-h6';
```
nmerget, line 7: Those exports don't work right now.

We need to add them inside `packages/components/scripts/post-build/components.ts` and update the path for `indexFile` inside `packages/components/scripts/post-build/angular.ts` and `packages/components/scripts/post-build/stencil.ts`. So maybe we should update the `export type Component` to overwrite the the path for the index file

nmerget (dismissed):

.changeset/tidy-headlines-share-levels.md
```diff
@@ -0,0 +1,15 @@
+---
+"@db-ux/core-foundations": minor
+"@db-ux/core-components": minor
+"@db-ux/ngx-core-components": minor
+"@db-ux/react-core-components": minor
+"@db-ux/v-core-components": minor
+"@db-ux/wc-core-components": minor
```
mfranzke:
```suggestion
```

This is exclusively happening in `@db-ux/core-foundations`

packages/components/src/components/heading/heading-h1.lite.tsx
```diff
@@ -0,0 +1,27 @@
...
+			class={cls('db-heading', props.className)}
+			data-size={props.size}
+			data-font-weight={props.fontWeight}
+			data-alignment={props.alignment}
+			data-paragraph-spacing={getBooleanAsString(
+				props.paragraphSpacing,
+				'paragraphSpacing'
+			)}>
+			{props.children}
```
nmerget, line 30:
```suggestion
<Show when={props.text}>{props.text}</Show>
			{props.children}
```

We shoul add TextProps as alternative to align it with other similar components like DBInfotext etc.

nmerget:
image
We need to make sure that our components doesn't break inside a custom-heading

packages/components/src/components/heading/heading-h1.lite.tsx
michaelmkraus: Good point. Implemented.

.changeset/soft-headings-speak.md
nmerget, line 1: Do we need a own changeset for this?

michaelmkraus: You’re right, the separate changeset is not needed. Removed it.

nmerget (dismissed):

nmerget (dismissed):

.changeset/bright-headings-arrive.md
```diff
@@ -0,0 +1,13 @@
+---
+"@db-ux/core-components": minor
+"@db-ux/ngx-core-components": minor
+"@db-ux/react-core-components": minor
+"@db-ux/wc-core-components": minor
+"@db-ux/v-core-components": minor
+---
+
+feat: add beta static and custom Heading components
```
mfranzke:
```suggestion
feat: add static and custom Heading components (beta)
```

maybe we add the state to the end of that phrase so that it's clearer / the name and status are separated.

.changeset/tidy-headlines-share-levels.md
```diff
@@ -0,0 +1,10 @@
+---
+"@db-ux/core-foundations": minor
+---
+
+refactor: expose the headline level mapping as `fonts.$headlines`
```
mfranzke:
```suggestion
refactor(SASS): expose the headline level mapping as `fonts.$headlines`
```

nmerget (approved):

mfranzke:
@codex review

packages/components/scripts/post-build/components.ts
```diff
@@ -28,7 +39,35 @@ export type Component = {
...
+	name,
+	folder: 'heading',
+	spec: index === 0 ? 'heading' : undefined,
+	overwrites: {
+		vue: [{ from: 'props.class', to: 'props.className ?? props.class' }]
+	}
+}));
+
 export const getComponents = (): Component[] => [
+	...headingComponents,
```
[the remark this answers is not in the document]
mfranzke: https://github.com/db-ux-design-system/core-web/issues/7849

PR #7643, opened 2026-08-12
fix: resolve multiple DBPopover issues with controlled open state

### LINKED ISSUE #6176: DBPopover hides when moving mouse onto it (sometimes)

- [x] All
- [ ] HTML
- [ ] React
- [ ] Angular
- [ ] Vue
- [ ] Web components
- [ ] Power Apps

### Reproduction case

- Visit https://design-system.deutschebahn.com/core-web/version/v5.1.1/components/utilities/popover/overview
- Hover over the first example trigger button
- Move the mouse down in order to interact with the popover content
- If you move the mouse too slowly, when the mouse enters the gap between trigger and popover, the popover hides

### Expected Behaviour

While moving the mouse from the trigger to the popover content, the popover should not close

### Screenshots

Image

### Browser version

None

Thanks for maintaining this :)

None

_No response_

_No response_

### LINKED ISSUE #6510: DBPopover - wrong content position on programatic "open" prop

- [ ] All
- [ ] HTML
- [x] React
- [ ] Angular
- [ ] Vue
- [ ] Web components
- [ ] Power Apps

### Reproduction case

The DBPopover component calculates its children's position wrongly if the "open" prop is programatically controlled.
Example: https://stackblitz.com/edit/db-ux-design-system-examples-mglwgmth?file=src%2FApp.tsx

If the default behaviour is used (on hover), the children's position calculation seems to be working fine.

### Expected Behaviour

The children's position is always calculated properly.

### Screenshots

_No response_

### Browser version

None

_No response_

None

_No response_

_No response_

### LINKED ISSUE #7408: DBPopover changes aria-expanded independently from open state

- [x] All
- [ ] HTML
- [ ] React
- [ ] Angular
- [ ] Vue
- [ ] Web components
- [ ] Power Apps

### Reproduction case

```ts
@Component({
  selector: 'app-repro',
  template: `
    <db-popover [open]="isOpen()">
      <db-button trigger (click)="isOpen.update(open => !open)">Toggle</db-button>
      <span>I am the content</span>
    </db-popover>
  `,
  imports: [DBPopover],
})
export class Repro extends Component {
  readonly isOpen = signal(false)
}
```

The button will have `aria-expanded=true` when I hover over it, even though the dropdown is not open. When the dropdown is open but I don't hover over it, it will falsely have `aria-expanded=false`.

### Expected Behaviour

The button has `aria-expanded=true` when (and only when) the dropdown is open.

### Screenshots

_No response_

### Browser version

None

_No response_

DB Fernverkehr

_No response_

nmerget:
## Proposed changes

Fixes multiple interrelated issues with the `DBPopover` component when used in controlled mode (with the `open` prop):

1. **aria-expanded now follows `open` prop** — In controlled mode, `aria-expanded` on the trigger is derived from `props.open` instead of internal hover/focus state. In uncontrolled mode, it continues to follow hover/focus as before.

2. **Hover/focus handlers skip in controlled mode** — `handleEnter` and `handleLeave` return early when `props.open` is defined, preventing internal state from overriding the externally controlled visibility.

3. **Auto-placement on programmatic open** — A new `onUpdate` hook triggers `handleAutoPlacement()` when `props.open` changes, ensuring correct positioning when the popover is opened programmatically.

4. **Gap size accounts for border width** — The SCSS `$popover-gap-size` calculation now includes border width, preventing the popover from hiding when the mouse moves slowly across the gap between trigger and content.

resolves #7408
resolves #6176
resolves #6510

### Code Quality

### Validation

### General

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/fix-popover>


mfranzke (dismissed):

mfranzke:
@codex review

packages/components/src/components/popover/popover.lite.tsx
```diff
@@ -204,10 +212,22 @@ export default function DBPopover(props: DBPopoverProps) {
 		if (_ref) {
 			const child = state.getTrigger();
 			if (child) {
-				child.ariaExpanded = Boolean(state.isExpanded).toString();
+				// In controlled mode (open prop set), aria-expanded follows open;
+				// otherwise it follows internal hover/focus state
+				const expanded =
+					props.open !== undefined
+						? Boolean(props.open)
+						: Boolean(state.isExpanded);
```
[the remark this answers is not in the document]
nmerget: Fixed in c342216. Now using `getBoolean(props.open, 'open')` which correctly handles `open="false"` as falsy, instead of `Boolean(props.open)` which would treat any non-empty string as truthy.

packages/components/src/components/popover/popover.lite.tsx
```diff
@@ -54,6 +54,10 @@ export default function DBPopover(props: DBPopoverProps) {
 			}
 		},
 		handleEnter(): void {
+			if (props.open !== undefined) {
+				return;
+			}
```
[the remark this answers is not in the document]
nmerget: Fixed in c342216. Added a `manualOpen` parameter to `handleEnter`/`handleLeave` so the controlled-open `onUpdate` reuses the same observer registration logic. When `props.open` becomes true, `handleEnter(undefined, true)` is called — registering scroll, resize, and intersection observers. When it becomes false, `handleLeave(undefined, true)` tears them down. No more duplicated observer code.

packages/components/src/styles/internal/_popover-component.scss
```diff
@@ -4,7 +4,9 @@
 @use "@db-ux/core-foundations/build/styles/screen-sizes";
 @use "component";
 
-$popover-gap-size: var(--db-popover-distance);
+$popover-gap-size: calc(
+	var(--db-popover-distance) + #{variables.$db-border-width-3xs} * 2
+);
 $popover-inset: calc(100% + #{$popover-gap-size} * var(--db-popover-gap, 0));
```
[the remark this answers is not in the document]
nmerget: Fixed in c342216. Separated the two concerns: `$popover-gap-size` (with border width) is now only used for the hover bridge `::before` pseudo-element, while `$popover-inset` uses the raw `var(--db-popover-distance)` to stay aligned with `handleFixedPopover`'s JS positioning. No more 2px jump when JS placement initializes.

mfranzke:
@codex review

packages/components/src/components/popover/popover.lite.tsx
```diff
@@ -204,10 +218,24 @@ export default function DBPopover(props: DBPopoverProps) {
...
 			}
 		}
-	}, [_ref, state.isExpanded]);
+	}, [_ref, state.isExpanded, props.open]);
+
+	onUpdate(() => {
+		if (getBoolean(props.open, 'open')) {
+			state.handleEnter(undefined, true);
+		} else if (props.open !== undefined) {
+			state.handleLeave(undefined, true);
```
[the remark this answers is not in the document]
nmerget: fixed

mfranzke:
@codex review

packages/components/src/components/popover/popover.lite.tsx
```diff
@@ -53,7 +58,11 @@ export default function DBPopover(props: DBPopoverProps) {
 				state.handleAutoPlacement();
 			}
 		},
-		handleEnter(): void {
+		handleEnter(_parent?: HTMLElement, manualOpen?: boolean): void {
+			if (!manualOpen && props.open !== undefined) {
```
michaelmkraus: `null` is normalized as “unset” by both `getBoolean()` and `getBooleanAsString()`, but this guard treats it as controlled because `null !== undefined`. With a nullable Angular/Vue binding (or untyped React usage), `data-open` is omitted so CSS reveals the popover on hover/focus, while `handleEnter` returns, `aria-expanded` stays `false`, and the positioning observers are never registered. Please derive controlled mode from the normalized value (or treat both `undefined` and `null` as uncontrolled) consistently in this guard, `handleLeave`, and the ARIA branch.

packages/components/src/components/popover/popover.lite.tsx
```diff
@@ -204,10 +218,24 @@ export default function DBPopover(props: DBPopoverProps) {
...
+					props.open !== undefined
+						? Boolean(getBoolean(props.open, 'open'))
+						: Boolean(state.isExpanded);
+				child.ariaExpanded = expanded.toString();
 			}
 		}
-	}, [_ref, state.isExpanded]);
+	}, [_ref, state.isExpanded, props.open]);
+
+	onUpdate(() => {
```
michaelmkraus, line 236: This new controlled-state effect has no direct regression coverage; the existing Popover spec only checks uncontrolled focus/hover and screenshots. Please add component tests for `open=true`, `open=false`/`open="false"`, ARIA synchronization, hover/focus suppression, programmatic positioning, and transitions back to uncontrolled mode (including pointer/focus remaining inside). Those assertions would have caught both currently open state-transition defects.

michaelmkraus (commented):
🤖 **AI Review** — Fachlich sind Änderungen erforderlich.

- 🔴 **major:** `null` wird inkonsistent als kontrollierter Zustand behandelt; Details im neuen Inline-Kommentar.
- 🔴 **major:** Der bereits offene, nicht aufgelöste Thread zum Wechsel von controlled zu uncontrolled bleibt reproduzierbar. Ich dupliziere ihn nicht.
- 🟡 **minor:** Für die neuen kontrollierten Zustandsübergänge fehlt direkte Regressionstest-Abdeckung.

Der Review ist an `6629a8c9d50458c91014719d59df7c0439374e71` gebunden. GitHub-Event ist gemäß Repository-Workflow `COMMENT`; fachliche Empfehlung: vor dem Merge korrigieren.

packages/components/src/components/popover/popover.lite.tsx
nmerget: Fixed in 88b4b4f. All `props.open !== undefined` guards are now `props.open != null` (loose equality), so `null` from Angular/Vue nullable bindings is correctly treated as uncontrolled — consistent with how `getBoolean()` and `getBooleanAsString()` already normalize `null` as unset.

packages/components/src/components/popover/popover.lite.tsx
```diff
@@ -204,10 +218,24 @@ export default function DBPopover(props: DBPopoverProps) {
...
+		if (getBoolean(props.open, 'open')) {
+			state.handleEnter(undefined, true);
+		} else {
+			state.handleLeave(undefined, true);
```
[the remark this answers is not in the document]
nmerget: Regarding the "preserve hover state when dropping controlled mode" concern: we considered handling the `open → undefined` transition (dynamically removing the `open` prop while hovering), but this is an extreme edge case that adds significant complexity. In practice, components are either always controlled or always uncontrolled. The `onUpdate` now only reacts when `props.open` is actually set (`!= null`), keeping the logic simple and predictable.

packages/components/src/components/popover/popover.lite.tsx
nmerget: done

packages/components/src/components/popover/popover.lite.tsx
```diff
@@ -204,10 +218,24 @@ export default function DBPopover(props: DBPopoverProps) {
...
```
michaelmkraus, line 244: Could you either explicitly guard/document controlled → uncontrolled transitions as unsupported, or handle the transition consistently? As currently implemented, the popover can remain visible through `:hover` / `:focus-within` while `aria-expanded` is reset and the positioning observers are removed.

nmerget: Added an explicit early-return guard (`if (props.open == null) return;`) plus a comment documenting that controlled → uncontrolled transitions are unsupported by design. Components are expected to be either always controlled or always uncontrolled — switching at runtime is an extreme edge case that adds disproportionate complexity for no practical benefit.

michaelmkraus (approved):

PR #7648, opened 2026-08-12
fix: prevent global CSS from overriding component libraries

### LINKED ISSUE #3867: Global CSS makes recombination with other Component Frameworks difficult
https://github.com/db-ux-design-system/core-web/blob/8b2bd48a44f1bff58575c18683d0d6519b2728b4/packages/foundations/scss/defaults/default-required.scss#L103-L104

Just an example

We use primeng, e.g. for muliteselects and therefore we need to manually "disable" all global styles, because they spill over into these components.

michaelmkraus:
## Proposed changes

Keeps DB UX foundation and component style imports in the named `db-ux` cascade layer, including optimized/direct import examples and the import-style fixture. This prevents global normalization rules from outranking third-party component-library styles.

Documents explicit layer ordering for integrations such as PrimeNG and routes JavaScript consumers through a layered CSS entry file.

resolves #3867

### Code Quality

### Validation

Validated with:

- `pnpm run build`
- `pnpm run test`
- `pnpm run lint`
- `pnpm run build-outputs`

### General

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/fix-global-css-layer-3867>


packages/components/src/styles/layered.scss
```diff
@@ -0,0 +1,3 @@
+/* Opt-in entry point that keeps all DB UX styles in a low-priority layer. */
+@layer db-ux;
+@import "./bundle.css" layer(db-ux);
```
nmerget, line 2:
```suggestion
@forward "./bundle.css" layer(db-ux);
```

packages/components/src/styles/layered.scss
nmerget, line 1: Why do we need this file?

In our documentation we already use this example:


```css
@layer db-theme, db-ux;
@import "@db-ux/db-theme/build/styles/rollup.css" layer(db-theme);
@import "@db-ux/core-components/build/styles/bundle.css" layer(db-ux);
```

So we just "hide" the `layer(db-ux)` inside the `layered.css` file. So in fact we need to add another comment again to say: "If you want to order the the layers do this and that."

In my opinion the issue `https://github.com/db-ux-design-system/core-web/issues/3867` is already solved. We use layers for the design system, if there is any conflict with other Libraries, the should handle it on their side. If they use a technology which doesn't support layers, they should use another one

packages/components/src/styles/layered.scss
michaelmkraus: `layer()` is a CSS `@import` modifier and is not supported by Sass `@forward`. Sass rejects this with `Error: expected ";"`. The native CSS import is intentional because it assigns the complete generated bundle to `db-ux`.

packages/components/src/styles/layered.scss
michaelmkraus: Yes, consumers can already add `layer(db-ux)` manually. The separate entry primarily establishes an opt-in package contract for JS/TS imports, where consumers cannot attach a layer to an CSS import. Existing unlayered entries remain unchanged for compatibility. Explicit ordering is still unavoidable when another library also creates a named layer. For unlayered third-party styles no additional ordering is needed.

nmerget:
We'll close this, by answering https://github.com/db-ux-design-system/core-web/issues/3867 that layers are the way to go to avoid a conflict with other libraries, right?

michaelmkraus:
We have decided not to proceed with this PR. Cascade layers are the appropriate mechanism for controlling precedence between the DB UX Design System and other component libraries. However, the layer configuration and ordering belong to the consuming application, as only the application knows which libraries are used and which styles should take precedence.

Consumers can already place the DB UX styles in a named layer:

```css
@layer db-theme, db-ux;

@import "@db-ux/db-theme/build/styles/rollup.css" layer(db-theme);
@import "@db-ux/core-components/build/styles/bundle.css" layer(db-ux);
```

When another component library also uses a named layer, its layer can be included in the application's explicit layer order. Unlayered third-party styles already take precedence over layered author styles.

The additional layered.css entry introduced by this PR would only hide the layer(db-ux) import modifier. It could not define the correct ordering relative to other libraries, so consumers would still need to configure that ordering themselves.

Issue #3867 can therefore be addressed with the existing cascade-layer integration rather than an additional package entry point. I am closing this PR without merging it.

PR #7659, opened 2026-08-13
fix(header): react to `forceMobile` property changes after initial render

### LINKED ISSUE #6638: Header doesn't change when disabling forceMobile

- [ ] All
- [ ] HTML
- [ ] React
- [x] Angular
- [ ] Vue
- [ ] Web components
- [ ] Power Apps

### Reproduction case

Go to https://design-system.deutschebahn.com/core-web/review/main/angular-storybook/?path=/story/components-dbheader-examples--without-application-name, then set `forceMobile` (not `forcedToMobile`!) to `true`, then to `false` again

### Expected Behaviour

Mobile mode is disabled again

### Screenshots

_No response_

### Browser version

Edge

_No response_

DB Fernverkehr

_No response_

mfranzke:
## Proposed changes

`DBHeader` only applied `forceMobile` once, right after mount. Setting it to `true` later did nothing, and switching it back to `false` never released the mobile layout.

Two causes in `header.lite.tsx`:

- the `onUpdate` dependency list was `[state.initialized, _ref]`, so it never re-ran when `forceMobile` changed
- it hardcoded `data-force-mobile="true"` and never wrote the `"false"` counterpart

The effect now depends on `props.forceMobile`, derives the value through `getBoolean` (so the bare `force-mobile` attribute in web components counts as `true`) and writes `String(forceMobile)`. Headers that were never forced stay untouched, so we don't stamp `data-force-mobile="false"` onto every child of every header. `data-on-forcing-mobile` uses `getBoolean` as well, so the anti-flicker rule also applies to string prop values.

No SCSS change needed: `_screen-sizes.scss` already matches both `:not([data-force-mobile])` and `[data-force-mobile="false"]`.

Verified in the React storybook by toggling the `forceMobile` control on/off/on and checking `data-force-mobile` on the header children; against the previous commit the first toggle already fails. Generated React, Angular, Vue and Stencil output confirmed to watch the prop.

resolves #6638

### Code Quality

### Validation

- [ ] ~I have added/updated tests that cover my changes~ with the release of the Shell, the header is a legacy component

### General

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/6638-header-doesnt-change-when-disabling-forcemobile>


mfranzke:
@codex review

.changeset/lucky-headers-toggle.md
```diff
@@ -0,0 +1,9 @@
+---
+"@db-ux/core-components": patch
```
mfranzke:
```suggestion
```

nmerget (approved):

PR #7686, opened 2026-08-14
fix: handle safe-area-inset correctly for drawer direction up/down

### LINKED ISSUE #7685: Drawer: safe-area-inset CSS values not applied correctly per direction
## Which generators are affected?

- [x] All

## Reproduction case

On devices with `viewport-fit=cover` (PWA / mobile Safari), the `DBDrawer` component uses `env(safe-area-inset-top)` and `env(safe-area-inset-bottom)` directly in its SCSS. This causes two issues:

1. **Direction-specific safe area handling is missing:** When the drawer opens with `direction="up"`, the top safe area inset should not add padding to the header (the drawer does not touch the top of the viewport). Instead, a margin should push the drawer down. Similarly, `direction="down"` should not apply bottom padding but use a margin instead. Currently, both directions apply the same padding, leading to incorrect spacing.

2. **Not debuggable on desktop:** Since `env(safe-area-inset-*)` always resolves to `0px` on desktop browsers (no notch/home indicator), there is no way to simulate or visually verify safe area behavior without a real device or iOS Simulator.

**Steps to reproduce:**
1. Open a `DBDrawer` with `direction="up"` on a device with a notch (e.g. iPhone with `viewport-fit=cover`)
2. Observe that the header still receives top safe area padding even though the drawer does not reach the top edge of the screen
3. On desktop, there is no way to override `env()` to test the layout

## Expected Behaviour

- `direction="up"`: No top padding in the header; instead, a `margin-block-start` pushes the drawer below the notch/status bar. Bottom safe area padding is applied to footer/content.
- `direction="down"`: No bottom padding in footer/content; instead, a `margin-block-end` keeps the drawer above the home indicator. Top safe area padding is applied to the header.
- `direction="to-left"` / `direction="to-right"` (default): Top and bottom safe area padding applied as before.
- The safe area values should be exposed as CSS custom properties (e.g. `--db-drawer-safe-area-inset-top`, `--db-drawer-safe-area-inset-bottom`) so they can be overridden in DevTools for desktop debugging.

## Screenshots

_N/A – requires real device or simulated safe area insets to observe._

## Browser version

- Safari (iOS PWA primarily, but affects all browsers on notch devices)

## Additional context

The drawer uses `env(safe-area-inset-top)` and `env(safe-area-inset-bottom)` directly in `drawer.scss` and `drawer-header.scss`. These cannot be overridden via CSS custom properties, making desktop testing impossible and preventing direction-aware behavior.

d-koppenhagen:
## Proposed changes

Fixes `env(safe-area-inset-*)` handling in `DBDrawer` for `direction="up"` and `direction="down"`. Previously, the safe-area padding was applied inside the container (header/footer) regardless of direction, causing unwanted inner spacing in PWAs where `viewport-fit=cover` is set.

Now:
- **`direction="up"`**: No top padding in header; margin-block-start on container keeps content below notch/status bar. Bottom padding in footer/content remains for home indicator.
- **`direction="down"`**: No bottom padding in footer/content; margin-block-end on container keeps content above home indicator. Top padding in header remains for notch.
- **`to-left` / `to-right`** (default): Unchanged behavior — top + bottom padding inside.

Uses CSS custom properties (`--db-drawer-safe-area-inset-top`, `--db-drawer-safe-area-inset-bottom`, `--db-drawer-safe-area-margin-block-start`, `--db-drawer-safe-area-margin-block-end`) so the safe-area behavior can be simulated on desktop for debugging without a real device or iOS Simulator.

Adds Playwright screenshot tests for all 4 directions with simulated safe-area insets.

resolves #7685

### Code Quality

### Validation

### General

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/fix-7685-drawer-safe-area-insets>


mfranzke:
@codex review

showcases/e2e/drawer/drawer-safe-area-visual-snapshot.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+		top: 0;
+		height: 47px;
+	}
+
+	.db-drawer[open]::after {
+		content: "Safe Area Bottom (34px)";
+		bottom: 0;
+		height: 34px;
+	}
+`;
```
mfranzke, line 53: I'm questioning whether it might be beneficial to have this CSS extracted to an external CSS file (mainly for clarity and separation, and code highlighting).

packages/components/src/components/drawer/drawer.scss
```diff
@@ -267,10 +295,14 @@
 
 		&[data-direction="up"] {
 			@extend %direction-from-bottom-to-top;
+
+			--db-drawer-safe-area-inset-top: 0px;
```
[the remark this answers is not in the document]
d-koppenhagen: The scenario doesn't apply. A `direction="up"` drawer opens from the bottom.
Its header sits at the content edge, not the viewport edge.
In the default partial-height case the header never reaches the notch area, so top inset padding would just be wasted space.

For the full-height edge case (`containerSize="full"` + `direction="up"`): I've added dedicated Up (Full) and Down (Full) test variants to the direction example and extended the safe-area E2E spec to cover them.
The screenshots confirm that the current behavior is correct — even at full height, the drawer container already covers the viewport fully, so the overlay zones are visible but the content stays within bounds.

If this combination ever needs special handling, it should be scoped to `containerSize="full"` specifically — not a blanket `env(safe-area-inset-top)` for all bottom drawers.

showcases/e2e/drawer/drawer-safe-area-visual-snapshot.spec.ts
d-koppenhagen: I intentionally kept the styles inline. They're a test-only fixture (simulating safe-area insets + rendering visual overlays for screenshot comparison) — not application code. Having them co-located with the test gives immediate context when reading the spec, without jumping to a separate file. At ~45 lines it's still manageable, and `page.addStyleTag({ content })` is the idiomatic Playwright approach for this kind of injection.

packages/components/src/components/drawer/drawer.scss
mfranzke, line 299: by setting these to a fixed value (`0px`) on `[data-direction="up"]` the possibility for others to overwrite it (for whatever reason) decreases (not a disaster, but kind of a nitpick). What about a simplification and setting that variable depending on `:not([data-direction="up"])` on line 139?

Lines 305 and 140 accordingly.

d-koppenhagen: I am not sure about this. 

The current pattern is intentional: set a default (`env(…)`), then override per direction. This "defaults first, specialize later" approach reads top-down and scales better IMO. Adding a new direction only requires one positive override instead of updating a `:not()` list.

The override-ability is a feature here, not a side effect I would say. The whole reason we use custom properties instead of `env()` directly is to allow debugging overrides (as documented in the comment block above). But a consumer can still set `--db-drawer-safe-area-inset-top` on `[data-direction="up"]` if they need it which is a bennefit.

mfranzke (dismissed):

mfranzke (approved):

PR #7713, opened 2026-08-17
refactor: resolve changesets changelog module by path instead of package subpath

mfranzke:
## Proposed changes

Per the [Customize Changelog Format guide](https://changesets.dev/guide/customize-changelog-format#using-a-custom-changelog-generator), a custom generator is either a **path relative to the config file** (`"./changelog-generator.ts"`) or an **installed module** — see also the [`changelog` config option](https://changesets.dev/guide/config#changelog), which describes it as an npm package or a relative file path.

Our bare `scripts/github/changesets/create-changelog.js` is neither: it is not relative, so it was resolved as a *module*. Changesets resolves `changelog[0]` from `<root>/.changeset` first and falls back to its own `dist` directory (see [`getNewChangelogEntry`](https://github.com/changesets/changesets/blob/main/packages/apply-release-plan/src/index.ts)), and only that fallback made it work — as a package subpath import of the `scripts` workspace package via pnpm's virtual store (`node_modules/.pnpm/node_modules/scripts`).

That means changelog generation silently depended on:

- the workspace package keeping the name `scripts`,
- that package not gaining an `exports` field (which would forbid the subpath),
- pnpm's store layout hoisting workspace packages.

Any of those changing would break `changeset version` at release time with `Could not resolve changelog generation functions`, and the behaviour differs under npm or yarn.

This switches to the relative form documented by the config schema (`"./some-module"`), so the module resolves on the first attempt straight to the file, with no `node_modules` involvement:

```jsonc
"changelog": [
	"../scripts/github/changesets/create-changelog.js",
	{ "owner": "db-ux-design-system", "repo": "core-web" }
]
```

No behaviour change for the generated changelogs — same generator, same output.

### Validation

- From `.changeset/`, the bare specifier fails with `ERR_MODULE_NOT_FOUND` while `../scripts/…` resolves to the real file and exports both `getReleaseLine` and `getDependencyReleaseLine`. Confirmed for both resolvers changesets uses: CJS `resolve-from` (v2, currently on `main`) and ESM `import.meta.resolve` (v3), so this is correct before and after the pending `@changesets/cli` v3 bump.
- Ran a real `changeset version` in a throwaway git worktree: the entry was generated in the repo's usual format (`- fix(...): ... - [see commit abc1234](...)`), dependency release lines stayed suppressed, and versions bumped as expected.

No changeset entry: `.changeset/config.json` is release tooling and is not shipped to consumers.

### Code Quality

### Validation

### General

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/fix-resolve-changesets-changelog-module-by-path-instead-of-package-subpath>


mfranzke:
@codex review

.changeset/config.json
```diff
@@ -1,7 +1,7 @@
 {
 	"$schema": "https://unpkg.com/@changesets/config/schema.json",
 	"changelog": [
-		"scripts/github/changesets/create-changelog.js",
+		"../scripts/github/changesets/create-changelog.js",
```
nmerget, line 4: It is working on a release and locally, no matter if I use `../`.

See this release, the link to the commit is present:
https://github.com/db-ux-design-system/core-web/releases#:~:text=ngx%2Dcore%2Dcomponents-,5.1.5,react%20to%20forceMobile%20changes%20after%20the%20initial%20render%20%2D%20see%20commit%20d2cbdf7,-%40db%2Dux/react

mfranzke: This is not about that it wouldn't be working. It's about that the path is incorrect

mfranzke:
If it could be either a path or a module, than this is correctly interpreted as a module.

PR #7824, opened 2026-08-26
feat: add `width` and `height` properties to DBStack and remove default `block-size: 100%`

### LINKED ISSUE #7718: DBStack: Unwanted whitespace in layout due to block-size: 100%

- [ ] All
- [x] HTML
- [ ] React
- [ ] Angular
- [ ] Vue
- [ ] Web components
- [ ] Power Apps

### Reproduction case

The following HTML (including nested `.db-stack` components) causes empty space to appear in the layout.
```html
<div class="db-stack">
              <div>
                <div class="card-header">
                  <h2 data-headline-size="2xs">Heading</h2>
                </div>
                  <div class="db-divider"></div>
                  <div class="db-stack" data-direction="row">
                    <div class="proof-measure-number">
                      <span>1</span>
                    </div>
                    <div>
                      <strong>Lorem ipsum</strong>
                        <div class="html-content" data-density="functional" data-on-bg-color="emphasis-80"><p>redacted</p>

</div>
                        <div>
                              <a href="https://localhost" target="_blank" rel="noopener noreferrer nofollow" class="db-link">Link</a>
                        </div>
                    </div>
                  </div>
                    <ul class="db-accordion mbs-xs" data-density="functional">
                      <li class="db-accordion-item">
                        <details>
                          <summary>Summary</summary>
                          <ul>
                              <li>Text</li>
                          </ul>
                        </details>
                      </li>
                    </ul>
              </div>
        </div>
```

### Expected Behaviour

We identified that the `block-size: 100%` rule causes this behavior. Is the rule really needed? Replacing `db-stack` with our own flex implementation (same code, just without the block-size) allows us to nest without restrictions.

We would ideally like to use nested `db-stack` components without layout bugs. This is especially relevant for nesting horizontal stacks within vertical stacks, which is a common pattern on our app.

### Screenshots

Empty space bug:

Image

This is a screenshot of the same code with the `block-size` rule disabled in DevTools:

Image

### Browser version

None

_No response_

FIA

KIGS

nmerget:
## Proposed changes

Adds `width` and `height` properties to `DBStack`, fixes the default `block-size: 100%` that caused unwanted whitespace in nested layouts, and cleans up the component's SCSS structure to align it with other components.

**What changed:**

- **Default height is now `auto`** — removes the unconditional `block-size: 100%` that forced nested stacks to stretch to their parent's full height, creating empty space when content is shorter. Use `height="full"` to opt into `block-size: 100%` when needed.
- Added `width` property (`full` | `auto`, default: `full`) — `full` gives `inline-size: 100%`, `auto` uses `fit-content` so the stack shrinks to its children
- Added `height` property (`full` | `auto`, default: `auto`) — `full` gives `block-size: 100%` for layouts that need it (e.g. justify-content column, alignment row, column wrap)
- Inlined styles from `_stack-components.scss` directly into `stack.scss` — the `%default-stack` placeholder was only used in one place (unlike `%default-button` which is shared across button, tag, badge, input). This aligns DBStack with other components like DBDivider that keep their styles in the main SCSS file
- Removed unused `_stack-components.scss` (dead `arr-join` function, unused `custom-elements` and `sass:list` imports, single-consumer placeholder)
- Removed unused `stack-web-component.scss` (one-liner `@forward` with zero references)
- Updated examples that rely on full height (`justify-content-column`, `alignment-row`, `wrap-column`) to explicitly set `height="full"`

resolves #7718

### Code Quality

### Validation

### General

🔭🐙🐈 Test this branch here:

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/fix-stack-block-size>


mfranzke:
@nmerget the description of this PR is **VERY LOUD**

.changeset/fix-stack-block-size.md
```diff
@@ -0,0 +1,9 @@
+---
+"@db-ux/core-components": minor
+"@db-ux/ngx-core-components": minor
+"@db-ux/react-core-components": minor
+"@db-ux/wc-core-components": minor
+"@db-ux/v-core-components": minor
+---
+
+feat: add `width` property to DBStack and remove default `block-size: 100%`
```
mfranzke:
```suggestion
feat(DBStack): add `width` property and remove default `block-size: 100%`
```

__snapshots__/stack/showcase/mobile-chrome/DBStack-should-match-screenshot-1/DBStack-should-match-screenshot.png
mfranzke, line 1: some of these changes do not look like expected (especially the ones that should show different vertical spacing).

nmerget: Good catch! The issue was that removing `block-size: 100%` broke column wrap and the justify-content/alignment showcases (they need the stack to fill the parent height to demonstrate vertical spacing).

Fixed in the next push:
- `block-size: 100%` is now set only when `direction="column"` + `wrap="true"` (required for flex column wrap to work)
- Added a new `height` property (`"full"` | `"auto"`) mirroring the existing `width` property — so users can explicitly opt in to `block-size: 100%` when needed
- Updated the justify-content-column examples to use `height="full"` so they properly demonstrate vertical spacing again

The screenshots should look correct after the new snapshots are generated.

packages/components/src/styles/internal/_stack-components.scss
```diff
@@ -73,9 +73,18 @@ $stack-gaps: (
...
 	justify-content: flex-start;
-	block-size: 100%;
-	inline-size: 100%;
-	overflow: auto;
+
+	&:is([data-width="full"], :not([data-width])) {
+		inline-size: 100%;
+	}
+
+	&[data-height="full"] {
```
michaelmkraus: The final code does not match the previous review reply: `column + wrap` only gets `block-size: 100%` when `height="full"` is also set. The existing “Wrap: Column” example does not set it, so the stack grows with its children instead of wrapping. Could we restore the `column + wrap` selector or document and test `height="full"` as a required opt-in?

packages/components/src/styles/internal/_stack-components.scss
```diff
@@ -73,9 +73,18 @@ $stack-gaps: (
...
+		block-size: 100%;
+	}
+
+	&:not([data-wrap="true"]) {
```
michaelmkraus: Is this overflow change in scope for the `block-size` fix? With `wrap=true`, an oversized unbreakable child now spills outside the stack instead of remaining scrollable, while non-wrapping stacks still retain the focus-outline clipping risk of `overflow: auto`. Could we revert this here or document and test the intended overflow contract separately?

packages/components/src/shared/model.ts
```diff
@@ -231,6 +231,15 @@ export type WidthProps = {
 	width?: WidthType | string;
 };
 
+export const HeightList = ['full', 'auto'] as const;
+export type HeightType = (typeof HeightList)[number];
+export type HeightProps = {
+	/**
+	 * Height of the component. Auto height based on children size, full height based on parent elements height.
+	 */
+	height?: HeightType | string;
```
michaelmkraus: `HeightType | string` effectively becomes `string`, although the implementation only supports `full` and `auto`. Values such as `height="ful"` or `"20rem"` compile but silently behave like auto. Since only `DBStack` currently uses this prop, could we keep it as an exact Stack-local union until shared or arbitrary-value semantics are actually needed?

packages/components/src/styles/internal/_stack-components.scss
nmerget: Good point! Moved the `block-size: 100%` rule inside the `&[data-wrap="true"]` selector where it logically belongs — column wrap needs a constrained height to actually wrap. Now reads as:

```scss
&[data-wrap="true"] {
    flex-wrap: wrap;

    // Column wrap requires explicit block-size to function correctly
    &:not([data-direction="row"]) {
        block-size: 100%;
    }
}
```

This keeps it co-located with the wrap behavior and makes the dependency obvious. The "Wrap: Column" example works without needing `height="full"` explicitly.

packages/components/src/styles/internal/_stack-components.scss
nmerget: The overflow change is intentional and in scope: when `wrap=true`, content wraps by design — `overflow: auto` would fight against that by adding scrollbars instead of allowing wrapping. Non-wrapping stacks still get `overflow: auto` to handle overflowing content gracefully.

Regarding focus-outline clipping on non-wrapping stacks — that's a valid concern but a pre-existing issue (the old code had unconditional `overflow: auto`). Happy to address it in a follow-up PR focused specifically on the overflow contract.

packages/components/src/shared/model.ts
nmerget: Agreed — removed `| string` so `height` is now a strict `"full" | "auto"` union. Typos like `"ful"` will be caught at compile time.

mfranzke:
@codex review

.changeset/fix-stack-block-size.md
```diff
@@ -0,0 +1,9 @@
+---
+"@db-ux/core-components": minor
```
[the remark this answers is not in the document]
nmerget: Fixed — `block-size: 100%` is now the default again (applies when no `data-height` is set), preserving the original behavior. The selector changed from `&[data-height="full"]` to `&:is([data-height="full"], :not([data-height]))`. Consumers only see a difference when they explicitly opt in to `height="auto"`, so this stays `minor` — no breaking change.

packages/components/src/styles/internal/_stack-components.scss
```diff
@@ -73,9 +73,18 @@ $stack-gaps: (
 
 	gap: variables.$db-spacing-fixed-sm;
 	justify-content: flex-start;
-	block-size: 100%;
-	inline-size: 100%;
-	overflow: auto;
+
+	&:is([data-width="full"], :not([data-width])) {
+		inline-size: 100%;
+	}
```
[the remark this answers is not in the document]
nmerget: Good catch — added an explicit `&[data-width="auto"]` rule with `inline-size: fit-content` so the stack actually shrinks to its children's size. A block-level flex container fills available width by default, so without this rule `auto` was indistinguishable from `full`.

packages/components/src/styles/internal/_stack-components.scss
```diff
@@ -92,6 +101,11 @@ $stack-gaps: (
 
 	&[data-wrap="true"] {
 		flex-wrap: wrap;
+
+		// Column wrap requires explicit block-size to function correctly
+		&:not([data-direction="row"]) {
+			block-size: 100%;
```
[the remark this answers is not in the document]
nmerget: Fixed — the column-wrap selector now excludes `height="auto"`: `&:not([data-direction="row"]):not([data-height="auto"])`. When a consumer explicitly sets `height="auto"`, the column-wrap override no longer forces `block-size: 100%`, so the stack grows with its content as expected.

packages/components/src/components/stack/stack.scss
```diff
@@ -1,5 +1,103 @@
...
+		inline-size: 100%;
+	}
+
+	&[data-width="auto"] {
+		inline-size: fit-content;
+	}
+
+	&[data-height="full"] {
+		block-size: 100%;
+	}
```
michaelmkraus, line 70: This changes the default height and breaks existing layouts unless consumers add `height="full"`. Should we preserve the old default or release this as a major change?

packages/components/src/components/stack/stack.scss
```diff
@@ -1,5 +1,103 @@
...
+	justify-content: flex-start;
+	overflow: auto;
+
+	&:is([data-width="full"], :not([data-width])) {
+		inline-size: 100%;
+	}
+
+	&[data-width="auto"] {
+		inline-size: fit-content;
+	}
```
michaelmkraus, line 66: `height="auto"` can stretch to the parent’s height in flex/grid layouts. We could add content sizing, such as `block-size: fit-content`, to match the documented behavior.

mfranzke:
@codex review

mfranzke:
@codex review

PR #7825, opened 2026-08-26
fix: radio checked state barely visible at `size=small` with `functional` density

### LINKED ISSUE #6317: Radio (size="small", density="functional") checked state barely visible compared to Figma

- [ ] All
- [ ] HTML
- [x] React
- [ ] Angular
- [ ] Vue
- [x] Web components
- [ ] Power Apps

### Reproduction case

Use a `Radio` component with the following configuration:
- size="small"
- data-density="functional"

When the radio button is toggled between **checked** and **unchecked**, the visual difference between the two states is barely noticeable in the implementation.

In the **Figma design**, the inner fill / border proportions are larger, making the checked state clearly visible even at the small size (16px).

Screenshots comparing **Figma** and the **implemented component** are attached.

### Expected Behaviour

The visual appearance of the radio button should match the **Figma design**.

Specifically:

- The **checked indicator (inner fill)** should be more prominent at `size="small"` so the difference between **checked** and **unchecked** is clearly visible.
- The proportions of the inner fill and border should align with the **Figma component specification**.

### Screenshots

Figma
Image

Image
^^ This is checked!

### Browser version

Edge

_No response_

InfraGO

CTMS

nmerget:
## Proposed changes

The `data-size="small"` selector for the radio checked border-width was incorrectly placed on the `input` element (`input:checked[data-size="small"]`), but `data-size` is actually set on the parent `<label>` wrapper. This meant the small-specific border-width calculation never applied, falling back to the medium formula — resulting in a barely visible checked indicator at small size + functional density (3px border on a 16px circle).

**Fix:** Moved the selector to `.db-radio[data-size="small"] input:checked` so the intended thicker border-width (`(icon-font-size - sizing-3xs) / 2`) now correctly applies for small radios.

**Added:** An `examples.example.lite.tsx` showcase covering all density × size × checked combinations for visual regression testing.

resolves #6317

### Code Quality

### Validation

### General

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/fix-radio-small-checked-state>


.changeset/fix-radio-small-checked.md
```diff
@@ -0,0 +1,9 @@
+---
+"@db-ux/core-components": patch
+"@db-ux/ngx-core-components": patch
+"@db-ux/react-core-components": patch
+"@db-ux/wc-core-components": patch
+"@db-ux/v-core-components": patch
+---
+
+fix: radio checked state barely visible at size="small" with density="functional"
```
mfranzke:
```suggestion
fix(radio): checked state barely visible at `size="small"` with `density="functional"`
```

nmerget: Can you add this to the AGENTS.md file if you want to have this rule?

.changeset/fix-radio-small-checked.md
michaelmkraus, line 1: The changeset should use `minor` for all five packages. `AGENTS.md` explicitly requires a minor bump for new or modified files under `src/components/*/examples/`. Otherwise, we should adapt the `AGENTS.md`.

packages/components/src/components/radio/examples/examples.example.lite.tsx
```diff
@@ -0,0 +1,53 @@
...
+	storybookArgTypes: StorybookRadioArgTypes
+});
+
+export default function RadioExamples() {
+	return (
+		<div role="radiogroup" aria-label="Examples">
+			<DBRadio
+				data-density="functional"
+				size="small"
+				name="Examples1"
```
michaelmkraus, line 24: This container declares one `radiogroup`, but the six radios use different `name` values and are all checked. Native grouping and keyboard behavior therefore conflict with the ARIA semantics. Please remove `role="radiogroup"` for this visual matrix or split it into correctly grouped examples.

.changeset/fix-radio-small-checked.md
mfranzke: > The changeset should use `minor` for all five packages. `AGENTS.md` explicitly requires a minor bump for new or modified files under `src/components/*/examples/`. Otherwise, we should adapt the `AGENTS.md`.

oh no, not adapting it again … :-( (well, actually, you're right …)

nmerget: > - **New or changed examples** (`src/components/*/examples/`) are a `minor` bump.

This is wrong, a new example isn't a `minor`

nmerget: Updated `packages/components/AGENTS.md` to clarify that examples don't determine the bump type — the underlying change does. A bug fix with a new regression example stays `patch`; an example for a newly added prop is `minor` because the prop itself is the feature.

packages/components/src/components/radio/examples/examples.example.lite.tsx
nmerget: Fixed — removed `role="radiogroup"` and `aria-label` from the wrapper `<div>`. Since each radio uses a different `name` and they're all individually checked (visual matrix), they're not a semantic radio group.

mfranzke:
@codex review

mfranzke:
> Codex Review: Didn't find any major issues. Can't wait for the next one!

@nmerget, what do you think of that ?!?

nmerget:
> > Codex Review: Didn't find any major issues. Can't wait for the next one!
> 
> @nmerget, what do you think of that ?!?

Sometimes a miracle happens

__snapshots__/radio/showcase/chromium/DBRadio-should-match-screenshot-1/DBRadio-should-match-screenshot.png
mfranzke, line 1: could we position the "Examples" to the very bottom? Than our regular property list wouldn't be interrupted by those examples.

mfranzke (dismissed):

mfranzke (approved):

PR #7830, opened 2026-08-26
fix(table): remove unwanted white background on component wrapper for `variant=spaced`

### LINKED ISSUE #7620: Table `variant="spaced"` has unwanted white background on wrapper

- [ ] All
- [ ] HTML
- [x] React
- [ ] Angular
- [ ] Vue
- [ ] Web components
- [ ] Power Apps

### Reproduction case

When using the DB UX Table with `variant="spaced"`, the table wrapper has a white background in addition to the individual table rows.

The wrapper background fills the gaps between the spaced rows and also creates an additional white area below the last row.

This appears to be caused by both `.db-table` and `.db-table-row` having the background color `var(--db-adaptive-bg-basic-level-1-default)`:

```css
.db-table {
  background-color: var(--db-adaptive-bg-basic-level-1-default);
}

.db-table-row {
  background-color: var(--db-adaptive-bg-basic-level-1-default);
}
```

For `variant="spaced"`, the wrapper background therefore remains visible between the individual row cards.

The issue can currently be worked around with:

```csss
.db-table[data-variant='spaced'] {
  background-color: transparent;
}
```

### Expected Behaviour

When using `variant="spaced"`, the individual table rows should retain their background, while the space between the rows should show the background of the surrounding container.

The table wrapper itself should therefore have a transparent background for the `spaced` variant.

Ideally, this should be the default behaviour of the `spaced` variant and should not require an additional CSS override in consuming applications.

The behaviour of the `flat` and `zebra` variants should remain unchanged.

### Screenshots

_No response_

### Browser version

Edge

Not browser-specific. The issue is caused by the component styling and can be reproduced consistently.

The issue seems to originate from the background color being applied to the table wrapper itself.

A possible solution would be to make the wrapper background transparent specifically for `variant="spaced"`:

```css
.db-table[data-variant='spaced'] {
  background-color: transparent;
}
```

DB InfraGO AG

_No response_

InfraNatur (Data4Quality)

nmerget:
## Proposed changes

Removes the unwanted white background on the `.db-table` wrapper when using `variant="spaced"`.

**What changed:**
- Moved `background-color` from the `.db-table` wrapper div to the `table` element, scoped to flat/zebra/default variants only. This prevents the padding area around the table from being colored.
- For the spaced variant, the thead rows without `data-sub-header-emphasis` now get `background-color: transparent` and `border-color: transparent`, so the row background doesn't bleed through the gaps between spaced cards.
- The `data-mobile-variant="list"` also resets the table's `background-color` to transparent when switching to the spaced layout at small viewports.

resolves #7620

### Code Quality

### Validation

### General

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/fix-table-spaced-background>


.changeset/fix-table-spaced-background.md
```diff
@@ -0,0 +1,9 @@
+---
+"@db-ux/core-components": patch
+"@db-ux/ngx-core-components": patch
+"@db-ux/react-core-components": patch
+"@db-ux/wc-core-components": patch
+"@db-ux/v-core-components": patch
+---
+
+fix: remove unwanted white background on table wrapper for `variant="spaced"`
```
mfranzke:
```suggestion
fix(table): remove unwanted white background on component wrapper for `variant="spaced"`
```

packages/components/src/components/table/table.scss
```diff
@@ -147,6 +150,7 @@
 	&[data-variant="flat"],
 	&[data-variant="zebra"] {
 		table {
+			background-color: colors.$db-adaptive-bg-basic-level-1-default;
```
michaelmkraus, line 153: Moving the background to `<table>` makes the wrapper padding, scrollbar area, and `width="auto"` margins transparent for non-spaced variants. Could we keep the wrapper background and only make `spaced` and mobile `list` transparent?

packages/components/src/components/table/table.scss
michaelmkraus, line 231: This `thead` background still fills the gaps between header rows for `spaced` and mobile `list` tables with a sticky header. Could the sticky background be applied per row instead?

packages/components/src/components/table/table.scss
nmerget: image

It looks like this at the moment. So the tiny padding for the focus creates a backround next to the table border, which looks pretty wrong. By moving it to the table we have the correct background only on the table itself. 

We add the transparent background for the `spaced` and `list` already

packages/components/src/components/table/table.scss
nmerget: The background on `thead` is intentional for the sticky horizontal header — without it, the underlying body rows would show through the header when scrolling vertically. We've now added `border-radius` to the sticky `thead` for the spaced variant so it visually matches the rounded row cards. I also added a comment in the SCSS explaining why the background is necessary here.

For the `list` mobile variant this shouldn't be an issue since the thead gets `border-block-start-color: transparent` and the rows use flex layout there anyway.

michaelmkraus (commented):
🤖 **AI Review** — Two new findings, prioritized below. I did not duplicate the existing discussions about non-spaced wrapper gutters or the sticky `thead` background.

packages/components/src/components/table/table.scss
```diff
@@ -247,8 +270,8 @@
 	@extend %sticky-header;
 
 	container-type: inline-size;
-	background-color: colors.$db-adaptive-bg-basic-level-1-default;
 	inline-size: 100%;
+	max-block-size: 100%;
```
michaelmkraus, line 274: This global height constraint also applies to `width="auto"`, but that branch has `overflow: visible`. In a fixed-height parent, the flexed inner table shrinks to the wrapper while its rows keep painting beyond it and over following siblings. A browser reproduction with a 100px parent and ten 30px rows produced a 102px wrapper with `scrollHeight: 531px`; the final row overlapped the next sibling. Without this declaration, the wrapper remains 532px tall and the sibling follows the content. Could this constraint be limited to the full/default branches that already use `overflow: auto`, or could the auto-width behavior be handled explicitly? It is unrelated to the spaced-background fix and currently introduces a consumer-visible layout regression.

packages/components/src/components/table/table.scss
```diff
@@ -132,7 +132,10 @@
 
 		tr {
 			&:not([data-sub-header-emphasis]) {
-				@extend %transparent-border;
+				@extend %default-card;
+
+				background-color: transparent;
```
michaelmkraus, line 137: This overrides `.db-table-row`'s level-1 background, so an ordinary spaced header row becomes fully transparent while body rows remain level-1 cards. With the wrapper and table now transparent, the whole header surface—not only the row gaps—shows the consumer's surrounding background. That conflicts with #7620's expectation that individual rows retain their background and can also leave default header text on an arbitrary host surface. The compiled CSS reproduces `header: transparent`, `body: rgb(255,255,255)`. Could we keep the header row background and only make the inter-row area transparent (for example, retain `%default-card`/transparent border but remove this background override)?

nmerget: This is on purpose. This is how it looks:

Image

 The thead first row shouldn't have a background color. Otherwise we would get something like this:

Image

It this case the first row looks very similar to the tbody rows (which are cards) but missing the border. This looks very off. The individual rows keep their background, so this is what the user asked

packages/components/src/components/table/table.scss
nmerget: Good catch, thanks. The problem was that `max-block-size: 100%` was paired with `overflow: visible` on the `data-width="auto"` branch, so the box got capped but the rows still painted outside it and overlapped siblings.

Fixed in bb503234e802 by splitting the axes instead of dropping the height cap:

- `.db-table` now always sets `overflow-y: auto`, so a tall table (any width mode) respects its parent's height and scrolls vertically instead of overflowing.
- `overflow-x: auto` is applied only to the `:not([data-width])` / `[data-width="full"]` branch, where a table can be wider than its wrapper.
- The `[data-width="auto"]` branch intentionally sets no `overflow-x`: its table is `fit-content` and centered, so it never exceeds the wrapper width. (Setting `overflow-x: visible` there would be promoted to `auto` by the browser anyway, since `overflow-y` is `auto`.)

This keeps the parent-height behavior the sticky header needs while removing the visible-overflow regression for auto-width tables.

mfranzke:
@codex review

packages/components/src/components/table/table.scss
```diff
@@ -247,14 +270,17 @@
 	@extend %sticky-header;
 
 	container-type: inline-size;
-	background-color: colors.$db-adaptive-bg-basic-level-1-default;
 	inline-size: 100%;
+	max-block-size: 100%;
 	// stylelint-disable-next-line db-ux/use-spacings
 	padding: variables.$db-border-width-xs; // offset for focus
+	// Always allow vertical scrolling so a tall table respects its parent's height
+	overflow-y: auto;
```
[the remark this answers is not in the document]
nmerget: Good catch. Fixed in c83ed332be40 by switching the `[data-width="auto"]` branch from `justify-content: center` to `justify-content: safe center`.

Since `.db-table` sets `overflow-y: auto`, an oversized auto-width table promotes the computed `overflow-x` to `auto`, so the wrapper becomes a horizontal scroll container. With plain `center` the overflow was split to both sides, pushing the leading columns before the scroll origin where they were unreachable. `safe center` keeps the table centered while it fits and falls back to `start` alignment once it overflows, so the leading columns stay pinned to the scroll origin and remain reachable. `safe` alignment is covered by our browserslist targets.

packages/components/src/components/table/table.scss
```diff
@@ -263,7 +289,15 @@
 
 	&[data-width="auto"] {
```
michaelmkraus, line 290: Since `.db-table` now sets `overflow-y: auto`, this branch becomes a
scroll container too. Popovers and tooltips are `position: absolute`, so overlays inside cells get clipped here from now on. Default and full-width tables were already scroll containers — `auto` is the only new case. Was that intended?

nmerget: image

Our popovers and tooltip are absolute in the first place but the run JS to set themselfs to fixed and recalculate the position. This is not a problem for those

michaelkraus (approved):

michaelmkraus (approved):

__snapshots__/table/showcase/chromium-highContrast/DBTable-should-match-screenshot-1/DBTable-should-match-screenshot.png
mfranzke, line 1: @nmerget this example now looks somehow destroyed, reaching over the card it is included in. Even also possible to get reproduced on the following pages:
- https://design-system.deutschebahn.com/core-web/version/v5.4.0/components/data-display/table/overview
- https://design-system.deutschebahn.com/core-web/review/main/components/data-display/table/overview

Does someone (other than me) have a look at these screenshots ?

mfranzke: https://github.com/db-ux-design-system/core-web/issues/8152

PR #7833, opened 2026-08-26
refactor(DBSelect): hide the empty option with the native hidden attribute

### LINKED ISSUE #7554: DBSelect: controlled value not updating when `required={true}` (React 19)
## Description

`DBSelect` does not reflect `value` prop changes when `required={true}`. The `onChange` fires correctly and state updates, but the rendered `<select>` element visually stays on the previous value.

Setting `required={false}` fixes the issue — identical code works as expected.

## Reproduction

```bash
npm create vite@latest dbselect-bug -- --template react-ts
cd dbselect-bug
npm install react@19.1.6 react-dom@19.1.6 @db-ux/react-core-components@4.7.2 @db-ux/core-components
```

Replace `src/App.tsx` with:

```tsx
import { useState, useCallback } from "react";
import { DBSelect } from "@db-ux/react-core-components";
import "@db-ux/core-components/build/styles/rollup.css";

const OPTIONS = ["IT-Projekte", "Anwendungsbetrieb", "Services", "Consulting"];

export default function App() {
  const [value, setValue] = useState<string>("");

  const handleChange = useCallback((e: React.ChangeEvent<HTMLSelectElement>) => {
    console.log("onChange:", e.target.value);
    setValue(e.target.value);
  }, []);

  return (
    <div style={{ padding: "2rem", maxWidth: "500px", fontFamily: "sans-serif",
                  background: "white", color: "black" }}>
      <h2>DBSelect Bug: required breaks controlled updates</h2>
      <p>Current state: <strong>{value || "(empty)"}</strong></p>

      <h3>required=true (BROKEN)</h3>
      <DBSelect label="Test" value={value} onChange={handleChange} required={true}>
        <option value="" disabled>Bitte auswählen...</option>
        {OPTIONS.map((opt) => <option key={opt} value={opt}>{opt}</option>)}
      </DBSelect>

      <h3>required=false (WORKS)</h3>
      <DBSelect label="Test" value={value} onChange={handleChange} required={false}>
        <option value="" disabled>Bitte auswählen...</option>
        {OPTIONS.map((opt) => <option key={opt} value={opt}>{opt}</option>)}
      </DBSelect>
    </div>
  );
}
```

In `src/main.tsx`, remove the `import './index.css'` line. Then `npm run dev`.

## Expected

Both selects visually update when a value is selected.

## Actual

The `required={true}` select fires `onChange` correctly (visible in console and state text), but the displayed value does not change. The `required={false}` select works correctly.

## Environment

- `@db-ux/react-core-components`: 4.7.2
- `@db-ux/core-components`: latest
- React: 19.1.6
- Vite: 8.x

michaelmkraus:
## Proposed changes

Hides the empty option of a `placeholder` or floating label `DBSelect` through the native `hidden` attribute instead of a conditional CSS rule.

Previously the option was rendered with `data-show-empty-option="false"` and hidden by a stylesheet rule combining `:has(option:checked:not(.placeholder))` and `:open`. The state now lives in the markup, which removes a piece of style-driven state handling and follows the repo's HTML → CSS → JS preference. `hidden` is added to the Angular `nativeAttributes` so the binding sets the property rather than the attribute.

The aria snapshots change accordingly: a hidden option is removed from the accessibility tree, so the affected examples no longer expose the selected empty option.

> **Note:** this PR no longer claims to fix #7554. Investigation showed that the reported defect is unrelated to the empty option — the reporter's example passes neither `placeholder` nor `variant="floating"`, so the internal empty option is never rendered in it. The actual cause is that `DBSelect` validates synchronously in `handleInput`, which re-renders while the controlled `value` is still stale, and React re-applies that value to the `select` on every commit, discarding the user's selection before `change` is dispatched. That fix is in #7860.

### Review points addressed

- `nativeAttributes` instead of a global replacement in the Angular plugin (thanks @nmerget).
- The `test-fixtures` harness is gone from this branch. It was only needed by the controlled regression test, so it moved to #7860 together with that test. This branch is now covered by three markup tests that need no harness and run for React and Vue.
- The changeset is corrected from `fix` to `refactor` and mentions the consumer-visible part: anyone who hand-wrote `data-show-empty-option="false"` and relied on our stylesheet hiding it needs to set `hidden` on that option instead.

### Code Quality

### Validation

### General

### Local validation

Select component tests (Chromium): 7 passed, `should match screenshot` fails with 254 pixels — that diff reproduces on an unmodified tree as well, so it is pre-existing locally. The cross-framework behaviour of `hidden` is covered by the showcase aria snapshots, which are shared across Angular, React, Vue and Stencil.

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/7554-fix-dbselect-controlled-required>


mfranzke:
@codex review

packages/components/configs/plugins/angular/index.cjs
```diff
@@ -6,6 +6,8 @@ module.exports = () => ({
 		post: (code) => {
 			return (
 				code
+					// Bind the native boolean property; hidden="false" still hides the element.
+					.replaceAll('[attr.hidden]=', '[hidden]=')
```
nmerget: `useMetadata({
	angular: {
		nativeAttributes: ['disabled']
	}
});`

There is a `nativeAttributes` for angular, which avoids `attr` already

packages/components/src/components/select/test-fixtures/controlled-select.fixture.lite.tsx
nmerget, line 1: This file is a new concept isn't it?

Such a component test is only testing vue and react, maybe we should create a new example (which will be generated for the showcases) and then we can create an e2e test inside the showcase folder?

packages/components/src/components/select/select.lite.tsx
```diff
@@ -335,10 +335,7 @@ export default function DBSelect(props: DBSelectProps) {
 					<option
 						class="placeholder"
 						value=""
```
nmerget, line 348:
```suggestion
						value={props.placeholder ?? "---"}
```

I tested it with the example provided by the user. The `hidden` fix didn't change anything at all. The only way I could fix it was to add a value to this <option>

packages/components/src/components/select/test-fixtures/controlled-select.fixture.lite.tsx
```diff
@@ -0,0 +1,65 @@
+import { For, useRef, useStore } from '@builder.io/mitosis';
+import DBSelect from '../select.lite';
+
+export default function ControlledSelectHarness() {
```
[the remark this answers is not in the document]
michaelmkraus: Fixed. Moved to `controlled.example.lite.tsx`. No longer in the production glob.

packages/components/configs/plugins/angular/index.cjs
michaelmkraus: Added `hidden` to `nativeAttributes` in `select.lite.tsx` and dropped the global replacement from `index.cjs`.

michaelmkraus:
Retitled and rescoped this PR to the refactor it actually is, and split the fix for #7554 out into #7860.

@nmerget your two observations were both right, and together they pointed at the real cause:

- The `hidden` change does not fix anything in the reporter's example — that example passes neither `placeholder` nor `variant="floating"`, so the internal empty option is never rendered there. This branch could not have affected it.
- Your `value` suggestion and your `onInput` workaround from the issue both "work" for the same underlying reason. `DBSelect` calls `handleValidation()` synchronously in `handleInput`. For a `required` select that just became valid, `_descByIds` flips, which is a real state change and therefore a re-render — at a moment when the consumer has not yet seen the new value, because `onChange` only fires on the native `change` event, which the browser dispatches in a *separate task*. React re-applies `value` on every commit of a `select` (`updateProperties` → `updateOptions`), so that re-render re-selects the old, empty option and the user's selection is gone. The following `change` then carries `""`.

That also explains the required/optional asymmetry: with `required={false}` validation produces no state change, React bails out of the re-render, nothing is reverted.

Your `value={props.placeholder ?? "---"}` fixes the symptom because the select is then never `valueMissing`, so the validity never flips and the re-render never happens. I would not take it though: per spec `valueMissing` depends on the selected option having an empty `value`, so a non-empty value switches off `required` validation entirely (measured: `validity.valueMissing` goes from `true` to `false`) and the placeholder string would be submitted.

#7860 defers the validation instead and adds a regression test that drives the real sequence — `input`, one frame, `change`. Worth knowing for future tests: `selectOption` cannot reproduce this at all, because Playwright dispatches both events in the same task, before the revert commit lands. That is why the suite never caught it.

On your other point about `test-fixtures` being a new concept: agreed, and it is gone from this branch. It moved to #7860 with the test that needs it, and the open question of where such a harness belongs is noted there.

nmerget:
But the hidden attribute doesn't change anything at all, right?

michaelmkraus:
Correct for the reported bug — it changes nothing there, and that is why this PR no longer claims to fix #7554. The fix is #7860.

It does change something else though, and I should have made that the reason for the PR from the start rather than leaving it implicit. The old rule was:

```scss
select:is(:has(option:checked:not(.placeholder)), :open) option[data-show-empty-option="false"] {
	display: none;
}
```

Two gaps in that:

**1. `:open` does not exist in any Safari we support.** Per [caniuse / MDN browser-compat-data](https://caniuse.com/mdn-css_selectors_open), `:open` landed in Chrome 133, Firefox 136 and Safari 26.5. Resolving our `.browserslistrc` gives:

```text
chrome 151, 150, 149
edge 151, 150, 149
firefox 140
safari 26.4, 26.3, 26.2, 26.1, 26.0, 18.5-18.7, 18.4, 18.3, 18.2, 18.1, 18.0
```

Chrome, Edge and Firefox ESR are fine, but every Safari in that list is below 26.5. There the `:open` branch never matched, so the empty option was hidden only by the *first* condition — after a non-placeholder option had been selected. Meaning: in Safari, the first time a user opens a required select, the empty option is still listed and selectable. That is exactly the case the feature was added for.

**2. Before the first interaction, neither condition matches in any browser.** The checked option *is* the placeholder, and the select is closed. So the option was rendered and exposed to assistive technology as a selected option with an empty name. That is what the aria snapshot diff in this PR reflects — `option [selected]` disappears from the combobox.

With `hidden` the state sits in the markup, so it holds from the first render, in every browser, and without depending on `:has()` or `:open` — or on our stylesheet reaching the consumer at all, which matters for the HTML/CSS package where people write the markup themselves.

The trade-offs, for the record: the option is now hidden from the start instead of after the first selection, and anyone who hand-wrote `data-show-empty-option="false"` and relied on our stylesheet has to set `hidden` instead. Both are in the changeset.

If you would rather not have the refactor on its own merits, I am fine closing this one — #7860 does not depend on it.

packages/components/src/components/select/select.lite.tsx
michaelmkraus: Right, not for the reported bug — that's why this PR no longer claims to fix #7554 (fix is in #7860).

It does change two things though. The old rule only hid the option once a non-placeholder option was checked or via :open — and :open lands in Safari 26.5, while our browserslist tops out at safari 26.4 (plus 18.x). So in every Safari we support, the empty option was still listed and selectable the first time a user opens a required select. And before any interaction no browser matches either condition, so the option was exposed to AT as a selected option with an empty name — that's the aria snapshot diff here.

packages/components/src/components/select/test-fixtures/controlled-select.fixture.lite.tsx
michaelmkraus: Fixed.

packages/components/src/components/select/select.lite.tsx
mfranzke: > Right, not for the reported bug — that's why this PR no longer claims to fix #7554 (fix is in #7860).
> 
> It does change two things though. The old rule only hid the option once a non-placeholder option was checked or via :open — and :open lands in Safari 26.5, while our browserslist tops out at safari 26.4 (plus 18.x). So in every Safari we support, the empty option was still listed and selectable the first time a user opens a required select.

besides the supporting status of `:open`, the `hidden` HTML attribute doesn't hide the `option` element in Safari either. So in both ways, for Safari it's a graceful degradation anyhow, isn't it?

mfranzke: > And before any interaction no browser matches either condition, so the option was exposed to AT as a selected option with an empty name — that's the aria snapshot diff here.

What's wrong about this? This is actually the current status of the select anyhow, isn't it ? If either the placeholder is shown or a floating label displays the label, no `option` would be selected but the first one that enables the placeholder or floating label display. So if he hide it via `hidden`, wouldn't it than instead provide an incorrect status to AT, like reporting that it wouldn't be the first, empty `option` being selected, but the first one of the list ?

michaelmkraus: Visually this remains graceful degradation because Safari does not reliably respect `hidden` on `<option>`. However, `hidden` does not change the native selection: the empty option remains selected with `value === ""`, `selectedIndex === 0`, and `valueMissing === true`. The first visible option remains unselected. I added regression assertions for this. The aria snapshots expose no selected visible option, not the first option as selected.

PR #7841, opened 2026-08-26
fix: synchronize custom select controlled reset

### LINKED ISSUE #6488: DBCustomSelect values and options reset not working

- [ ] All
- [ ] HTML
- [x] React
- [ ] Angular
- [ ] Vue
- [ ] Web components
- [ ] Power Apps

### Reproduction case

The DBCustomSelect component doesn't reset properly when both "values" and "options" props are reset at the same time: 
https://stackblitz.com/edit/db-ux-design-system-examples-tj8zpk3t?file=src%2FApp.tsx
Please click the "Reset" button after the DBCustomSelect "options" and "values" props have been updated (there is a timeout of 2 seconds).

### Expected Behaviour

The DBCustomSelect properly resets.

### Screenshots

_No response_

### Browser version

None

_No response_

None

_No response_

_No response_

michaelmkraus:
## Proposed changes

Fixes `DBCustomSelect` controlled synchronization when `values` and `options` change in the same React render. Derived selected options now always resolve to the filtered selection or an empty array, preventing stale summary labels and tags after a reset.

Adds a same-instance controlled React regression fixture covering simultaneous option/value updates, empty-array resets, `null` and `undefined` values, rendered options, checked state, and that synchronization does not emit `onOptionSelected`.

Adds a patch changeset for all packages generated from the shared Mitosis source.

resolves #6488

### Follow-up: the controlled example starts without a selection

The example was initially registered with `values={['de']}` and `selectedType="tag"`. That put a removable `DBTag` inside the interactive `<summary>` on page load, which axe-core reports as `nested-interactive` (serious, WCAG 4.1.2) and which failed the `DBCustomSelect` a11y suite in all four framework showcases.

The example now starts without a selection and lets its buttons drive it. No focusable descendant sits in the `<summary>` on load, so the scan is clean without disabling the rule. The bug under test is unaffected: "Load other options" changes `options` and `values` in the same render, which is the actual trigger — the initial preselection never was. The regression test's initial assertions follow the new empty state; tag coverage stays intact from the first click onwards.

The underlying component issue is **not** fixed here, because it needs a design decision and is a breaking DOM change. Tracked in #7866.

## Validation

- Controlled same-instance React regression: passes on chromium and mobile_chrome.
- `DBCustomSelect` axe-core suite on the built React showcase: 8 passed on chromium and mobile_chrome, zero `nested-interactive` findings. These are exactly the projects that failed before.
- `pnpm run build`: passed.
- `pnpm run build-outputs`: passed, Mitosis generated Angular, React, Vue and Stencil outputs.
- `pnpm run test`: passed (47 test files).
- `pnpm run lint`: passed with existing repository warnings.
- Firefox, WebKit and mobile Safari were not verified locally because those Playwright binaries are not installed. The fix removes the button from the DOM entirely, so it does not depend on rendering differences, but this is an assumption rather than a measurement — CI covers those projects.
- ARIA and visual snapshots are left to the regenerate pipeline. Expect the showcase jobs to fail once on the stale snapshots before `regenerate-snapshots-showcases` commits the refreshed ones. The ARIA diff is limited to the lines that the removed preselection produced: `text: Germany Germany`, `button "Remove Germany"`, `[checked]` and `button "Clear selection"`.

> An earlier revision of this description claimed `pnpm run build-outputs` was blocked by Footer `model.js` resolution failures. That was stale generated output from an unrelated local branch, not a repository problem; after removing those untracked artifacts the build completes. The note about visual snapshots differing by 252 pixels is likewise obsolete — those were regenerated in #7861.

### Code Quality

### Validation

### General

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/6488-fix-custom-select-reset>


mfranzke:
@codex review

packages/components/src/components/custom-select/test-fixtures/controlled-reset.fixture.lite.tsx
nmerget, line 1: Why do you create those new files here?

packages/components/src/components/custom-select/test-fixtures/controlled-reset.fixture.lite.tsx
```diff
@@ -0,0 +1,87 @@
...
+	values?: string[];
+	optionSelectedCalls: number;
+	recordOptionSelected: () => void;
+	setUpdated: () => void;
+	resetBoth: () => void;
+	resetValuesToNull: () => void;
+	resetValuesToUndefined: () => void;
+};
+
+export default function ControlledResetFixture() {
```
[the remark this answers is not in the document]
michaelmkraus: Fixed by moving it out of test-fixtures/ into `controlled.example.lite.tsx`.

packages/components/src/components/custom-select/test-fixtures/controlled-reset.fixture.lite.tsx
michaelmkraus: You are right, I moved it to `controlled.example.lite.tsx`.

nmerget: But it is still there

michaelmkraus: You were right, sorry for the noise: my previous reply described a change I had made on the branch of #7833 (the DBSelect harness), not here. Nothing was committed on this branch, so the fixture was still in the diff.

Now actually done in 716a5d7:

- `test-fixtures/controlled-reset.fixture.lite.tsx` and its source-only alias are gone
- replaced by `examples/controlled.example.lite.tsx`, following the existing custom-select examples (`useMetadata` storybook title, `DBButton` controls)
- the spec mounts the example and addresses it by role and text instead of `data-testid`, so the example carries no test-only attributes and doubles as a Storybook/Patternhub example
- the out-of-contract `values={null}` case is dropped from the same-instance test; it is already covered by the existing remount tests, so the example only documents the in-contract resets (empty array and `undefined`)

The unrelated move in #7833 has been reverted there.

Verified: `pnpm run build`, `pnpm run test`, knip clean, and the React CustomSelect suite in Chromium (19 passed, only the pre-existing visual snapshot diff remains).

.changeset/fix-custom-select-controlled-reset.md
```diff
@@ -0,0 +1,9 @@
+---
+"@db-ux/core-components": patch
+"@db-ux/ngx-core-components": patch
+"@db-ux/react-core-components": patch
+"@db-ux/wc-core-components": patch
+"@db-ux/v-core-components": patch
+---
+
+fix: synchronize custom select values and options when resetting
```
mfranzke:
```suggestion
fix(DBCustomSelect): synchronize values and options when resetting
```

nmerget (approved):

PR #7860, opened 2026-08-28
fix(DBSelect): keep the selection when validation runs on input

### LINKED ISSUE #7554: DBSelect: controlled value not updating when `required={true}` (React 19)

michaelmkraus:
## Proposed changes

A browser dispatches `input` and `change` for a `select` in **separate tasks**. `DBSelect` validated synchronously in `handleInput`, which changes internal state (`_descByIds` flips as soon as the value became valid) and therefore re-renders while the controlled `value` prop is still the previous one. React re-applies `value` on every commit of a `select` — `updateProperties` calls `updateOptions` unconditionally when `value != null`, and that re-selects the option matching the passed value. So the re-render discarded the selection the user had just made, before `change` was even dispatched. The subsequent `change` then carried the reverted value, which is why a controlled `required` select never received the new value.

With `required={false}` validation produces no state change, React bails out of the re-render and nothing is reverted — that is the required/optional asymmetry from the issue.

Validation now runs once the value has been propagated. `handleChange` still validates synchronously, which is safe because `props.onChange` runs first, so consumer state and validation land in the same batch.

resolves #7554

### Blast radius: DBSelect only

React routes `onChange` to the native `change` event only for `select` and `input[type=file]`, so only there does the stale window exist. Verified against the React 19.2.8 source and empirically with real interactions:

| Component | Result |
| --- | --- |
| `DBInput` `text` / `date` / `number` | unaffected — all types `DBInput` uses are in React's `supportedInputTypes`, so `onChange` fires on `input`, in the same batch as validation |
| `DBTextarea` | unaffected, same reason |
| `DBCheckbox` / `DBSwitch` | unaffected — `checkbox`/`radio` route through `click`, which precedes `input`/`change` |
| `DBInput type="file"` | unaffected in normal use. With `value=""` the selection is cleared, but a plain `<input type="file" value="">` behaves identically in React, so that is React's controlled-file-input behaviour and not this defect |

All four of these components call `handleValidation()` synchronously in the same way — it simply has no effect without the stale window.

### About the test

The regression test drives the real sequence: set the selection, dispatch `input`, wait a frame, dispatch `change`. Verified both ways — without the fix it reports `Expected "first", Received ""`, with the fix it passes.

`selectOption` cannot cover this: Playwright dispatches `input` and `change` in the same task, i.e. before the revert commit lands. That is why the defect was invisible to the existing suite, and it is worth keeping in mind for future controlled-component tests.

### Open point for review

The harness lives in `test-fixtures/`, which the production Mitosis glob (`src/**/*.{lite.tsx,ts}`, excluding only `*.agent.lite.tsx` and `*.figma.*`) still picks up, so it is generated for all four targets while only the React test needs it. Two ways out, happy to take either:

1. add a `test-fixtures` exclude for the non-React targets, or
2. move it to `examples/controlled.example.lite.tsx` (precedent: `tabs/examples/controlled.example.lite.tsx`) and cover it via showcase e2e — that would also reach Angular, Vue and Stencil, though the mechanism fixed here is React-specific.

This was already raised on #7833; the harness came from there and moves with the test that needs it.

### Code Quality

### Validation

### General

### Local validation

- Select component tests (Chromium): 5 passed, `should match screenshot` fails with 254 pixels — reproduces identically with the fix stashed, so pre-existing locally.
- `pnpm run test`: all suites green.
- Confirmed in a real browser with a real mouse: broken before, correct after, including repeated selections and the `placeholder` variant.

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/fix-select-validation-reverts-selection>


.changeset/fix-select-validation-reverts-selection.md
```diff
@@ -0,0 +1,11 @@
+---
+"@db-ux/core-components": patch
```
mfranzke:
```suggestion
```

This is JS only.

packages/components/src/components/select/test-fixtures/controlled-select.fixture.lite.tsx
nmerget, line 1: Why do we have this fixture again?

michaelmkraus: You are right, the fixture is gone.

nmerget (approved):

.changeset/fix-select-validation-reverts-selection.md
```diff
@@ -0,0 +1,10 @@
+---
+"@db-ux/ngx-core-components": patch
```
michaelmkraus, line 2: 🤖 🟡 [important] Please add `"@db-ux/core-components": patch` to this changeset. Both the root `AGENTS.md` Changesets section and `packages/components/AGENTS.md` explicitly require all five packages for consumer-facing shared component logic changes, regardless of whether the change is styling, template, logic, or properties. Without this entry, consumers of `@db-ux/core-components` will not receive this DBSelect fix in that package’s changelog.

michaelmkraus (commented):
🤖 **AI Review** — One important release-note issue found. The implementation and regression test otherwise look sound, and local build/test/lint validation passed.

PR #7901, opened 2026-08-31
fix(DBSwitch): icon no longer bleeds through a closed popover

### LINKED ISSUE #7735: Icon of switch inside a popover remains visible

- [ ] All
- [x] HTML
- [ ] React
- [ ] Angular
- [ ] Vue
- [x] Web components
- [ ] Power Apps

### Reproduction case

```
<div class="db-popover">
    <db-button aria-expanded="false" aria-haspopup="true" type="button" icon="person" variant="ghost" wrap="false">John</db-button>
    <article class="db-popover-content" data-show-arrow="true" data-spacing="medium" data-gap="true" data-animation="true" data-delay="none" data-width="fixed" data-placement="bottom-end">
        <db-switch data-theme-target="switch" size="small" visual-aid="true" show-label="true" icon-leading="sun" icon-trailing="moon">Switch me</db-switch>
    </article>
</div>
```

### Expected Behaviour

Whenever the popover is closed, the icon of the switch is hidden as well.

### Screenshots

Image

### Browser version

Chrome

_No response_

ourself

classified

nmerget:
## Proposed changes

The switch component's visual-aid icon (`::after` pseudo-element) remained visible when the switch was placed inside a closed popover. The switch thumb (`::before`) also briefly flashed when the popover closed.

**Root cause:** CSS `visibility: visible` on a child element overrides `visibility: hidden` inherited from an ancestor. The popover hides its `<article>` with `visibility: hidden`, but the switch icon had an explicit `visibility: visible` that punched through. The thumb had a `visibility` transition that delayed its disappearance for a few milliseconds.

**Fix:**
- Replace `visibility: hidden/visible` with `opacity: 0/1` on the switch icon (`input::after`), which doesn't have the inheritance override behavior
- Remove the unused `visibility` transition from the switch thumb (`input::before`) that caused the brief flash

**New example:** Added a \"Z-Index\" popover example showcasing Switch, Switch with Visual Aid, Input, Select, Checkbox, and Textarea inside popovers for regression coverage.

resolves #7735

### Code Quality

### Validation

### General

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/fix-switch-icon-popover-visibility>


mfranzke:
@codex review

packages/components/src/components/popover/showcase/popover.showcase.lite.tsx
```diff
@@ -9,6 +9,7 @@ import PopoverGap from '../examples/gap.example.lite';
 import PopoverPlacement from '../examples/placement.example.lite';
 import PopoverSpacing from '../examples/spacing.example.lite';
 import PopoverWidth from '../examples/width.example.lite';
+import PopoverZIndex from '../examples/z-index.example.lite';
```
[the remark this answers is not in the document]
nmerget: > **Do NOT manually edit showcase files** — they are generated

This is wrong. The showcase files will reference the examples. Adding a new example = add it to showcase file

packages/components/src/components/popover/examples/closed-popover-visibility.example.lite.tsx
```diff
@@ -0,0 +1,52 @@
...
+			<DBPopover
+				id="popover-zindex-switch"
+				trigger={<DBButton>Switch</DBButton>}>
+				<DBSwitch>Switch me</DBSwitch>
+			</DBPopover>
+			<DBPopover
+				id="popover-zindex-switch-visual-aid"
+				trigger={<DBButton>Switch Visual Aid</DBButton>}>
+				<DBSwitch visualAid={true}>Switch me</DBSwitch>
+			</DBPopover>
```
michaelmkraus, line 35: The snapshots capture this popover in its open state, so they do not protect against a recurrence of the reported closed-popover bleed. Please ensure that this visual-aid case remains closed in at least one regression snapshot or add a dedicated closed-state assertion.

packages/components/src/components/popover/examples/z-index.example.lite.tsx
```diff
@@ -0,0 +1,52 @@
...
+import DBButton from '../../button/button.lite';
+import DBCheckbox from '../../checkbox/checkbox.lite';
+import DBInput from '../../input/input.lite';
+import DBSwitch from '../../switch/switch.lite';
+import DBTextarea from '../../textarea/textarea.lite';
+import DBPopover from '../popover.lite';
+import { StorybookPopoverArgTypes } from './_popover.arg.types';
+
+useMetadata({
+	storybookTitle: 'Z-Index',
```
michaelmkraus: Could we name this example after the behavior being tested, such as `Closed Popover Visibility`? The underlying issue is caused by CSS visibility inheritance rather than stacking order or `z-index`.

packages/components/src/components/popover/examples/closed-popover-visibility.example.lite.tsx
```diff
@@ -0,0 +1,62 @@
...
```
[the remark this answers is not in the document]
nmerget: Good catch. `hoverPre()` sets `data-e2e-hover="true"` on every `.db-popover`, and the SCSS rule `&:is([data-e2e-hover="true"], ...) > article:not([data-open="false"])` then forces all of them open — so the visual-aid case never exercised the *closed* state the fix is about.

Fixed by adding a second visual-aid popover with `open={false}` to the example. The SCSS explicitly excludes `[data-open="false"]` from the forced-open rule, so it stays closed during the snapshot and will catch a recurrence of the icon bleeding through a closed popover (#7735). Snapshots regenerate in CI.

packages/components/src/components/popover/examples/closed-popover-visibility.example.lite.tsx
nmerget: Fixed in f596874. Added a second visual-aid popover with `open={false}` to the example. `hoverPre()` forces every other popover open via `data-e2e-hover="true"`, but the SCSS rule `> article:not([data-open="false"])` explicitly excludes it, so this one stays closed in the snapshot and now guards against a recurrence of the closed-popover icon bleed (#7735). Snapshots regenerate in CI.

packages/components/src/components/popover/examples/z-index.example.lite.tsx
nmerget: Good point, the name was misleading. Renamed to `Closed Popover Visibility` in f596874 (example file, function name, `storybookTitle`, showcase wrapper, and element ids all updated) since the root cause is CSS `visibility` inheritance, not stacking order / `z-index`.

michaelmkraus (approved):

PR #7915, opened 2026-08-31
fix(attribute-passing): deduplicate forwarded classes and preserve empty-valued attributes

### LINKED ISSUE #7000: Replace old forwarded classes instead of accumulating them
https://github.com/db-ux-design-system/core-web/blob/8de7430b6045bf0571d300c45717a8162fc4ae25/packages/components/configs/plugins/attribute-passing/index.cjs#L121-L123

**<sub><sub>P2 Badge</sub></sub>  Replace old forwarded classes instead of accumulating them**

When a dynamic host class changes from one value to another, `currentClass` already contains the previously forwarded class, so this expression appends the new value without removing the old one. For example, changing the host class from `invalid` to `valid` leaves both classes on the internal element, allowing obsolete selectors to continue affecting the component. Track and remove the prior forwarded class tokens before adding the new value.

_Originally posted by @chatgpt-codex-connector[bot] in https://github.com/db-ux-design-system/core-web/pull/5728#discussion_r3402326249_

### LINKED ISSUE #7001: Preserve empty-valued data and ARIA attributes
https://github.com/db-ux-design-system/core-web/blob/8de7430b6045bf0571d300c45717a8162fc4ae25/packages/components/configs/plugins/attribute-passing/index.cjs#L114-L118

**<sub><sub>P2 Badge</sub></sub>  Preserve empty-valued data and ARIA attributes**

When a consumer supplies an intentionally empty attribute such as `data-state=""` or `aria-label=""`, this branch removes it from the internal element rather than forwarding it. Empty-valued data attributes are commonly used as presence flags and CSS selectors such as `[data-state]`, so the component behaves differently from the host markup and from non-empty values. Distinguish an absent attribute from an attribute whose value is the empty string.

_Originally posted by @chatgpt-codex-connector[bot] in https://github.com/db-ux-design-system/core-web/pull/5728#discussion_r3402326252_

nmerget:
## Proposed changes

Fixes two bugs in the attribute-passing Mitosis plugin (`packages/components/configs/plugins/attribute-passing/index.cjs`):

**#7000 — Deduplicate forwarded classes:** When the `MutationObserver` re-triggered `enableAttributePassing`, class tokens from the parent were appended to the inner element's existing class list without removing duplicates. For example, setting `class="invalid"` twice would produce `class="db-tag invalid invalid"`. Now deduplicates class tokens via `Set` before applying them.

**#7001 — Preserve empty-valued data/aria attributes:** The `if (attr.value)` guard treated empty strings as falsy, so `data-state=""` or `aria-label=""` were removed from the inner element instead of being forwarded. These are valid and commonly used as CSS presence flags (e.g. `[data-state]`). Replaced with `attr.value != null` which lets empty strings through while still guarding against null/undefined.

Additionally, the `MutationObserver` callback now cleans up `data-`/`aria-` attributes on the inner element when they are removed from the parent host element — previously they would linger.

resolves #7000
resolves #7001

### Code Quality

### Validation

### General

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/fix-attribute-passing-bugs>


packages/components/configs/plugins/attribute-passing/index.cjs
```diff
@@ -127,7 +130,9 @@ module.exports = () => ({
 					)
 					.replace(
 						'`${currentClass ? `${currentClass} ` : ""}${value}`',
-						'`${currentClass ? currentClass : ""}${value ? ` ${value}`: ""}`'
+						// Issue #7000: Deduplicate class tokens to prevent accumulation
+						// when the observer triggers enableAttributePassing repeatedly.
+						'[...new Set(`${currentClass ? currentClass : ""}${value ? ` ${value}` : ""}`.split(" "))].filter(Boolean).join(" ")'
```
michaelmkraus, line 135: This deduplicates identical tokens, but it does not remove previously forwarded classes. For example, changing the host class from `invalid` to `valid` still results in `invalid valid` on the inner element. Could we track and remove the previously forwarded tokens before adding the new ones, as required by #7000?

packages/components/configs/plugins/attribute-passing/index.cjs
```diff
@@ -117,8 +117,11 @@ module.exports = () => ({
...
 							'          parent.removeAttribute(attr.name);',
-						// Remove attribute from child if value is empty, otherwise forward it
-						'if (attr.value) {\n' +
+						// Issue #7001: Use != null instead of truthiness check so that
+						// empty-valued attributes (e.g. data-state="" or aria-label="")
+						// are forwarded correctly. They are commonly used as presence
+						// flags for CSS selectors like [data-state].
+						'if (attr.value != null) {\n' +
 							'            element.setAttribute(attr.name, attr.value);\n' +
 							'          } else {\n' +
```
michaelmkraus, line 126: This `else` branch is unreachable because `attr` comes from `parent.attributes`, so `attr.value` is always a string. When a host attribute is removed, it is no longer part of that collection, leaving the previously forwarded attribute on the inner element. Could we handle removals via the mutation record, or remove the corresponding claim from the PR description?

mfranzke:
@codex review

PR #7939, opened 2026-09-02
feat: pagination component (beta)

michaelmkraus:
## Proposed changes

Adds the beta `DBPagination` component as a controlled pagination for Angular, React, Vue, Web Components and the CSS package.

The component provides:

- one-based `currentPage` plus `totalCount` and `pageSize`, all optional with the defaults `1`, `0` and `10`
- configurable `siblingCount` and `boundaryCount` with automatic ellipses
- previous, next and numbered page items, rendered as buttons by default and as anchors when `hrefPattern` is set
- `medium` and `small` sizes based on DB design tokens, applied to the page items and to previous/next alike
- responsive collapsing below the `sm` breakpoint: the pages next to the current one give way, so only the boundary pages, the current page and the ellipses remain
- localizable landmark, navigation and page labels, where a `pageLabel` may repeat its placeholders
- semantic `<nav><ul><li>` markup, `aria-current="page"` and native disabled boundary controls
- a controlled `onPageChange` callback without internal page mutation

A consumer-supplied `aria-label` takes precedence over the `label` prop in every target. Blank and invalid numeric values fall back to their defaults, and the page count is always at least `1`.

`hrefPattern` replaces `{page}` with the page number and turns the pages into anchors, which makes the pagination deep linkable, shareable and usable without JavaScript. Previous and next additionally carry `rel="prev"` and `rel="next"`. At the boundaries they stay native disabled buttons, because there is no page to link to and an anchor announced as a link that cannot be followed would need `aria-disabled` plus `tabindex="-1"` to be inert. Without `hrefPattern` the rendered markup is byte-identical to the button-only version.

Both responsive layouts live in the same DOM, marked up per item with `data-pagination-item`, and CSS decides which one is shown. There is no `ResizeObserver`, no `matchMedia` and no width measurement, as required by [Shift-left: HTML → CSS → JS](docs/shift-left-web-development.md). Hidden items use `display: none`, so they leave the tab order and the accessibility tree. The `flex-wrap` on the list stays as the last resort for what a viewport media query cannot see: a narrow container on a wide viewport, a high `siblingCount` or large text zoom.

Pointer targets are at least 24 by 24 CSS pixels to satisfy WCAG 2.2 SC 2.5.8. At functional density `$db-sizing-sm` resolves to 20px, so the items of the small variant fell below that floor, and the 4px gap is too small for the spacing exemption. An invisible pseudo element grows the target without touching the visible box, so the densities keep their designed proportions.

`DBPaginationItem` renders the list items and is exported alongside `DBPagination`, so the pagination is built from a real sub-component like the other components that render a `<ul>` of items. It owns the `<li>`, the split between a page and a truncation item, the button or anchor inside it, `aria-current`, `aria-hidden` and the derivation of `data-pagination-item`. `DBPagination` keeps computing which items exist and passes them down; its own API is unchanged. Previous and next stay outside the item, they are icon buttons - the same split the Figma component set makes.

The feature includes framework documentation, seven examples (density, controlled, position, truncation, link, collapsing, size), showcase pages for all five frameworks plus Patternhub under the Navigation category, 36 component tests, E2E accessibility coverage, Figma Code Connect definitions and snapshots, and a single `minor` changeset for all five affected component packages.

resolves https://github.com/db-ux-design-system/core-team/issues/3017

## Decisions

- Pagination is controlled and uses one-based page numbers.
- `totalCount` represents records, while `pageSize` determines the page count.
- The required props were made optional so that the documented defaults actually apply instead of being flagged as unreachable.
- The collapsed page list is the existing truncation algorithm run with `siblingCount` reduced to `0`, so it inherits its guarantees instead of re-deriving them: ascending unique pages, the current page always present, and no ellipsis standing in for a single page.
- Collapsing is driven by a viewport media query, not a container query. A container query would need `container-type: inline-size` on the `<nav>`, and because that implies `contain: inline-size` it collapses the current `inline-size: fit-content` to zero width — verified in a browser. Making the `<nav>` full width would be a visible layout change; being container aware instead of viewport aware is a candidate for a follow-up.
- URLs are passed as a `hrefPattern` template rather than an `hrefBuilder` callback. A template covers query parameters, path segments and fragments and can be set as a plain HTML attribute, which is what `db-pagination` needs for server-rendered usage.
- The anchors carry the same class and `data-*` attributes that `DBButton` renders. `set-basic-button` styles by class and attribute and explicitly resets `text-decoration` for anchor use, so both modes are identical in every size and density without additional CSS.
- `onPageChange` is unchanged and no `preventDefault` is called, otherwise the plain href usage would be broken. A second callback parameter is not an option either: Angular and Stencil derive their event payload as `Parameters<Required<DBPaginationProps>["onPageChange"]>[number]`, so an event argument would silently widen their public payload type.
- The truncation item stays a real element with `aria-hidden` rather than a pseudo element. Chromium exposes generated `content` as a text node, verified against the accessibility tree, so an ellipsis drawn with `::before` would be announced. It also could not express the collapsed layout, where ellipses sit between arbitrary items rather than only at the ends.
- `DBPaginationItem` follows the existing item components: `active` comes from the shared `ActiveProps`, `href` is declared locally because no other model uses `Pick<>`, and the state helper is `getActive` to match the dominant `get*` naming.
- Client-side router interception, item-per-page selection and cursor mode are outside this scope. `DBPaginationItem` is the groundwork for the first one, but consumers cannot compose the items themselves yet, so a framework `<Link>` is still not possible.

## Known follow-ups for the design review

- Sizes now match the Figma concept at regular density: page items and previous/next are 40px in the medium and 24px in the small variant, the gap is 4px, and the ellipsis is 16px wide in the small variant. The medium ellipsis is 24px against 22px in the concept, because our value is a `min-inline-size` while the concept frame has a fixed width.
- Client-side routing is not integrated. The component exposes no click event object, so a consumer intercepts the click on a wrapper element and reads the `href` from the anchor. Documented for React, Angular and Vue. Letting consumers pass their own `<Link>` needs a composition API on top of `DBPaginationItem` and would change what `DBPagination` computes, so it is its own step.
- In Angular and Stencil a `<db-pagination-item>` host wraps every `<li>`, the same as `db-accordion-item` and `db-tab-item`. `display: contents` keeps layout and the accessibility tree intact, but only the pipeline can confirm that Axe still accepts the list semantics there.
- `DBPagination` and `DBPaginationItem` have no `agent/` examples. Three of the five existing item components do, so this is worth catching up for the whole family rather than for the item alone.
- `key` reconciliation only applies to React and Stencil; Angular and Vue track by position. Relevant once focus management is added.
- A page that has focus while the viewport shrinks loses it to the document, because `display: none` removes the element from the flow. Restoring it would require JavaScript.
- Every small icon-only `.db-button` is 20px at functional density, not just in this component. The pointer target is corrected inside the pagination only; a fix in `_button-components.scss` would affect the whole system and needs its own discussion.

## Snapshots

The pipeline already regenerated the snapshots for the previous and next size, which grew from 24px to 40px in the medium variant and made the component 32px wider.

Still open: extracting the items adds `class="db-pagination-item"` and a per-item `data-size` to every `<li>`, and in Angular and Stencil a `<db-pagination-item>` host around it. React and Vue render pixel-identical - the component screenshot differs only by the 56 pixels of macOS font antialiasing at unchanged dimensions - so the showcase snapshots of the Angular and Stencil targets are the ones to watch.

### Code Quality

### Validation

### General

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/3017-pagination-initial-development>


nmerget:
image
Can we move densitiy as the first example?

Why are the next and previous button size=small but the numbers size=medium?

.changeset/bright-pages-turn.md
nmerget, line 1: We shouldn't add 5 different changesets for the same new component

nmerget:
There should be a seperated `db-pagination-item` component to align it with all the other `<ul>` and `<li>` components we already have.

Also this enables the consumer to use a custom `<Link>` component to make it to an anchored/link pagination.
This requires some more JS to align everything but it allows to use option and composition api again.

So the `ellipsis` might be just an `:before` (first item) or `:after` (last-item) which will be ignored for the screen-readers because it's just for the visual users. With this you can easy use the `DBPaginationItem`. Furthermore there should be an option to set the previous or the next button maybe also a first and last button in the future. We can control it with JS or CSS to show it visually correct.

nmerget:
image


Using a iPhone SE layout the next button moves to the next line, I guess this is not intended? Maybe we should allow only 999 pages? Everything else is bad anyways from a UX perspective

.changeset/bright-pages-turn.md
michaelmkraus: Fixed.

mfranzke:
@codex review

packages/components/scripts/post-build/components.ts
```diff
@@ -82,6 +82,13 @@ const controlPanelActionsComponents: Component[] = [
 }));
 
 export const getComponents = (): Component[] => [
+	{
+		name: 'pagination-item'
```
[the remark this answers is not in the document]
michaelmkraus: Fixed. `onClick` is now in the item's React propsPassingFilter.

packages/components/AGENTS.md
```diff
@@ -377,6 +396,38 @@ Mitosis compiles `.lite.tsx` to multiple frameworks. Be aware of these constrain
...
+[Shift-left: HTML → CSS → JS](../../docs/shift-left-web-development.md).
+
+`DBPagination` is the reference. Its `<li>` elements carry
+`data-pagination-item` (`page`, `sibling`, `ellipsis`, `collapse-ellipsis`,
+`wide-ellipsis`) and `pagination.scss` toggles `display` per layout inside
+`screen-sizes.screen("sm", "max")`. Three things made it work:
+
+- **Derive each layout from the same function.** The collapsed page list is the
+  existing algorithm run with `siblingCount` reduced to `0`, so it inherits every
+  guarantee that algorithm already makes instead of re-deriving them by hand.
```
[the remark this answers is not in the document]
michaelmkraus: The bullet no longer sais the collapsed list comes from the wide algorithm with `siblingCount` `0`, and instead documents `getCollapsedPages`.

packages/components/scripts/post-build/components.ts
```diff
@@ -82,6 +82,13 @@ const controlPanelActionsComponents: Component[] = [
...
+	},
+	{
+		name: 'pagination'
```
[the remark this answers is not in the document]
michaelmkraus: Applied for consistency with `tabs`, though I could not reproduce it.

packages/components/src/components/pagination-item/pagination-item.lite.tsx
```diff
@@ -0,0 +1,106 @@
...
+			class={cls(
+				'db-pagination-item',
+				state.getPage() === 0 ? 'db-pagination-ellipsis' : undefined,
+				props.className
+			)}
+			ref={_ref}
+			data-pagination-item={state.getItemAttribute()}
+			data-size={props.size}
+			aria-hidden={state.getPage() === 0 ? 'true' : undefined}>
+			<Show when={state.getPage() > 0} else={<span>...</span>}>
```
nmerget: Instead of having another `<li><span>...</span></li>` can we just set this as an `::after` or `::before`. We can move it to the `.db-pagination` where we set it to the first/last pagination-item (besides of the chevrons). 

We just need to set the size of the `::before,::after` to `--db-sizing-medium` and add the same gap as a margin.

The content has to be `content: ... / ""` to avoid reading it for screen-reader users.

packages/components/src/components/pagination-item/pagination-item.lite.tsx
```diff
@@ -0,0 +1,106 @@
...
+				<Show
+					when={props.href}
```
nmerget, line 87: Instead of providing a `props.href` we should provide a `props.text`.

If we use `props.text` we render a `button` otherwise `props.children`. Also we should move all the layouts to the `<li>` instead of the `<DBButton>`. Then we can apply the styling based on the ghost/filles button.

This allows us to use something like this.

`<DBPagination><DBPaginationItem><Link to="bla">My Label will be overwritten by a number anyways</Link></DBPaginationItem></DBPagination>`

What we need to do then is that we don't apply any `onClick` directly in the pagination-item.lite.tsx, instead control everything directly inside the pagination.lite.tsx.
We do the same in the tabs, where we control everything there.

packages/components/src/components/pagination/pagination.lite.tsx
```diff
@@ -0,0 +1,436 @@
...
+							data-variant="ghost"
+							aria-label={props.previousLabel}
+							onClick={() =>
+								state.handlePageChange(
+									state.getCurrentPage() - 1
+								)
+							}>
+							{props.previousLabel}
+						</a>
+					</Show>
```
nmerget, line 461: If you follow [this](https://github.com/db-ux-design-system/core-web/pull/7939/changes#r3947112472) approach. You can use `<DBPaginationItem>` here instead as well

packages/components/src/components/pagination/pagination.lite.tsx
```diff
@@ -0,0 +1,436 @@
...
+							page={item.page}
+							layout={item.layout}
+							size={props.size}
+							active={state.getCurrentPage() === item.page}
+							href={state.getHref(item.page)}
+							label={state.getPageLabel(item.page)}
+							onClick={() => state.handlePageChange(item.page)}
+						/>
+					)}
+				</For>
```
nmerget: We should wrap it into a `<Show when={state.getPaginationItems().length} else{props.children}` to allow the composition as well

nmerget:
https://github.com/user-attachments/assets/2bd11ee8-db10-4254-93ed-c364446c0991

I guess this behavior isn't correct as well? It should be always 1 row right?

packages/components/src/components/pagination/pagination.scss
```diff
@@ -0,0 +1,87 @@
...
+			position: absolute;
+			inset-block-start: 50%;
+			inset-inline-start: 50%;
+			translate: -50% -50%;
+			inline-size: 100%;
+			block-size: 100%;
+			// stylelint-disable-next-line db-ux/use-sizing
+			min-inline-size: 1.5rem;
+			// stylelint-disable-next-line db-ux/use-sizing
+			min-block-size: 1.5rem;
```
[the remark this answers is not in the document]
michaelmkraus: Changed to `max(24px, 1.5rem)`

packages/components/src/components/pagination/pagination.lite.tsx
```diff
@@ -0,0 +1,431 @@
...
+								)
+							}>
+							{props.previousLabel}
+						</a>
+					</Show>
+				</li>
+				<For each={state.getPaginationItems()}>
+					{(item: PaginationItemType, index: number) => (
+						<DBPaginationItem
+							key={'pagination-item-' + index}
```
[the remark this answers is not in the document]
michaelmkraus: Items are now keyed by page identity instead of position, so the focus stays on the page that was activated.

nmerget:
image

It shouldn't be possible to click an active item again, also the hover/pressed effects shouldn't appear

packages/components/src/components/pagination-item/pagination-item.lite.tsx
michaelmkraus: The component now uses `TextProps`. It renders the control if text is provided; otherwise, it renders children. Click events are now handled by the parent. One exception: We use `totalCount` to decide what to render, instead of checking if children exist.

packages/components/src/components/pagination-item/pagination-item.lite.tsx
michaelmkraus: You were right about `content: … / ""`. Two adjustments: data attributes instead of `:first-child` (with `boundaryCount: 2` the gap sits after the second item, and each layout needs its own marker), and size stays `sm` since `md` widens the collapsed layout by ~30px. A plain `content: "..."` sits in front as a fallback so the dots can't vanish where the syntax is unknown.

packages/components/src/components/pagination/pagination.lite.tsx
michaelmkraus: The arrows go through `DBPaginationItem` as children now, no extra props needed. They pass no `page`, which keeps them out of `data-pagination-item `and therefore out of the collapsing and the delegation. `pagination.scss` lost the separate arrow box and the separate touch target overlay, both come from the item rule now.

packages/components/src/components/pagination/pagination.lite.tsx
michaelmkraus: Done, but keyed on `totalCount` rather than the list length. Angular also can't test projected content, so `props.children` isn't an option either; the accordion keys on its option prop for the same reason.

mfranzke:
@codex review

packages/components/src/components/pagination/pagination.lite.tsx
```diff
@@ -0,0 +1,465 @@
...
+		},
+		getTotalPages: () => {
+			const totalCount = state.getInteger(props.totalCount, 0, 0);
+			const pageSize = state.getInteger(props.pageSize, 10, 1);
+			return Math.max(1, Math.ceil(totalCount / pageSize));
+		},
+		getCurrentPage: () => {
+			return Math.min(
+				state.getTotalPages(),
+				state.getInteger(props.currentPage, 1, 1)
```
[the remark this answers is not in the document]
michaelmkraus: Right, and it also disabled both arrows: with the fallback total of 1, next was `1 >= 1`. The clamp now only applies in option mode, `getHref` got the same treatment so the next link is no longer dropped, and next stays enabled in composition because the end of the list belongs to the consumer.

packages/components/src/components/pagination/pagination.lite.tsx
```diff
@@ -0,0 +1,465 @@
...
+							{props.previousLabel}
+						</a>
+					</Show>
+				</DBPaginationItem>
+				{/* totalCount is the discriminator, not the presence of children.
+				Angular can only test inputs, never projected content, which is why the
+				accordion keys on its option prop as well. Without totalCount the
+				consumer owns the item list and with it the truncation, because the
+				component cannot know which pages the children stand for. */}
+				<Show when={props.totalCount}>
```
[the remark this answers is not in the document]
michaelmkraus: Fixed. A new `hasTotalCount()` tests whether the prop is set instead of whether it is truthy - the same distinction `getInteger` already draws for a blank value - and it is used for both mode branches and the upper-bound guard, so `0` and `"0"` now behave the same in every framework.

packages/components/src/components/pagination-item/pagination-item.lite.tsx
```diff
@@ -0,0 +1,104 @@
...
+		// The attribute is the contract the stylesheet works against: a page is part
+		// of both layouts, a sibling only of the wide one. An item without a page is
+		// not a page at all - the previous and next controls use this component as a
+		// shell - and stays out of the contract, so the collapsing never touches it
+		// and the tests do not read it as a page.
+		getItemAttribute: () => {
+			if (state.getPage() === 0) {
+				return undefined;
+			}
+			return props.layout === 'wide' ? 'sibling' : 'page';
```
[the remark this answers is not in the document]
michaelmkraus: Fixed: `collapsed` gets its own `data-pagination-item` value, hidden by default and shown below the breakpoint, mirroring `sibling`. `DBPagination` still never emits it - its own collapsed pages are a subset of the wide ones - so the value exists purely for composed lists, which the test covers.

packages/components/src/components/pagination/docs/React.md
```diff
@@ -0,0 +1,150 @@
...
+import { DBPaginationItem } from "@db-ux/react-core-components";
+
+const App = () => (
+	<nav className="db-pagination" aria-label="Pagination">
+		<ul>
+			<DBPaginationItem
+				page={1}
+				label="Page 1 of 2"
+				onClick={() => console.log("page 1")}
+			/>
```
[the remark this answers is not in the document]
michaelmkraus: Correct, the items had neither `text` nor children. All three framework examples now pass `text`, and the paragraph below says so - it also no longer claims that leaving `page` out renders a truncation item, which stopped being true when the ellipsis became a pseudo element.

packages/components/src/components/pagination/docs/React.md
```diff
@@ -0,0 +1,150 @@
...
+the item, so your child never gets a handler attached to it.
+
+```tsx App.tsx
+import { DBPagination, DBPaginationItem } from "@db-ux/react-core-components";
+import { Link } from "react-router-dom";
+
+const App = () => (
+	<DBPagination currentPage={2} onPageChange={(page) => console.log(page)}>
+		<DBPaginationItem page={1} label="Page 1 of 2">
+			<Link to="/results/1">1</Link>
```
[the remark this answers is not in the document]
michaelmkraus: Fixed: the composed links in the React, Angular and Vue examples carry `aria-label`, and the section now states that `label` only names the control the item renders itself and never reaches a child you provide.

packages/components/src/components/pagination/pagination.lite.tsx
```diff
@@ -0,0 +1,465 @@
...
+export default function DBPagination(props: DBPaginationProps) {
+	const _ref = useRef<HTMLElement | any>(null);
+
+	const state: DBPaginationState = useStore<DBPaginationState>({
+		getInteger: (
+			value: number | string | undefined,
+			fallback: number,
+			minimum: number
+		) => {
+			// Number('') and Number(null) both return 0, which is finite and would
```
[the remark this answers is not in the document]
michaelmkraus: Reworded to "An empty string and null both convert to 0".

packages/components/src/components/pagination/docs/HTML.md
```diff
@@ -0,0 +1,290 @@
...
+`set-basic-button` resets `text-decoration` for exactly this case, so an `<a>` with
+`class="db-button"` looks identical to the `<button>`. Swap the element, keep every
+class and `data-*` attribute, and drop `type="button"`:
+
+```html index.html
+<li
+	class="db-pagination-item"
+	data-pagination-item="sibling"
+	data-page="4"
+	data-size="medium"
```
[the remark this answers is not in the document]
michaelmkraus: Right, added `data-variant="ghost"` to the list item in that example.

packages/components/AGENTS.md
```diff
@@ -377,6 +396,48 @@ Mitosis compiles `.lite.tsx` to multiple frameworks. Be aware of these constrain
...
+  boundary pages, the current page, and nothing else. What both must share is the
+  set of invariants — ascending unique pages, the current page always present, no
+  ellipsis standing in for a single page — and the collapsed pages must stay a
+  subset of the wide ones, because that is what lets one list of items carry both.
+  Assert those invariants for both layouts in the spec instead of deriving one
+  from the other.
+- **The narrow layout is not a subset of the rendered wide layout.** Removing
+  items opens gaps that need their own separators, so a layout can need elements
+  the other one does not have at all. Emit them and hide them in the other
+  layout, rather than trying to reuse one element for both.
```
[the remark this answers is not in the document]
michaelmkraus: Fixed, and two more stale bits in the same section: the marker list named `ellipsis`, `collapse-ellipsis` and `wide-ellipsis`, which no longer exist (it is `page`, `sibling`, `collapsed` now), and the first bullet still claimed the collapsed layout renders the `boundaryCount` pages at each end instead of one. The second bullet now says a gap belongs to a layout and is drawn as a pseudo element per layout, plus the new invariant that a layout has to render the pages it pins.

packages/components/scripts/post-build/components.ts
```diff
@@ -82,6 +82,29 @@ const controlPanelActionsComponents: Component[] = [
...
 export const getComponents = (): Component[] => [
+	{
+		name: 'pagination-item',
+		config: {
+			// The React pass-through forwards every prop starting with `on` to the
+			// root, which here is the <li>. The button or anchor inside it already
+			// calls onClick through handleClick, so a click would bubble into the
+			// forwarded handler as well and invoke the consumer callback twice.
+			react: {
+				propsPassingFilter: ['onClick']
```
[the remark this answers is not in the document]
michaelmkraus: Resolved by removing `onClick` rather than restoring it. The item never attached it anyway: `DBPagination` delegates once on the list and reads `data-page` back, which is what keeps a composed child untouched. In React the filter made the prop inert while it still bound natively in the other frameworks, so the same prop behaved differently per target. `ClickEventProps` is gone from the item and the filter with it, so the delegation is the single click path - a standalone consumer owns the surrounding `<ul>` and the child, and can listen there.

nmerget:
image
The small pagination-items aren't square, is this on purpose?

packages/components/src/components/pagination-item/pagination-item.lite.tsx
```diff
@@ -0,0 +1,111 @@
...
+			ref={_ref}
+			data-pagination-item={state.getItemAttribute()}
+			data-page={state.getPage() > 0 ? state.getPage() : undefined}
+			data-size={props.size}
+			data-ellipsis-wide={props.wideEllipsis}
+			data-ellipsis-collapsed={props.collapsedEllipsis}
+			data-variant={state.getActive() ? 'filled' : 'ghost'}
+			aria-current={
+				!props.text && state.getActive() ? 'page' : undefined
+			}>
```
nmerget, line 77: All those attributes will be controlled by the DBPagination right?

Why don't we handle everything there? This component should just be a wrapper to apply the `<li>` and the `.db-pagination-item`. Everything else is handled by the Pagination itself.

We should even replace `<DBButton>` with a `<button>` because we need a custom styling anyways.

Maybe we can get rid of the `text` property as well because we need to set the number anyways based on the index of the DBPaginationItem right?

There should be 3 ways to use a Pagination:

````
<DBPagination>
  <DBPaginationItem label="Go to first page" />
  <DBPaginationItem label="Go to second page" />
</DBPagination>
````

````
<DBPagination>
  <DBPaginationItem><a href="#1">Go to first page</a></DBPaginationItem>
  <DBPaginationItem><a href="#2">Go to second page</a></DBPaginationItem>
</DBPagination>
````

````
<DBPagination size="2"/>
````

File: packages/components/src/components/pagination/pagination.lite.tsx [hunk 20]
```diff
@@ -0,0 +1,491 @@
...
+							aria-label={props.nextLabel}
+							onClick={() =>
+								state.handlePageChange(
+									state.getCurrentPage() + 1
+								)
+							}>
+							{props.nextLabel}
+						</a>
+					</Show>
+				</DBPaginationItem>
```
nmerget, line 520: Instead of using a `DBPaginationItem` here we can just use a `<li class="db-pagination-next">` (same applies to the previous).

Instead of splitting with `state.getNextHref()` to a button and anchor, we could just rely on the <DBButton>. The logic could be just to show always the next item (if there is one) so the collapsed state might change a little bit. If we always have the next item available we can easily search for it and do a `.click()` on it.

Why shall we do it?

1. This helps to reduce the code + complexity inside this component a lot
2. User provided anchors (especially custom components with a own router) will be triggered, not a plain `<a>`

michaelmkraus:
> image The small pagination-items aren't square, is this on purpose?

That was a bug. They are 24×24 now, as in the concept.

packages/components/src/components/pagination-item/pagination-item.lite.tsx
michaelmkraus: Done. The item renders a plain` <button>/<a>` and takes its appearance from the shared button placeholders instead of `DBButton`. `data-variant `and `data-size` now sit on the `<li>` only, so a composed child is styled from there too.

packages/components/src/components/pagination/pagination.lite.tsx
michaelmkraus: Added the `.click()` for the buttons. Previous and next now click the item of the next or previous page, so a `<Link>` from the consumer runs too. If that item is not there (`siblingCount: 0`), they just report the page like before. The anchors stay real links, otherwise they lose `rel`, middle click and no-JS.

I kept the arrows inside `DBPaginationItem`, because taking them out is the opposite of what you asked before.

mfranzke:
@codex review

packages/components/src/components/pagination-item/pagination-item.lite.tsx
```diff
@@ -0,0 +1,110 @@
...
```
nmerget, line 77: Let's talk about this again, I'm pretty sure we don't need all of those properties here and should reduce this component even more

packages/components/src/components/pagination/examples/density.example.lite.tsx
```diff
@@ -0,0 +1,97 @@
...
+export default function PaginationDensity() {
+	const state = useStore<PaginationDensityState>({
+		functionalPage: 5,
+		regularPage: 5,
+		expressivePage: 5,
+		// Routed through state methods with an `any` parameter: an inline typed
+		// callback breaks the Angular showcase, where $event is number | void, and
+		// the Stencil showcase, where the payload is a CustomEvent.
+		setFunctional(page: any) {
+			state.functionalPage = page;
```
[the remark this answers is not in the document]
michaelmkraus: Confirmed against the generated output. All setters across the pagination examples now unwrap with `page?.detail ?? page`.

packages/components/src/components/pagination/docs/HTML.md
```diff
@@ -0,0 +1,273 @@
...
+page give way, so only the boundary pages and the current page remain. Both layouts
+live in the same markup, which is why the list carries more items than any single
+layout shows.
+
+Which pages are shown is one attribute:
+
+| `data-pagination-item` | Rendered                  | Meaning                                                       |
+| ---------------------- | ------------------------- | ------------------------------------------------------------- |
+| `page`                 | always                    | Boundary page, current page, or a page closing a one-page gap |
+| `sibling`              | only above the breakpoint | Page next to the current one                                  |
```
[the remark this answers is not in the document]
michaelmkraus: Added, the table was incomplete. `collapsed` is now documented as narrow-only next to `page` and `sibling`, with a note that `DBPagination` never emits it.

packages/components/src/components/pagination/pagination.lite.tsx
```diff
@@ -0,0 +1,524 @@
...
+			}
+
+			// Guard against a nested pagination: only handle items of this instance.
+			if (item.closest('.db-pagination') !== _ref) {
+				return;
+			}
+
+			const page = Number(item.getAttribute('data-page'));
+			if (Number.isFinite(page)) {
+				state.handlePageChange(page);
```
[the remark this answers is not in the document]
michaelmkraus: Fixed. `isModifiedLinkClick` rejects an activation the browser is about to handle elsewhere, and it is applied in both places that report a page. Two tests cover it.

packages/components/src/components/pagination/docs/Angular.md
```diff
@@ -0,0 +1,137 @@
...
+
+Leave `totalCount` out and pass the items yourself. The pagination then projects your
+content instead of computing the page list, which is what lets you bring a router
+link. It still reports the page: it listens on the list and reads `page` back from the
+item, so your child never gets a handler attached to it.
+
+```html app.component.html
+<db-pagination [currentPage]="2" (pageChange)="onPageChange($event)">
+	<db-pagination-item [page]="1" label="Page 1 of 2">
+		<a routerLink="/results/1" aria-label="Page 1 of 2">1</a>
```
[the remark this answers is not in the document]
michaelmkraus: The composition section now has its own setup block importing `DBPagination`, `DBPaginationItem` and `RouterLink` together.

packages/components/src/components/pagination/pagination.lite.tsx
```diff
@@ -0,0 +1,524 @@
...
+		},
+		// The current page gets no href. It is not somewhere to go, so a link to it
+		// promises a change and delivers none - the item keeps its focus and its
+		// aria-current, but stops being a link, the way the ARIA APG treats the last
+		// breadcrumb item. Previous and next keep using getHref, they can never point
+		// at the current page.
+		getPageHref: (page: number) => {
+			if (page === state.getCurrentPage()) {
+				return undefined;
+			}
```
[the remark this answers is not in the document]
michaelmkraus: Fixed. The current page keeps its `href`, so the node survives the update and focus stays. `getPageHref` is gone. The APG breadcrumb example marks the last link with `aria-current` rather than dropping the link, so the old reasoning was wrong. Docs and tests updated. The showcase aria snapshots now expose `link` for the current page.

packages/components/src/components/pagination-item/pagination-item.lite.tsx
michaelmkraus: Reduced the item as far as possible: `size` removed and the two ellipsis props are now one `data-ellipsis`.

The rest cannot be move, because the `<li>` lives inside the item and a parent cannot set attributes on a child's root. In Angular and Stencil there is even a host in between. `page` (identity the delegation reads back, and what the arrows step to), `active` for `aria-current`, `layout` for the collapsing, `label` for the accessible name, href, and text have to stay.

nmerget: Why can't we set it?

We do it already for other components. Please create a new example with option API and composition API, then you will see the problems with the current implementation. For composition API the user has to set props, which are available only in the DBPagination (and the pagination should control all items). The users should not set those properties themselfs.

Instead we can just use the DOM to set all those properties correctly, no matter if we are in angular/stencil or somewhere else. We just need to use `_ref.querySelectorAll("li")` inside DBPagination and we can safe this in the state if we want to. We are doing the exact same thing inside the Tabs. From then we can search for the `button` or `a` inside the `li` and set all properties there. No need for `aria-current` on the `li`, we can set it directly to the `a` for example. If we follow this, there won't be any properties (except of `text` and `children`) for the `pagination-item/model.ts`, all properties are global properties like the truncation etc. and should be controlled by the DBPagination. 
Furthermore, with this approach we can remove the `props.href` we don't need it. Users should throw in their own `<a>` with composition or use the `props.totalCount` inside the DBPagination if they want to use the optionApi. This means we will can pass a `<a>` inside the `DBPagination` to the `DBPaginationItem`, instead of hardcode it in the DBPaginationItem

For `props.totalCount`, I like the idea, but I would recommend to allow `items` as well. Right now every components with list-item children, has this `options` or `items` property where they can control the list-items. I think it would be convinient to continue this pattern, this would allow to control single items with optionApi as well, for exampel if they want to disable an item (which is currently missing in the design). So the `props.totalCount` will generate an internal `state._items` if `props.items` are not set by the user.

Another thing to consider is when there is compositionApi how to we handle the placement of the correct labels like 1, 2, 3 etc. I guess a property for the DBPagination makes sense which is on `auto` and can be toggled to `manual`. If we have auto, it should use the text of the `<a>` or the `<button>` and move it to the `aria-label` and replaces the innerText with `1` or `2` etc.

packages/components/src/components/pagination/pagination.lite.tsx
```diff
@@ -0,0 +1,562 @@
...
+				<DBPaginationItem>
+					{/* Plain controls, for the same reason the item renders one: the
+					box comes from .db-pagination-item and the look from the shared
+					button placeholders, so both branches are identical without a line
+					of extra CSS. data-icon is the only thing an arrow declares - the
+					item hides the label and centres the glyph off that attribute -
+					while data-variant sits on the list item around it and the size comes
+					from the data-size of the nav. */}
+					<Show
+						when={state.getPreviousHref()}
```
nmerget: This will not work for compositionAPI, because we don't controll the `<a>`.

Instead we should just use a normal `<button>` here with an `onClick` which will just find the previous `li > :is(button,a)` and do a normal `.click()` on it. This would allow us the same function for optionApi and compositionApi

michaelmkraus: Done in the wrapper rebuild.

packages/components/src/components/pagination-item/pagination-item.lite.tsx
michaelmkraus: Done. `DBPaginationItem` is a thin wrapper (`text` + `children` + `disabled`); `DBPagination` writes `data-page`, `data-pagination-item`, `data-ellipsis`, `data-variant` and `aria-current` onto the `<li>` and its control through the DOM via `querySelectorAll`, the same as tabs. `href` is gone - links come from composition, the consumer passes their own `<a>`/`<Link>`. items drives the data mode and can disable a single page. The numbering is automated: a composed control's text moves into `aria-label` and the number is shown.

packages/components/src/components/pagination/pagination.lite.tsx
```diff
@@ -0,0 +1,496 @@
...
+								: undefined
+						}
+						aria-label={props.previousLabel}
+						onClick={() =>
+							state.stepToPage(state.getCurrentPage() - 1)
+						}>
+						{props.previousLabel}
+					</button>
+				</li>
+				<Show when={state.isDataDriven()}>
```
nmerget:
```suggestion
				<Show when={state.isDataDriven()} else={props.children}>
```

This should be possible as well

packages/components/src/components/pagination/pagination.lite.tsx
```diff
@@ -0,0 +1,496 @@
...
+					<For each={state.getPaginationItems()}>
+						{(item: PaginationItemType, index: number) => (
+							<DBPaginationItem
+								key={item.key}
+								text={state.getPageText(item.page)}
+							/>
+						)}
+					</For>
+				</Show>
+				<Show when={!state.isDataDriven()}>{props.children}</Show>
```
nmerget:
```suggestion
```

__snapshots__/pagination/showcase/chromium/DBPagination-should-match-screenshot-1/DBPagination-should-match-screenshot.png
nmerget, line 1: Something is currently wrong with the previous/next buttons design wise

packages/components/src/components/pagination/examples/link.example.lite.tsx
```diff
@@ -0,0 +1,51 @@
...
+				</DBInfotext>
+				<DBPagination
+					label="Linked pagination"
+					currentPage={state.linkedPage}
+					onPageChange={(page: any) => state.setLinked(page)}>
+					<DBPaginationItem>
+						<a href="#linked-page=1">1</a>
+					</DBPaginationItem>
+					<DBPaginationItem>
+						<a href="#linked-page=2">2</a>
```
nmerget, line 42: I guess we need to set `pointer-events: none` and or tabindex=-1 on the current selected element for a link, otherwise we can click it, even though there is no hover/press effect

packages/components/src/components/pagination-item/pagination-item.lite.tsx
```diff
@@ -0,0 +1,22 @@
...
+export default function DBPaginationItem(props: DBPaginationItemProps) {
+	const _ref = useRef<HTMLLIElement | any>(null);
+
+	return (
+		<li
+			id={props.id ?? props.propOverrides?.id}
+			class={cls('db-pagination-item', props.className)}
+			ref={_ref}>
+			<Show when={props.text} else={props.children}>
+				<button class="db-pagination-page" type="button">
```
nmerget: Why do we need this class? Only for the tests? We should remove it and use another locator for the tests instead

packages/components/src/components/pagination-item/model.ts
```diff
@@ -0,0 +1,11 @@
+import type { GlobalProps, GlobalState, TextProps } from '../../shared/model';
+
+export type DBPaginationItemDefaultProps = {
+	disabled?: boolean | string;
```
nmerget: We don't use it inside the `.lite.tsx` file don't we?

packages/components/src/components/pagination/model.ts
```diff
@@ -0,0 +1,136 @@
...
+	key: string;
+};
+
+/**
+ * An entry of the `items` prop. A consumer who drives the pagination through data
+ * rather than composed children passes these; `disabled` greys a single page out,
+ * which the design otherwise cannot express.
+ */
+export type PaginationItemOption = {
+	disabled?: boolean | string;
```
nmerget, line 43:
```suggestion
	disabled?: boolean | string;
	label?: string;
```

I think it would make sense to give the people the chance to customize the label for the items, only having `disabled` here feels a little bit wrong

mfranzke:
@codex review

packages/components/src/components/pagination/pagination.lite.tsx
michaelmkraus: Done, both branches are one Show with else now.

packages/components/src/components/pagination/pagination.lite.tsx
michaelmkraus: Removed.

__snapshots__/pagination/showcase/chromium/DBPagination-should-match-screenshot-1/DBPagination-should-match-screenshot.png
michaelmkraus: The arrows moved out of `DBPaginationItem`, which used to set `data-variant`, and `syncItems` skips the arrow shells — so the rules keyed on it stopped applying and the arrows kept an adaptive border and font-weight 700. They carry `data-variant="ghost"` directly now.

packages/components/src/components/pagination-item/pagination-item.lite.tsx
michaelmkraus: Class removed, tests locate a page control via `li[data-page]`.

packages/components/src/components/pagination-item/model.ts
michaelmkraus: Right, removed it.

packages/components/src/components/pagination/model.ts
michaelmkraus: Added. An items entry can carry a `label` that wins over the `pageLabel` template; the visible text stays the page number.

packages/components/src/components/pagination-item/model.ts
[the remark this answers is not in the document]
michaelmkraus: Removed. `DBPaginationItemProps` is now `GlobalProps & TextProps` only, so the prop is gone.

packages/components/src/components/pagination/docs/React.md
```diff
@@ -0,0 +1,89 @@
...
+gets a handler attached to it.
+
+```tsx App.tsx
+import { DBPagination, DBPaginationItem } from "@db-ux/react-core-components";
+import { Link } from "react-router-dom";
+
+const App = () => (
+	<DBPagination currentPage={2} onPageChange={(page) => console.log(page)}>
+		<DBPaginationItem>
+			<Link to="/results/1">1</Link>
```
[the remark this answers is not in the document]
michaelmkraus: Fixed. The inconsistency was the other way round, though: React.md was the updated file, while Angular.md and Vue.md still described the pre-refactor behaviour. The React example now sets `aria-label` on both Links, and all three composition sections document the numbering automation identically.

packages/components/src/components/pagination/pagination.lite.tsx
```diff
@@ -0,0 +1,496 @@
...
+				(typeof event.button === 'number' && event.button > 0)
+			);
+		},
+		handleClick: (event: any) => {
+			if (state.isModifiedLinkClick(event)) {
+				return;
+			}
+
+			const target = event.target as HTMLElement;
+			const item = target.closest('li.db-pagination-item');
```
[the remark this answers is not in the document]
michaelmkraus: `handleClick` now requires the event to originate from the item's `a`/`button` before reading `data-page`, so a click on the `::before`/`::after` ellipsis (or on the `li` padding) reports nothing. Regression test added.

packages/components/src/components/pagination/examples/link.example.lite.tsx
michaelmkraus: The click was indeed doing nothing useful while still navigating. Rather than `pointer-events: none` / `tabindex="-1"` I stopped the navigation itself: a click on the current page now calls `preventDefault()` and reports no change. A bubbling guard covers plain anchors, and a capture-phase listener on the `<nav>` runs before a consumer router link so composition links on the current page are stopped too.

I kept the link focusable and in the tab order on purpose: the APG breadcrumb pattern keeps the current item a link with `aria-current="page"`, and we need the anchor node to survive a controlled update so keyboard focus stays on the page that was just activated (an earlier thread flagged the focus loss). `pointer-events: none` would also drop the context menu and middle-click. Regression test added.

nmerget (dismissed):

packages/components/src/components/pagination/model.ts
```diff
@@ -0,0 +1,139 @@
...
+ * breakpoint would take its ellipsis with it. One token per side, so a page standing
+ * between two gaps carries two of them instead of a value meaning both.
+ */
+export const PaginationEllipsisList = [
+	'wide-before',
+	'wide-after',
+	'collapsed-before',
+	'collapsed-after'
+] as const;
+export type PaginationEllipsisType = (typeof PaginationEllipsisList)[number];
```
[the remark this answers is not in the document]
michaelmkraus: Removed instead of tagged. `PaginationEllipsisList` and `PaginationEllipsisType` were unused leftovers from before the wrapper rebuild.

showcases/patternhub/components/component-parser/data.ts
```diff
@@ -6,6 +6,7 @@ export type ComponentType = {
 	index?: string | number;
 	type?:
 		/* Template hygen type */
+		| 'pagination'
 		| 'footer'
```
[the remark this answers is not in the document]
michaelmkraus: Added. `'pagination-item'` is now in the union next to `'pagination'`, consistent with the other sub-components (`accordion-item`, `navigation-item`, `tab-panel`, `footer-meta`).

packages/components/src/components/pagination/pagination.lite.tsx
```diff
@@ -0,0 +1,557 @@
...
+				const itemOptions = props.items;
+				const option = itemOptions ? itemOptions[page - 1] : undefined;
+				items.push({
+					page,
+					layout: inCollapsed ? 'always' : 'wide',
+					ellipsis: tokens.length > 0 ? tokens.join(' ') : undefined,
+					disabled: option
+						? Boolean(option.disabled) &&
+							String(option.disabled) !== 'false'
+						: false,
```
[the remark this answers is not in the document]
michaelmkraus: Fixed. `disabled=""` was treated as `false`, while an empty attribute is exactly how a web component spells true. Now uses the shared `getBoolean(option.disabled, 'disabled')`, the same call tabs makes for its own disabled. Test covers '' and 'false'.

packages/components/src/components/pagination-item/model.ts
```diff
@@ -0,0 +1,5 @@
+import type { GlobalProps, GlobalState, TextProps } from '../../shared/model';
+
+export type DBPaginationItemProps = GlobalProps & TextProps;
+
+export type DBPaginationItemState = GlobalState;
```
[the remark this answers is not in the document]
michaelmkraus: Added. `DBPaginationItemDefaultProps` and `DBPaginationItemDefaultState` follow the `drawer-footer` pattern.

packages/components/src/components/pagination/pagination.lite.tsx
```diff
@@ -0,0 +1,557 @@
...
+			<ul onClick={(event: any) => state.handleClick(event)}>
+				<li class="db-pagination-item" data-variant="ghost">
+					<button
+						class="db-pagination-previous"
+						type="button"
+						data-icon="chevron_left"
+						disabled={
+							state.isDataDriven() && state.getCurrentPage() <= 1
+								? true
+								: undefined
```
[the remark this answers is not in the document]
michaelmkraus: Fixed. The `isDataDriven()` guard is gone from the previous button, so page 1 disables it in both modes. Added test.

packages/components/src/components/pagination/pagination.lite.tsx
```diff
@@ -0,0 +1,557 @@
...
+			if (!_ref) {
+				return;
+			}
+
+			const item = _ref.querySelector('[data-page="' + page + '"]');
+			if (item && item.closest('.db-pagination') === _ref) {
+				const control = item.querySelector('a, button');
+				if (control) {
+					control.click();
+					return;
```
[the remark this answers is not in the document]
michaelmkraus: Fixed by skipping instead of disabling. `getStepPage(direction)` skips pages the items API disabled and returns `0` when none is left, driving both the `disabled` state and the click. So the arrow can no longer look enabled and do nothing. Disabling it instead would trap users on a disabled page in the middle of the list. Tests added.

nmerget (approved):

packages/components/scripts/post-build/components.ts
```diff
@@ -82,6 +82,18 @@ const controlPanelActionsComponents: Component[] = [
...
 
 export const getComponents = (): Component[] => [
+	{
+		name: 'pagination-item'
+	},
+	{
+		name: 'pagination',
+		config: {
+			react: {
+				propsPassingFilter: ['onPageChange']
```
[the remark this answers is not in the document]
mfranzke: https://github.com/db-ux-design-system/core-web/issues/8238

packages/components/src/components/pagination/examples/link.example.lite.tsx
```diff
@@ -0,0 +1,51 @@
...
+					Links come from composition: pass an anchor per item and it
+					stays a working link. The pagination numbers the items and
+					reports the page.
+				</DBInfotext>
+				<DBPagination
+					label="Linked pagination"
+					currentPage={state.linkedPage}
+					onPageChange={(page: any) => state.setLinked(page)}>
+					<DBPaginationItem>
+						<a href="#linked-page=1">1</a>
```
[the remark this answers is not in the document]
mfranzke: https://github.com/db-ux-design-system/core-web/issues/8239

PR #8034, opened 2026-09-08
feat(mcp-server): migrate to MCP SDK v2

michaelmkraus:
## Proposed changes

Migrates `@db-ux/mcp-server` to the **MCP TypeScript SDK v2** and starts serving the [2026-07-28 protocol revision](https://modelcontextprotocol.io/specification/2026-07-28/changelog). The single `@modelcontextprotocol/sdk` package is discontinued and frozen on the 2025-era protocol.

Along the way, exercising the built server against every tool surfaced three pre-existing bugs, which are fixed here as separate commits.

### 1. `refactor`: SDK v2

- `@modelcontextprotocol/sdk@1.30.0` → `@modelcontextprotocol/server@2.0.0` (runtime) plus `@modelcontextprotocol/client@2.0.0` (dev, for the stdio tests), both exact-pinned
- Tool and prompt schemas move from `zod/v3` to the zod 4 main entry, and every raw `{ field: z.string() }` shape is explicitly wrapped in `z.object()`. **Not optional**: SDK v2 dropped zod 3 support, and a raw shape built with a foreign zod fails at registration or silently breaks the first `tools/list` — a zod-3 schema typechecks fine and only fails at runtime.

### 2. `feat`: serve 2026-07-28

- `StdioServerTransport` + `server.connect()` → `serveStdio(() => buildServer())`. The server answers on **both** protocol eras: the 2025 `initialize` handshake and the 2026-07-28 `server/discover` probe.
- `serveStdio` decides the era from the opening exchange and pins one instance from the factory per connection, so the module-level `McpServer` singleton becomes a `buildServer()` factory and all 14 tools / 4 prompts move out of `index.ts` module side effects into it. Registrations outside the factory would be invisible to the served instance.
- `legacy` stays at its default (`'serve'`), so hosts that have not adopted 2026-07-28 keep working unchanged. `index.ts` keeps its handler re-exports, so the package export surface is unchanged.
- Static list results now carry cache hints, so hosts stop re-fetching them on every connection.

### 3. `fix`: `assets/` was unreachable in the published bundle

The visuals and design-token tools derived their paths with a fixed `../../assets`. That is correct from `src/tools/`, but esbuild flattens the server into `dist/index.js`, shifting `import.meta.dirname` up one level — so the path resolved to `packages/assets`, which does not exist. The published tarball has the same shape as the bundle, so it was broken there too.

Three symptoms in the artifact consumers get from `npx`, worst first:

- `get_design_tokens` **silently** served unresolved `var()` references instead of values: `var(--db-sizing-3xs)` instead of `0.25rem`, `var(--db-adaptive-bg-…)` instead of `#0d0e11`. `isError` was `false`.
- `list_visuals` reported "No visual references available" instead of the four committed visuals, also with `isError: false`.
- `get_visual_reference` failed for every name; `get_design_tokens` failed outright for `elevation`, `border` and `opacity`, which have no SCSS fallback.

Both call sites now use an `ASSETS_DIR` constant anchored on the nearest ancestor containing a `package.json` — the same directory in the sources, the bundle and the tarball, so no per-layout branching is needed.

### 4. `fix`: shadowed `error()` helper, `docs_search`, scanner report

- `handleGetExampleCode` named its catch parameter `error`, shadowing the imported `error()` helper, so the error path called the caught exception as a function and threw `TypeError: error is not a function` instead of returning a readable result.
- `docs_search` returned three arbitrary documents for any query shorter than three characters (`[].every()` is `true`), never emitted its "more results" notice (the loop breaks at 3, the check tested for more than 3), and silently searched all docs when `category: 'component'` came without a `componentName`.
- `scan_v2_migration` truncated its report mid-JSON inside a fenced block.

### 5. `ci`: the test suite never ran

Root `pnpm run test` only covers the `scripts` package, and no workflow invoked this package's tests — CI built the server but never executed it. Adds `02-mcp-server-test.yml`, wired into `default.yml` at all five required points including the `needs` list of `checks-done`; missing that one would have left the job ungated. It depends only on `init`, so it reports in ~1 minute instead of waiting ~18 for `build-outputs` artifacts it never reads.

Also groups the two `@modelcontextprotocol/*` packages for Dependabot, per the Codex review comment and the `AGENTS.md` rule for tightly coupled sets: both pin `@modelcontextprotocol/core` to an exact version, so a one-sided bump resolves two incompatible core copies into the tree.

### What we get from the new revision, and what does not apply

Useful here: cacheable list results (`ttlMs` / `cacheScope`) and deterministic tool ordering; stdio robustness (non-JSON stdout lines no longer kill the connection — exactly the failure mode `pnpm run dev` hits through tsx; bounded read buffer); in-flight handlers now abort on transport close.

Not applicable: the stateless HTTP core, session removal, header routing / `Mcp-Param-*`, multi-round-trip `input_required`, `subscriptions/listen`, the tasks extension and the whole OAuth hardening — this is a local stdio server with no state, no auth and no server→client requests. It also uses none of the newly deprecated Roots / Sampling / Logging features.

### Notable decisions and trade-offs

- **One `minor` and two `patch` changesets.** `minor` for serving the new revision, `patch` for each bug fix. `major` would be wrong: `serveStdio` keeps serving 2025-era clients.
- **This bumps the whole design system to a minor.** `@db-ux/mcp-server` sits in the changeset `fixed` group, so all 12 packages move together — see the changeset bot comment. Flagging it explicitly, since it is a design-system-wide minor driven by an MCP-server change; happy to downgrade to `patch` if preferred.
- **Bundle grows ~485 KB** (4.88 MB → 5.37 MB, +10%), from the bundled zod 4 and `@modelcontextprotocol/core`. 3.9 MB of it is the embedded `manifest.json` either way. Affects `npx` cold start.
- **No codemod.** The official `@modelcontextprotocol/codemod` would need a real devDependency (repo rules forbid `npx` / `pnpm dlx`) for three import sites. Done by hand.
- Nothing else in the monorepo imports the MCP SDK, so the blast radius is this package only.

### Verification

- **104/104 tests pass.** The suites cover both eras and both spawn targets; the modern era is pinned to `2026-07-28`, and a pin never falls back, so it fails loudly on a regression.
- **The built server was driven over stdio against its full surface**: all 14 tools, all 4 prompts and 13 error paths, in four combinations (bundle and sources × legacy and modern). All four report **49/49**. Before the asset fix the bundle failed 9 of them.
- New `bundle.test.ts` runs the built `dist/index.js` rather than the sources. This is the gap that let the asset bug through: every other suite spawns `src/index.ts` through tsx, where the old path happened to be correct, so a green suite said nothing about the shipped artifact. All four of its tests were verified to fail against the bundle before the fix; the same was done for the `error()` regression test.
- `tsc` clean. Package lint count unchanged for every modified file; the only delta is +2 in the new test file, from rule classes every test file in the package already trips (`__tests__` directory name, `process` global).
- cspell, knip, publint, prettier, markdownlint and jscpd clean.

### Follow-up notes for reviewers

A raw `server/discover` without the `io.modelcontextprotocol/protocolVersion` `_meta` envelope key is correctly answered with `-32601`, because a claim-less request counts as a 2025-era opening. This looks like a bundling bug when smoke-testing by hand and is not one — recorded in `AGENTS.md`.

Still open, deliberately out of scope: `resolveComponentPath` in `tools/components.ts` is dead code, which leaves the hardened `resolveSafePath` without a production caller — `scan_v2_migration` uses its own weaker check. Not exploitable as far as I could tell, but the protection does not sit where the input arrives. Happy to open an issue.

Not verifiable from the repo: how the real hosts (Kiro, VS Code, Claude Desktop) behave against a 2026-capable server. The default era is unchanged, so the risk is low, but a smoke test in at least one host before merging would be worthwhile.

### Code Quality

### Validation

- [ ] I have tested across all relevant frameworks (React, Angular, Vue) if applicable — n/a, this package ships no framework code

### General

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/refactor-mcp-sdk-v2>


mfranzke:
@codex review

mfranzke:
@codex review

packages/mcp-server/package.json
```diff
@@ -40,11 +40,12 @@
     "test:watch": "vitest"
   },
   "dependencies": {
-    "@modelcontextprotocol/sdk": "1.30.0",
+    "@modelcontextprotocol/server": "2.0.0",
     "zod": "4.5.4"
   },
   "devDependencies": {
     "@db-ux/db-theme-icons": "4.2.0",
+    "@modelcontextprotocol/client": "2.0.0",
```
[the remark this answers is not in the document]
michaelmkraus: Valid, and fixed in `7666b9e` — thanks.

Verified the premise before acting: `AGENTS.md:379` does require a `groups:` entry for tightly coupled sets, and `.github/dependabot.yml` had no `@modelcontextprotocol` entry. Renovate is not an alternative venue here, since `renovate.json` disables it for all packages except `pnpm` and `nvm` ("Dependabot owns everything except the exceptions below"), so `dependabot.yml` is the right file.

The coupling is in fact tighter than described: both packages pin `@modelcontextprotocol/core` to an **exact** version (`"@modelcontextprotocol/core": "2.0.0"`), so a one-sided bump does not just mix API versions — it resolves two `core` copies into the tree, and the SDK documents that brand-based `instanceof` and error identity do not cross separately bundled copies.

Grouped by pattern rather than the two literal names, so a future direct dependency on `core` or a framework adapter is covered without touching the config again, and placed ahead of the `patch-dependencies` catch-all (which previously only grouped these two on patch bumps by coincidence):

```yaml
modelcontextprotocol:
  patterns:
    - "@modelcontextprotocol*"
    - "zod"
```

`zod` was added to the group in a follow-up commit for the same reason: the SDK declares `zod: ^4.2.0` while this package pins it exactly.

One correction to the severity: `@modelcontextprotocol/client` is a devDependency and is not in the shipped bundle — verified that none of its exclusive dependencies (`pkce-challenge`, `jose`, `eventsource-parser`, `cross-spawn`) appear in `dist/index.js`. So consumers were never exposed; the blast radius was local development and CI. Worth fixing regardless, since it is an explicit repo rule.

mfranzke:
@codex review

packages/mcp-server/src/index.ts
```diff
@@ -63,90 +19,25 @@ export {
...
+// not adopted 2026-07-28 keep working unchanged.
+//
+// `onerror` is not optional in practice: it is the ONLY channel for out-of-band
+// failures. `serveStdio` reports a rejected `transport.start()` (and every later
+// transport error) through this callback and then swallows the rejection, so
+// without it a server that never came up is indistinguishable from a healthy one
+// — the v1 wiring at least surfaced it through the `unhandledRejection` handler.
+const handle = serveStdio(() => buildServer(), {
+	onerror(error) {
+		console.error('[DB UX MCP] Transport error:', error);
```
nmerget, line 39: **Bug — `serveStdio` startup failure leaves a zombie process exiting 0.** The comment above correctly identifies that `serveStdio` swallows a rejected `transport.start()`, but `onerror` here only logs. The process then stays alive with no transport and eventually exits `0` — exactly the "looks healthy while dead" failure the comment says it wants to avoid. The callback needs to terminate:

```ts
onerror(error) {
    console.error('[DB UX MCP] Transport error:', error);
    process.exit(1);
}
```

Separately, `console.error('[DB UX MCP] Server is starting on stdio...')` on line 43 prints unconditionally, including when startup has already failed synchronously.

packages/mcp-server/src/tools/scanner.ts
```diff
@@ -239,25 +295,45 @@ export async function handleScanV2Migration({
...
+	];
+
+	// Cap the findings themselves rather than the rendered report: truncating
+	// the text would cut the JSON mid-object and leave the fenced block
+	// unterminated, so the caller could not parse a single finding while the
+	// header still claimed the full count.
+	const reported = findingsWithinBudget(
+		findings,
+		MAX_JSON_OUTPUT - header.join('\n').length - REPORT_OVERHEAD
+	);
```
nmerget, line 314: **Bug — negative budget produces a nonsensical note.** `MAX_JSON_OUTPUT - header.join('\n').length - REPORT_OVERHEAD` is not clamped, so a long header can make the budget negative. `findingsWithinBudget` then returns `[]`, and with `omitted > 0` the report reads "Listing the first **0** findings (sorted by line number); 137 more were omitted … Migrate these first." — a message telling the caller to migrate zero findings.

Clamp the budget with `Math.max(0, …)` and give `reported.length === 0` its own message (e.g. "The report is too large to include any single finding; …").

packages/mcp-server/src/tools/scanner.ts
```diff
@@ -239,25 +295,45 @@ export async function handleScanV2Migration({
 		)
 	];
 
-	const summary = [
+	const header = [
```
nmerget, line 298: **Bug (both reviewers) — the total response size is no longer bounded.** Dropping `truncate(summary, MAX_JSON_OUTPUT)` (see line ~333) means only the findings *array* is budgeted; the header is subtracted from the findings budget but is itself never capped. `uniqueComponents.join(', ')` here is unbounded, and combined with a 500-char `filePath` the header alone can exceed `MAX_JSON_OUTPUT`, producing a response large enough to trip MCP/client message limits.

Either cap the rendered component list (with an omitted-count indicator), or assert/clamp the final rendered length before returning. This was the one finding both review models raised independently, so it's the highest-confidence item.

packages/mcp-server/src/utils/path.ts
```diff
@@ -20,8 +67,17 @@ function normalize(p: string): string {
...
-		decoded = decodeURIComponent(decoded);
+	try {
+		while (decoded !== decodeURIComponent(decoded)) {
+			decoded = decodeURIComponent(decoded);
+		}
+	} catch {
+		// A malformed escape sequence makes decodeURIComponent throw — a literal
+		// '%' in a real filename does exactly that. Treating it as a traversal
+		// attempt would reject legitimate paths, so the raw input is used and
+		// still has to pass the containment check below. An undecoded string can
```
nmerget, line 78: **Concern — percent-decoding is wrong for real filesystem paths.** `scan_v2_migration` feeds real host paths through `resolveSafePath`, which now decodes them. At the filesystem layer `%2E%2E%2F` is *not* traversal — `resolve()` treats it as a literal filename inside cwd, so the old `startsWith` check already contained it (it failed with "not found" rather than "Path traversal"). So the decode adds no protection here, but it *does* break legitimate names: `report%20final.tsx` is silently rewritten to `report final.tsx` and reported missing.

The new test only covers the `%`-that-fails-to-decode case, not the `%NN`-that-decodes-to-a-different-valid-name case. Consider skipping the decode for the scanner path, or restricting decoding to the component-name callers it was originally written for.

packages/mcp-server/src/tools/docs.ts
```diff
@@ -77,16 +90,40 @@ export async function handleDocsSearch({
 	componentName?: string;
 	docType?: string;
 }): Promise<ToolResult> {
+	if (
+		category === 'component' &&
+		(componentName === undefined || componentName === '')
+	) {
+		return error(
+			"Error: 'componentName' is required when category is 'component'. Pass the component name (e.g. 'button'), or use category 'global' to search the foundation docs."
+		);
```
nmerget, line 99: **Design — enforce this constraint in the schema, not the handler.** `componentName` is declared `.optional()` in `docsSearchSchema`, but the handler rejects `category: 'component'` without it. Hosts see a schema that permits the call and only learn otherwise from an `isError` result at call time. Model it in zod so the SDK's validator rejects it and the emitted JSON Schema communicates it:

```ts
z.object({ /* … */ }).refine(
    (v) => v.category !== 'component' || Boolean(v.componentName),
    { message: "'componentName' is required when category is 'component'" }
)
```

The runtime guard can stay as a defensive fallback, but the schema should be the source of truth.

packages/mcp-server/src/tools/schemas.ts
```diff
@@ -58,13 +58,13 @@ export const getExampleCodeSchema = {
 			.describe(
 				"Target framework: 'react', 'angular', 'vue', 'web-components', or 'html'"
 			)
-	}
+	})
```
nmerget, line 61: **Consistency — `get_example_code` still advertises `html`, which always errors.** This PR introduces `ExampleCodeFramework` (in `types.ts`) precisely to encode "frameworks that have generated examples" and narrows `FRAMEWORK_EXT` to the four supported values. But the tool's input `framework` enum still offers `html` and the handler returns an error for it unconditionally. Narrow the enum here to the four supported values (`react`, `angular`, `vue`, `web-components`) and move the "for HTML, see docs/HTML.md" pointer into the tool description, so hosts don't offer a value that can only fail.

packages/mcp-server/src/utils/path.ts
```diff
@@ -8,6 +9,52 @@ function normalize(p: string): string {
...
+
+/**
+ Absolute path to the package's `assets/` directory.
+
+ Runtime asset reads must go through this constant. Deriving the path with a
+ fixed `../../assets` is correct for the sources but points one level outside
+ the package once bundled — which made the visuals tools report an empty
+ directory and the design-token tool silently fall back to raw SCSS.
+ */
+export const ASSETS_DIR = join(findPackageRoot(), 'assets');
```
nmerget, line 56: **Concern — `ASSETS_DIR` does synchronous I/O and can throw at import time.** `findPackageRoot()` runs during module evaluation of `utils/path.ts`, which is transitively imported by everything — including the publicly exported `resolveSafePath`. A broken layout throws before any handler can turn it into a readable tool result, and any library consumer that imports the package pays 1-8 `existsSync` calls just on import. A memoized lazy getter gives the same fail-fast behavior at first use without the import-time side effect:

```ts
let _assetsDir: string | undefined;
export function getAssetsDir(): string {
    return (_assetsDir ??= join(findPackageRoot(), 'assets'));
}
```

packages/mcp-server/src/__tests__/e2e.test.ts
```diff
@@ -76,3 +195,85 @@ describe('MCP server — stdio transport', () => {
...
+		const schemas = Object.fromEntries(
+			tools
+				.toSorted((a, b) => a.name.localeCompare(b.name))
+				.map((tool) => [tool.name, tool.inputSchema])
+		);
+
+		expect(schemas).toMatchSnapshot();
+	}, 10_000);
+});
+
```
nmerget, line 245: **Test — snapshot comment claims coverage it doesn't have.** The comment says the snapshot guards against `additionalProperties` changes, but the generated schemas in the `.snap` contain no `additionalProperties` key at all — so a change there can't be caught by this snapshot. Either drop that clause from the comment, or add `.strict()` to the zod objects (emitting `additionalProperties: false`) so unknown keys are actually rejected and the guard becomes real.

package.json
```diff
@@ -70,7 +70,7 @@
     "release:publish": "node scripts/github/publish-npm.js",
     "rm:builds": "pnpm --filter scripts run rm:builds",
     "start": "pnpm --filter patternhub run start",
-    "test": "pnpm --filter scripts run test",
+    "test": "pnpm --filter scripts --filter @db-ux/mcp-server run test",
```
nmerget, line 73: **Docs drift — root `pnpm run test` now triggers a build.** `@db-ux/mcp-server`'s `test` script is `pnpm run build && vitest run`, so the root `test` script now runs `prebuild.ts` (writing `assets/migration/`, `assets/tokens/tokens.json`, `src/manifest.json`) plus esbuild. The root `AGENTS.md` still documents `pnpm run test` as "~10 seconds" and lists it among read-only pre-commit validation steps. Either update that documentation, or point the root test at a `vitest run`-only script for mcp-server with the build handled separately by `pnpm run build`.

.github/workflows/default.yml
```diff
@@ -109,6 +109,12 @@ jobs:
     needs: [build-outputs]
     secrets: inherit
 
+  test-mcp-server:
+    if: ${{ needs.init.outputs.isChangesetRelease != 'true' }}
+    uses: ./.github/workflows/02-mcp-server-test.yml
+    needs: [init]
+    secrets: inherit
```
nmerget, line 116: **CI — confirm fork-PR behavior before merge.** `test-mcp-server` gates only on `isChangesetRelease`, unlike the sibling `test-figma-code-connect` job right below, which also excludes forks (`needs.init.outputs.isFork != 'true'`). This job is the first to hard-require the decrypted `@db-ux/db-theme` SCSS (`scripts/build-tokens.ts` throws without it), and the `pnpm-cache` action runs db-theme's postinstall decrypt with empty `ASSET_*` secrets on forks. It likely works (the file appears shipped rather than decrypted), but I couldn't verify fork behavior locally — and since this job is now in the `checks-done` `needs` list, a failure here blocks the gate for every fork PR. Worth a quick check on a fork PR before merging.

.github/dependabot.yml
```diff
@@ -103,6 +103,18 @@ updates:
...
+      # MCP SDK: all v2 packages share one version number and pin
+      # @modelcontextprotocol/core to an exact version, so a one-sided bump
+      # resolves two incompatible core copies into the tree.
+      #
+      # zod belongs in the same group: the SDK declares "zod": "^4.2.0" while
+      # @db-ux/mcp-server pins it exactly. Bumping only one side lets the
+      # resolver put a second zod copy into the bundle, and the tool schemas
+      # must come from the same copy the SDK converts them with.
+      modelcontextprotocol:
+        patterns:
```
nmerget, line 115: **Nit — the "single zod copy" rationale is weaker than the comment implies.** Grouping `zod` with `@modelcontextprotocol*` is the right call, but note the SDK declares `zod` as a regular dependency (`^4.2.0`), not a peer. A zod release newer than this package's exact pin will resolve a *second* copy regardless of grouping — the group only keeps the two bumps in the same PR, it doesn't guarantee dedup. If a single bundled zod copy is a hard requirement (the AGENTS.md note suggests it is, for schema conversion), add a `pnpm.overrides` entry for `zod` rather than relying on the Dependabot group.

packages/mcp-server/src/tools/components.ts
```diff
@@ -181,12 +158,16 @@ export async function handleGetExampleCode({
...
+				// Looked up after the guard above, so `framework` is narrowed to
+				// the frameworks that actually have generated examples.
+				const ext = FRAMEWORK_EXT[framework];
 				const fwExamples = comp.exampleCode[framework] ?? {};
 				const directKey = `${kebab}.example.${ext}`;
 				const matchKey = fwExamples[directKey]
 					? directKey
 					: fuzzyMatchExample(Object.keys(fwExamples), kebab, ext);
-				if (!matchKey) {
+				const code = matchKey ? fwExamples[matchKey] : undefined;
```
nmerget, line 169: **Follow-through (relates to `schemas.ts`).** Now that `framework` is narrowed to `ExampleCodeFramework` for the lookup, the `html` branch here can only ever hit the "not found" error path. If the input enum in `schemas.ts` is narrowed to the four supported frameworks (see the comment there), this handler no longer has to defend against `html`/`vanilla` at all — worth aligning the two so the type, the schema, and the handler all agree on the same four values.

nmerget (commented):
## God review — MCP SDK v2 migration

This review synthesizes two independent AI reviews (Claude Opus via Kiro CLI + Codex/gpt-5.6 via Copilot CLI) of the `refactor-mcp-sdk-v2` branch. Kiro CLI verified locally: `tsc --noEmit` clean, 104/104 tests pass, cspell/prettier/markdownlint/knip clean on changed files. Individual line comments are attached to the relevant code.

### Highest confidence (raised by both models)
- **Scanner response size is no longer bounded** (`scanner.ts`) — dropping `truncate(summary, MAX_JSON_OUTPUT)` budgets only the findings array; the header (`filePath` + unbounded `uniqueComponents`) can exceed the limit on its own. This is the one finding both models raised independently.

### Correctness / concerns (single source, but concrete)
- `serveStdio` startup failure leaves a zombie process exiting 0 — `onerror` only logs (`index.ts`).
- Negative report budget yields "Listing the first 0 findings … Migrate these first" (`scanner.ts`).
- Percent-decoding in `resolveSafePath` breaks legitimate `%NN` filenames for the scanner and adds no real traversal protection there (`path.ts`); the new test doesn't cover this case.
- `docs_search` / `get_example_code` enforce constraints in the handler that belong in the zod schema (`docs.ts`, `schemas.ts`).
- `ASSETS_DIR` runs synchronous I/O and can throw at import time (`path.ts`).

### Docs / follow-through (no specific diff line)
- **Root `AGENTS.md` not updated for the new CI job / test timing.** The "Always Run These Commands Before Committing" list and the "~10 seconds" note for `pnpm run test` are now inaccurate (root test now builds mcp-server). The repo's own rule is to keep these aligned in the same commit.
- **Snapshot comment** in `e2e.test.ts` claims `additionalProperties` coverage the schemas don't have (see inline).

### Lower priority
- **Duplicated test helpers** — `text()`/`textOf()` and the content-block extraction exist in both `bundle.test.ts` and `e2e.test.ts`, and `e2e.test.ts` already spawns the bundle (`SPAWN_BUNDLE`), overlapping with `bundle.test.ts`'s "shipped bundle" describe block. Consider a shared `__tests__/helpers.ts` and a single bundle suite.
- **Package `lint` script is effectively dead** — `pnpm --filter @db-ux/mcp-server run lint` reports ~300+ errors, almost all pre-existing in untouched files (`n/file-extension-in-import`, `unicorn/single-line-block-comment-style`). Not introduced here, and root `lint:xo` is what CI runs — but since this PR adds a CI job for the package, a good moment to fix or remove the script so it stops giving a false signal.

No conflicts between the two reviews; Codex's sole surfaced finding is a subset of the scanner-bounding issue. None of these are blockers on their own, but the scanner-bounding and `onerror` items are worth addressing before merge.

packages/mcp-server/src/tools/scanner.ts
michaelmkraus: Fixed. The unique-component list is capped at 2000 chars with an indicator, so the header can no longer grow past the output limit on its own.

packages/mcp-server/src/index.ts
michaelmkraus: Fixed, but not with an unconditional exit. `onerror` also fires for probe-discard timeouts, late instance disposal, 2025-era requests on modern-only connections and every live-session `wire.onerror`. Exiting on those would undo the v2 guarantee that one bad line doesn't kill the connection. The factory is the discriminator: it only runs after the first inbound message, and `wire.start()` rejects before that. So we exit `1` only while no instance has been built. Note this wasn't a regression — `registerLifecycleHandlers` only logs too.

packages/mcp-server/src/utils/path.ts
michaelmkraus: Agreed, fixed. The decoded form is now only validated, never resolved. `resolveSafePath` returns the raw path. `%2E%2E%2F…` is still rejected, and `report%20final.tsx` addresses `report%20final.tsx` again. Added the test case that was missing.

packages/mcp-server/src/utils/path.ts
michaelmkraus: Fixed as suggested. `tokens.ts` and `visuals.ts` derived their paths at module scope from it, so those are lazy now too.

packages/mcp-server/src/tools/schemas.ts
michaelmkraus: Fixed. Enum narrowed to the four `ExampleCodeFramework` values, HTML pointer moved into the tool description, snapshot updated. The handler guard stays with a comment: `handleGetExampleCode` is exported from the package root, so a library caller can still pass `html` / `vanilla`.

package.json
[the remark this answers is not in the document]
michaelmkraus: Confirmed and fixed. No workspace dep between scripts and `@db-ux/mcp-server` and no workspace-concurrency, so both really ran in parallel over the same build output.

packages/mcp-server/src/__tests__/e2e.test.ts
michaelmkraus: You're right. I corrected the comment rather than the schemas.

packages/mcp-server/src/tools/docs.ts
michaelmkraus: Checked this against zod and the premise doesn't hold: `z.toJSONSchema()` returns a byte-identical schema with and without the `.refine()`, so it would not communicate the constraint to hosts. It would only make the SDK reject the call at validation time, replacing a readable `isError` result that a model can act on with an opaque protocol error. Leaving the handler guard as the source of truth.

.github/workflows/default.yml
michaelmkraus: Checked, left the gating as is. `_default_variables.scss` sits under `build/`, which is in db-theme's `files` — it ships in the tarball and is never decrypted; the postinstall decrypt only covers fonts and images, and AGENTS.md documents its failure with placeholder credentials as expected and non-fatal. `build-packages` is also gated only on `isChangesetRelease`, runs on forks and compiles foundations SCSS against db-theme. Adding `isFork != 'true'` would disable the suite for fork PRs, which seems worse than the risk.

.github/dependabot.yml
michaelmkraus: Right that `zod` is a regular dependency, not a peer. But the second copy doesn't follow while the ranges overlap — pnpm dedupes. This just played out for real: main bumped zod to 4.6.5 independently, and after the merge @modelcontextprotocol/server@2.0.0 resolves to that same 4.6.5. One copy. The three zod entries in the lock belong to the legacy transitive @modelcontextprotocol/sdk@1.30.0 in the Vite-devtools tree. A repo-wide override would force a version onto that tree too, so I'd rather keep the Dependabot group as the mechanism.

packages/mcp-server/src/tools/scanner.ts
michaelmkraus: Fixed — the unique-component list is capped at 2000 chars, so the header stays within the output limit while the counts stay complete.

packages/mcp-server/src/tools/scanner.ts
```diff
@@ -239,25 +295,45 @@ export async function handleScanV2Migration({
...
+		JSON.stringify(reported, null, 2),
 		'```'
 	].join('\n');
 
 	return {
 		content: [
 			{
 				type: 'text',
-				text: truncate(summary, MAX_JSON_OUTPUT)
+				text: summary
```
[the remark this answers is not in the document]
michaelmkraus: Fixed — the unique-component list is now capped at 2000 chars, so the header can no longer exceed the output limit on its own while the full counts stay intact.

package.json
michaelmkraus: Fixed. `mcp-server`'s test is now `vitest run` only, so root `pnpm run test` is read-only again and the AGENTS.md "~10 seconds" note holds. The build is handled separately by root `pnpm run build` and an explicit build step in the CI job.

packages/mcp-server/src/tools/components.ts
michaelmkraus: The schema enum is now narrowed to the four supported frameworks, so hosts can't reach this branch. Kept the guard because `handleGetExampleCode` is exported from the package root — a library caller can still pass `html`/`vanilla`, so it's the defensive path for that, not dead code.

.github/workflows/02-mcp-server-test.yml
nmerget, line 1: Can we move this into `.github/workflows/01-build-packages.yml` after the vite-plugin tests? or into 01-validate

michaelmkraus: Done. Moved the mcp-server tests into `01-build-packages.yml` as a step after the vite-plugin tests (`pnpm run build` there already produces `dist`, so no extra build needed) and removed `02-mcp-server-test.yml` plus its wiring in `default.yml`.

nmerget (approved):

PR #8040, opened 2026-09-08
fix(DBControlPanel): open drilldown with a transition instead of a clip-path

nmerget:
## Proposed changes

Reworks the control panel navigation drilldown so it reveals via absolute positioning and a transition instead of a `clip-path`. This lets the menu animate open correctly. Related control-panel and navigation-item styles were adjusted to support the new positioning (e.g. `position: relative` on the scroll container, and wrapping indicators in the `navigation-indicator` mixin).

**What changed and why:**
- `control-panel-navigation-item-group-menu-drilldown.scss` — drilldown now uses `position: absolute; inset: 0; overflow: hidden auto` on the base selector instead of `fixed`/`clip-path`-driven insets per breakpoint.
- `_control-panel-desktop.scss` — added `position: relative` to the scroll container and wrapped indicators in the `navigation-indicator` mixin.
- Supporting tweaks in `control-panel-mobile.scss`, `control-panel-navigation.scss`, `shell-sub-navigation.scss`, and `_navigation-item.scss`.

No public API change — this is a styling bug fix, so a `patch` changeset is included for all component packages.

resolves # (issue number)

### Code Quality

### Validation

### General

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/fix-drilldown-transition>


mfranzke:
@codex review

packages/components/src/components/shell-sub-navigation/shell-sub-navigation.scss
```diff
@@ -86,7 +87,7 @@
 
 			/* jscpd:ignore-end */
 
-			.db-control-panel-navigation {
+			*:not(.db-shell-sub-navigation-button) {
```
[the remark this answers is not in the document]
nmerget: Good catch. Fixed in 727b9159f10e — the collapsed hide rule now targets `.db-control-panel-navigation, > :not(.db-shell-sub-navigation-button)` instead of the bare descendant `*:not(...)`, so it hides the navigation subtree and non-toggle direct children while leaving the toggle container's inner `.db-button` visible.

packages/components/src/styles/internal/_navigation-item.scss
```diff
@@ -85,10 +92,13 @@
 %navigation-item {
 	@extend %db-overwrite-font-size-md;
 
-	position: relative;
 	display: flex;
 	block-size: fit-content;
 
+	@include navigation-indicator {
+		position: relative;
+	}
```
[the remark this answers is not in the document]
nmerget: Fixed in 727b9159f10e. Restored a positioning context by adding `position: relative` to `.db-control-panel-navigation-item` so its absolutely positioned `.db-control-panel-navigation-item-end-slot-container` resolves against the item again. It is added on the item only — item groups are intentionally left unpositioned so the drilldown overlay still anchors to the control-panel scroll container rather than the group.

packages/components/src/components/control-panel-navigation-item-group/control-panel-navigation-item-group-menu-drilldown.scss
```diff
@@ -6,14 +6,13 @@
...
-$drilldown-inset: calc(
-	#{variables.$db-sizing-md} + #{variables.$db-spacing-fixed-sm} * 2
-);
-
 %db-control-panel-navigation-item-group-menu-drilldown {
 	@extend %db-control-panel-navigation-item-group-menu-non-tree;
 
+	overflow: hidden auto;
+	position: absolute;
+	inset: 0;
```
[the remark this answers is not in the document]
nmerget: Not applicable here. `DBControlPanelNavigation` is a sub-component that is always rendered inside a `ControlPanel` wrapper (the desktop `.db-control-panel-desktop-scroll-container` or the mobile `.db-control-panel-mobile-drawer-scroll-container`), and both of those already establish `position: relative` as the overlay bounds. There is no supported standalone usage where the drilldown menu would resolve against the initial containing block, so no component-local fallback is needed.

packages/components/src/styles/internal/_navigation-item.scss
```diff
@@ -71,7 +78,7 @@
 	text-align: start;
 	align-items: center; // Centering the content vertically and horizontally
 	min-block-size: variables.$db-sizing-md;
-	gap: variables.$db-spacing-fixed-sm;
+	gap: variables.$db-spacing-fixed-xs;
```
michaelmkraus, line 81: This gap change applies to every control panel navigation item and expand button in all variants, not just the drilldown. None of the three changesets mentions it — consumers get a layout shift without a changelog entry.

packages/components/src/styles/internal/_navigation-item.scss
```diff
@@ -16,6 +16,13 @@
 	}
 }
 
+@mixin navigation-indicator {
+	> a,
+	> .db-control-panel-navigation-item-group-expand-button {
+		@content;
+	}
+}
```
michaelmkraus, line 24: Moving the indicator to `> a / > .db-control-panel-navigation-item-group-expand-button` orphans the overrides in `control-panel-navigation-item-group.scss` (lines 62–66): `--indicator-vertical-block-size` and `--indicator-vertical-inset-block-start` are declared on `.db-control-panel-navigation-item-group::after`, and custom properties on a pseudo-element are not inherited by a child's pseudo-element. Before, `direct-navigation-items` resolved to the same element, so they applied. The new behaviour is probably more correct, but those two declarations plus the "avoid issues with tree navigation" comment are now dead.

packages/components/src/components/control-panel-navigation-item-group/control-panel-navigation-item-group-menu-drilldown.scss
michaelmkraus, line 14: The nearest positioned ancestor here is a scroll container (`.db-control-panel-desktop-scroll-container`, `.db-control-panel-mobile-drawer-scroll-container`), so `inset: 0` resolves against its padding box and the overlay scrolls with the content. With a navigation long enough to scroll, opening a group after scrolling places the overlay `scrollTop` px too high and leaves a gap at the bottom. The old `position: fixed` was viewport-anchored. Can the overlay anchor to a non-scrolling wrapper instead?

packages/components/src/components/control-panel-navigation/control-panel-navigation.lite.tsx
```diff
@@ -704,6 +712,17 @@ export default function DBControlPanelNavigation(
...
+				// Also observe the nav root. Its box reflows across the
+				// mobile/desktop breakpoint while menuRef may not, so this
+				// guarantees the variant (popover vs drilldown) is recomputed
+				// on viewport changes instead of sticking to the previous one.
+				if (!state._rootResizeObserverCallbackId) {
+					state._rootResizeObserverCallbackId =
+						new ResizeObserverListener().observe(_ref, () => {
+							state._update();
+						});
+				}
```
michaelmkraus: `menuRef` and `_ref` resize together, so `_update()` now runs twice per resize tick. Either observe only the root or share one callback.

michaelmkraus:
`resolves # (issue number)` is still a placeholder, no issue is linked

mfranzke:
@codex review

packages/components/src/styles/internal/_navigation-item.scss
nmerget: Good catch, fixed in f526226746f5. The two overrides (`--indicator-vertical-block-size` and `--indicator-vertical-inset-block-start`) plus the "avoid issues with tree navigation" comment are now declared inside the `navigation-indicator` mixin scope (`> a` / `> .db-control-panel-navigation-item-group-expand-button`) instead of on `.db-control-panel-navigation-item-group::after`, so they land on the same element that now carries the indicator `::after` and are read again.

packages/components/src/components/control-panel-navigation/control-panel-navigation.lite.tsx
nmerget: Fixed in f526226746f5. Consolidated to a single ResizeObserver on the root (`_ref`). The `menuRef` observer is removed since `menuRef` and `_ref` resize together and both callbacks only called `state._update()`, so `_update()` no longer runs twice per resize tick. The root is kept as the sole target because its box reflows across the mobile/desktop breakpoint (menuRef may not), so it still covers the window-resize case. The now-unused `_rootResizeObserverCallbackId` state field was dropped too.

packages/components/src/components/control-panel-navigation-item-group/control-panel-navigation-item-group-menu-drilldown.scss
nmerget: This is already handled by the overflow removal in `%drilldown-navigation`: when a group is expanded the scroll owner gets `overflow: hidden`, which removes its scroll range and collapses any existing scroll offset back to the top, so the `inset: 0` overlay realigns with the top of the viewport. No extra anchoring is needed.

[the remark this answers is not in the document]
nmerget: Same as the thread above: the `overflow: hidden` applied while a drilldown level is open removes the scroll range and resets the offset to the top, so the overlay is no longer shifted up by a pre-existing `scrollTop`.

packages/components/src/styles/internal/_navigation-item.scss
nmerget: Added a changelog entry for this in f526226746f5. The gap change is now documented in the consolidated `drilldown-transition.md` changeset ("reduce the navigation item and expand button gap to fix wrong icon spacing").

packages/components/src/components/control-panel-navigation-item-group/control-panel-navigation-item-group-menu-drilldown.scss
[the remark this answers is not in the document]
nmerget: Fixed in bd09fdd5c96f.

You were right that the previous `overflow: hidden` on `.db-control-panel-navigation` did not clear the ancestor scroll offset, because that element is not always the scroll owner. There are actually three different scroll owners depending on layout:

- desktop main panel -> `.db-control-panel-desktop-scroll-container`
- mobile drawer -> `.db-control-panel-mobile-drawer-scroll-container`
- left sub-navigation -> `.db-control-panel-navigation` itself (it carries `overflow: hidden auto; position: relative`)

The lock is now centralized in a `lock-scroll-while-drilldown-open` mixin and applied on each of those scroll owners (guarded by `:has(.db-control-panel-navigation[data-variant="drilldown"])`). Locking `overflow: hidden` on the actual scroll owner while a level is open removes its scroll range, collapses any existing `scrollTop` back to the top, and realigns the `inset: 0` overlay with the visible scrollport - so it is no longer shifted up by a pre-existing offset, even for a level that has no focusable link to trigger a browser scroll.

Verified live in the react-showcase (mobile drilldown scrolled to `scrollTop=259`, desktop-left, and the left sub-navigation at a short viewport): opening a group sets the owner's `overflow-y: hidden`, the overlay pins to the top with the back button visible, and only the overlay shows a scrollbar (no second stacked scrollbar).

packages/components/src/components/control-panel-navigation-item-group/control-panel-navigation-item-group.lite.tsx
```diff
@@ -441,6 +439,8 @@ export default function DBControlPanelNavigationItemGroup(
 					state.isSubNavigationExpanded
 				)}
 				class="db-control-panel-navigation-item-group-expand-button"
+				data-icon={props.icon}
+				data-show-icon={getBooleanAsString(props.showIcon, 'showIcon')}
```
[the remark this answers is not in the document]
nmerget: Addressed in bd09fdd5c96f.

The group does not need `icon-passing` at all, so there is no CSS-only rewrite required. Unlike `.db-control-panel-navigation-item`, which wraps a consumer-provided `<a>` and therefore needs the absolute icon + child padding that `icon-passing` provides, the group owns its expand-button markup entirely. `data-icon` on the button renders a leading glyph via the global `[data-icon]::before` rule, and the button is a flex row with `gap`, so the glyph sits before the text with correct spacing - no absolute positioning or self-padding needed. I removed the now-unused `icon-passing` import from the group stylesheet; the mixin itself is untouched and still used by the navigation item.

Since the component always emits `data-icon` on the button (the documented markup moved with it), the glyph is on the button the consumer clicks, so the "clicking the glyph does not toggle" concern does not apply to the current output.

packages/components/src/styles/internal/_navigation-item.scss
```diff
@@ -16,6 +16,75 @@
...
+			transition: transform
+				#{variables.$db-transition-duration-extra-slow}
+				#{variables.$db-transition-timing-emotional};
+		}
+	}
+
+	&:has(
+		.db-control-panel-navigation-item-group-expand-button[aria-expanded="true"]
+	) {
+		overflow: hidden;
```
michaelmkraus, line 40: `overflow: hidden` does not reset the scroll offset — the element stays a scroll container, only user scrolling is disabled. What clamps `scrollTop` is the content shrinking via `block-size: 0` (line 62), and that is delayed by `$db-transition-duration-extra-slow` (line 77). So for the full slide the old content height is still there, `scrollTop` is unchanged, and the `inset: 0` overlay sits `scrollTop` px above the visible area before snapping into place.

Please measure during the animation, not at the end state: set `scrollTop`, open a group,
read the overlay `boundingBox()` after ~200 ms.

packages/components/src/components/control-panel-navigation/control-panel-navigation.scss
```diff
@@ -10,40 +10,26 @@
 @use "../control-panel-navigation-item-group/control-panel-navigation-item-group-menu-drilldown";
 
 %drilldown-navigation {
```
michaelmkraus, line 12: The "lock intermediate drilldown menus" rule that lived here is gone and not reproduced in `lock-scroll-while-drilldown-open` — that mixin only sets `overflow: hidden` on the scroll owner and on `.db-control-panel-navigation`, never on `.db-control-panel-navigation-item-group-menu`.

Intermediate overlays therefore keep their own `overflow: hidden auto`. Opening level 3 from a scrolled level 2 reproduces the offset one level down, where the scroll-owner lock cannot reach because the containing block is the scrolled level-2 overlay.

packages/components/src/components/control-panel-navigation-item-group/control-panel-navigation-item-group-menu-drilldown.scss
```diff
@@ -61,7 +47,5 @@ $drilldown-inset: calc(
 		); // Remove the border from fixed menu
 
 		padding: variables.$db-spacing-fixed-sm;
```
michaelmkraus, line 49: The vertical desktop navigation uses `padding-inline: calc($db-spacing-fixed-sm + env(safe-area-inset-left))`, this overlay only `$db-spacing-fixed-sm`. On devices with safe-area insets the overlay content is horizontally offset from the level behind it during the slide.

PR #8057, opened 2026-09-09
fix(mcp-server): docs_search matching and scan_v2_migration report

michaelmkraus:
## Proposed changes

Fixes five pre-existing defects in `@db-ux/mcp-server` that surfaced while reviewing #8034 (MCP SDK v2 migration). They are deliberately **not** part of that PR: they are unrelated to the protocol change, and its description already listed them under "Known gaps, deliberately left out of this PR".

### `docs_search`

- **A short query returned arbitrary results.** Terms shorter than three characters are discarded before matching, and an empty term list makes `every()` vacuously true — so `'db v3'` matched _every_ document and the caller got three arbitrary docs presented as search hits. Such a query is now rejected with an explanation. An explicitly empty query still lists the docs in scope, which the schema documents as the way to read a specific doc.
- **`category: 'component'` without `componentName` searched everything.** The scope filter was guarded by `componentName &&`, so a missing name skipped it entirely although the schema calls the name required. The name is enforced now, and the filter no longer short-circuits — if it were ever absent, the filter excludes everything, which fails visibly instead of silently widening the search.
- **The "more than 3 results" notice was unreachable.** The search loop broke at exactly three results while the notice tested for more than three, so callers were never told that matches had been withheld. The loop now collects one result beyond the limit as a lookahead, and both places share a single constant.

### `scan_v2_migration`

- **A long report was unparsable.** The whole rendered report was truncated at 20 000 characters, which cut the JSON mid-object and left the fenced block unterminated — no finding could be parsed while the header still claimed the full count. The findings themselves are capped now, sized against the actual header length, and the report states how many were omitted so the caller knows to scan again.
- **Path handling moved to the hardened helper.** The tool used an ad-hoc `startsWith(cwd + '/')` check, which left `resolveSafePath` without any production caller and did not collapse percent-encoding. It now uses `resolveSafePath`, so percent-encoded traversal is rejected. That helper in turn no longer throws on a legitimate filename containing a literal `%`, where `decodeURIComponent` fails — previously that would have been reported as a traversal attempt.

### Type safety and dead code

Clears the four pre-existing `tsc` errors in the package (now **0**):

- `FRAMEWORK_EXT` was typed `Partial<Record<Framework, string>>`, making every lookup `string | undefined`. It is keyed by the new `ExampleCodeFramework` (`Framework` minus `html`/`vanilla`) and read after the html/vanilla guard, where `framework` is narrowed.
- The same type makes the `as Manifest` cast in `utils/manifest.ts` unnecessary — the manifest no longer claims `exampleCode` entries for `html` and `vanilla`, which the build never produced.
- The `node:child_process` mock declares `Error | null`, matching what Node actually passes on success, rather than swapping it for `undefined`.

Removes dead code: `resolveComponentPath` (whose absence left `resolveSafePath` unused in production) and `FRAMEWORK_OUTPUT_DIR`.

### Verification

- 91/91 tests pass, including **eight new regression tests**. Five of them were verified to fail against the unfixed sources; the other three pin behaviour that has to keep working (an explicitly empty query, component scoping, a filename with a literal `%`).
- `tsc`: 4 pre-existing errors → 0.
- ESLint: 342 → 345. The three added are `no-unsafe-argument` from the `resetManifestCache(JSON.parse(...))` pattern this test file already uses 36 times.
- prettier, markdownlint and cspell clean.

### Note on merge order

`packages/mcp-server/src/tools/components.ts` is also touched by #8034 (the shadowed `error()` helper). The two changes sit in the same function but on different lines, so expect a small conflict for whichever PR merges second.

### Code Quality

### Validation

- [ ] I have tested across all relevant frameworks (React, Angular, Vue) if applicable — n/a, this package ships no framework code

### General

- [ ] If architecture, structure, or conventions changed inside a `packages/*` folder, the corresponding `AGENTS.md` has been updated — n/a, no convention changed


mfranzke:
@codex review

Issue #6092, opened 2026-02-19
Cursor jumps to beginning and input is cleared when entering decimal separator in number field


Pi-42:
### Which generators are affected?

- [ ] All
- [x] HTML
- [ ] React
- [x] Angular
- [ ] Vue
- [ ] Web components
- [ ] Power Apps

### Reproduction case

There is an issue with the input field configured with type="number".

Steps to Reproduce:
1. Focus on the number input field.
2. Enter the digit 1
3. Type a comma ,

Actual Result:
- The cursor jumps to the beginning of the field (before the 1).- 
- The comma is not displayed.

4. Instead of a comma, type a dot . after entering 1

Actual Result:
- The 1 is removed.
- The field becomes empty.

### Expected Behaviour

Expected Result:
- The input field should allow valid decimal separators according to the expected format.
- The cursor position should remain correct.
- Existing input should not be deleted unexpectedly.

### Browser version

None

### Add any other context about the problem here.

Link to Stackblitz https://stackblitz.com/edit/db-ux-design-system-examples-fbnkszru?file=src%2Fapp%2Fapp.component.html

### Which DB business unit do you work for?

DB InfraGO AG

PR #8058, opened 2026-09-09
fix(DBInput): keep a partially typed date that the browser cannot parse

### LINKED ISSUE #7748: Date Input resets on invalid partial entry

- [ ] All
- [ ] HTML
- [ ] React
- [x] Angular
- [ ] Vue
- [ ] Web components
- [ ] Power Apps

### Reproduction case

Visit
https://design-system.deutschebahn.com/core-web/review/main/angular-storybook/?path=/story/components-dbinput-example-types-with-min-and-max--date
and try to enter the date 29.02.2028 manually.
During typing, when the input should read 29.02.0202 (the final 8 not typed yet), the input field resets and is empty again.

Using the Date Picker works for this date.
React and Vue are not affected.

### Expected Behaviour

While typing, the input field goes invalid for 29.02.0202, but the user can still type the last digit (8) to complete the date.

### Screenshots

Expected behaviour (Screenshot taken from React):
Image

### Browser version

None

_No response_

DB Energie

_No response_

MaZuMo

michaelmkraus:
## Proposed changes

Typing a date manually into `DBInput` in Angular wiped the field as soon as the entry passed through a value the browser cannot parse.

**Reproduction (the issue case, `29.02.2028`):** the year segment runs through `0020` — a leap year, so the browser reports the complete date `0020-02-29` — and then through `0202`, which does not exist. At that second step the browser sets `value` to `""`, sets `validity.badInput` and fires `input` + `change`, while the editor still shows everything typed. Verified in Chromium:

```text
typed 2 -> value:"",           badInput:true     (year 0002 -> 29.02. does not exist)
typed 0 -> value:"0020-02-29", badInput:false    (leap year -> input + change)
typed 2 -> value:"",           badInput:true     (year 0202 -> input + change with "")
```

Writing that empty value back into the element clears the native date editor, so the day and month segments were gone and the final digit could not complete the date.

**Root cause — two write-back paths, both in Angular only:**

1. `handleFrameworkEventAngular` echoed user input through `writeValue`, which does `renderer.setProperty(element, 'value', value)`.
2. The same call sets the model signal, and the model feeds `[value]="value() ?? _value() ?? ''"`. Angular dirty-checks a binding against the last bound expression, sees `"0020-02-29"` become `""` and writes it too.

React binds `props.value` only (`scripts/post-build/react.ts`), and the Vue and Stencil renderers skip a property write when the element already holds the value — which is why only Angular was affected.

**Fix:**

- User input now goes through a generated `_setModelValue`, which updates the model signal without touching the DOM. `writeValue` keeps the DOM write and stays reserved for programmatic writes from reactive forms and `ngModel`, so a form reset still clears the field.
- For Angular the element renders from `state._value`, which mirrors `props.value` except while `validity.badInput` is set and the model is empty. The bound expression therefore stays put during an unparsable entry. `state._value` starts out `undefined` so the first render still uses `props.value`; React, Vue and Stencil keep their previous binding through `useTarget`.

Both form APIs still receive the empty value, so no stale date can be submitted while the field visibly holds an invalid entry — that was the open point of the earlier attempt in #7762, which worked around it by pushing `undefined` into the public `value` model.

**Known gap (documented in `packages/components/AGENTS.md`):** while the entry is unparsable, `state._value` holds the last parsable value while the element holds none. A programmatic write of exactly that value does not change the bound expression and so does not reach the element. Every other value does, because `writeValue` also sets the property directly.

**Verification:** new e2e spec `showcases/e2e/input/input-date.spec.ts` run against the Angular showcase — 3/3 fail on the unfixed output, 3/3 pass with the fix. The existing form suite (`e2e/home`: inputs, textareas, selects, checkboxes, radios, switches, custom-selects, covering plain bindings, `ngModel` and reactive forms) passes with the write-back removed, and the `@db-ux/core-components` unit suite (99 tests) passes.

resolves #7748

### Code Quality

### Validation

- [ ] I have tested across all relevant frameworks (React, Angular, Vue) if applicable — verified locally in Angular (the affected target) plus the unit suite; React, Vue and Stencil keep their previous binding and are covered by CI

### General

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/fix-date-input-badinput-reset>


mfranzke:
@codex review

nmerget:
Are https://github.com/db-ux-design-system/core-web/issues/6147 and https://github.com/db-ux-design-system/core-web/issues/6092 still working? We just created a unit test back then, which might not be the best choice for such component

michaelmkraus:
> Are #6147 and #6092 still working? We just created a unit test back then, which might not be the best choice for such component

Both are fine. I added some tests to ensure this.

mfranzke:
@codex review

showcases/e2e/input/input-number.spec.ts
nmerget, line 1: Can we add a number input to the `Example - Types with min and max` examples or another example and test this accross all frameworks?

showcases/e2e/input/input-undefined-value.spec.ts
nmerget, line 1: This test can be easily converted to an showcase example with a state holding the undefined value. We can test it accross all frameworks. And this should work as well for React to make it `undefined`

nmerget:
https://github.com/user-attachments/assets/abaa65c5-5d26-41a4-bf62-1a63922088e0

When I click on "Reset" all inputs are cleared. When I write something first, the input resets correctly back to "test1". Something is not working properly at the moment

packages/components/src/components/input/input.lite.tsx
```diff
@@ -317,7 +328,12 @@ export default function DBInput(props: DBInputProps) {
 				disabled={getBoolean(props.disabled, 'disabled')}
 				required={getBoolean(props.required, 'required')}
 				step={getStep(props.step)}
-				value={props.value ?? state._value ?? ''}
+				value={useTarget({
+					// state._value starts out undefined, so the first
+					// render still uses props.value.
+					angular: state._value ?? props.value ?? '',
+					default: props.value ?? state._value ?? ''
```
nmerget, line 335: I like this `useTarget` and we should apply it for all form-components.

Inside `packages/components/scripts/post-build/react.ts` line 168 we only set `props.value` for react we should remove it there and move it to all form-components instead for better maintainability

michaelmkraus:
> Bildschirmaufnahme.2026-09-15.080113.mp4 
> When I click on "Reset" all inputs are cleared. When I write something first, the input resets correctly back to "test1". Something is not working properly at the moment

Fixed. A native reset is a programmatic write, so it now goes through `writeValue` again instead of `_setModelValue`. The model-only path did nothing when the reset value already matched the model. Applies to `input`, `textarea` and `select`.

packages/components/src/utils/form-components.ts
```diff
@@ -41,9 +41,30 @@ export const handleFrameworkEventAngular = (
...
+ * The gap is deliberately narrow: it applies only while the element itself
+ * reports `badInput` **and** the incoming value is empty. Every other write
+ * reaches the display value, which is what keeps a programmatic reset to
+ * `undefined` working
+ * (https://github.com/db-ux-design-system/core-web/issues/6147).
+ *
+ * @internal
+ */
+export const shouldKeepDisplayValue = (element: any, value: any): boolean =>
+	Boolean(element?.validity?.badInput) && !value;
```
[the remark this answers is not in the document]
michaelmkraus: Reactive forms and `ngModel` are unaffected — they write through `writeValue`, which sets the property directly, and the reset path now does too. The remaining case is a direct `[value]` binding flipped to a `falsy` value while `badInput` is set; that's a documented edge case now.

showcases/e2e/input/input-number.spec.ts
michaelmkraus: Done. Added a `type="number"` to the min/max example and pointed the spec at the generated example page `#/03/input` like `input-date.spec.ts`, dropping the Angular-only skip so it runs across all framework showcases.

packages/components/src/components/input/input.lite.tsx
michaelmkraus: Done. The `value` binding is now an explicit `useTarget` in `input`, `textarea` and `select`, and the string replacement in `react.ts` is removed. Generated React output is unchanged (`value={props.value}`), it's just visible in the component source now.

showcases/e2e/input/input-number.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+
+	test('accepts an exponent without clearing the field', async ({ page }) => {
+		const input = getNumberInput(page);
+		await clear(page, input);
+
+		await page.keyboard.press('1');
+		await page.keyboard.press('e');
+		// `1e` is unparsable, so the element reports an empty value while the
+		// editor keeps the entry. Nothing may write that empty value back.
+		await expect(input).toHaveJSProperty('validity.badInput', true);
```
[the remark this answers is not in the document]
michaelmkraus: False positive. Playwright's `to.have.property` splits the first argument on `.` and walks the path, so this asserts `input.validity.badInput`, not `input['validity.badInput']`. Verified in the `to.have.property` matcher of `playwright-core@1.62.1`. No change needed.

showcases/e2e/input/input-undefined-value.spec.ts
michaelmkraus: Done for the cross-framework part: the reset fixture is now in all three form showcases and `input-undefined-value.spec.ts` runs on every framework. For React the fixture resets to `''` rather than `undefined`: React binds `props.value` alone, so `value={undefined}` makes the input uncontrolled instead of clearing it — clearing a controlled React input means `''`.

nmerget:
https://github.com/user-attachments/assets/7796d57f-d856-4788-a424-70b5f7c3fddd

There is some strange flashing for the placeholder when reseting in Angular.

Also for Vue and Angular the "Set value to undefined" works just once. I also add the `(change)` event and still it's not working, while React seems fine

michaelmkraus:
> Also for Vue and Angular the "Set value to undefined" works just once. I also add the `(change)` event and still it's not working, while React seems fine

Fixed with two-way binding (`[(value)]` / `v-model:value`) plus an e2e test that resets twice. It fails as expected when I switch back to one-way.

michaelmkraus:
> There is some strange flashing for the placeholder when reseting in Angular.

Confirmed on main too, so it predates this PR — tracked separately in #8231

PR #8060, opened 2026-09-09
fix(DBShell): prevent logo position jump between expanded and collapsed state

### LINKED ISSUE #7940: DBShell - Logo position expand/collapse

- [x] All
- [ ] HTML
- [ ] React
- [ ] Angular
- [ ] Vue
- [ ] Web components
- [ ] Power Apps

### Reproduction case

Logo position change, 'jump' between expanded and collapsed state

Image
Image

### Expected Behaviour

Logo position stay the same in the expanded and collapsed state

### Screenshots

_No response_

### Browser version

Edge

https://design-system.deutschebahn.com/core-web/review/feat-shell/components/navigation/shell/overview

InfraGO

Fplo-Verwaltung / Unterjährige Veröffentlichung UJV

nmerget:
## Proposed changes

Fixes the DBShell logo "jumping" between the expanded and collapsed states. Instead of switching the layout in the collapsed state, the logo is now animated into position via a `transform: translateX(...)` (centering it within the collapsed control panel width) with a transition, so the transition between states is smooth. The brand text container is visually hidden in the collapsed state instead of relying on `overflow: hidden`.

resolves #7940

### Code Quality

### Validation

### General

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/fix-shell-logo-position>


mfranzke:
@codex review

packages/components/src/styles/internal/_control-panel-desktop.scss
```diff
@@ -315,14 +328,27 @@ $control-panel-desktop-vertical-padding-inline-end: calc(
 
 		.db-control-panel-brand {
 			inline-size: variables.$db-sizing-md;
-			overflow: hidden;
+
+			.db-control-panel-brand-text-container {
+				@extend %a11y-visually-hidden-text;
+
+				position: absolute;
```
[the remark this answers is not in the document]
nmerget: Fixed in eefa31ea5533. The collapsed panel now hides `.db-control-panel-brand-end-slot` with `display: none`, so an `endSlot` (e.g. a badge) can no longer sit next to the logo, overflow the collapsed panel, or force the logo to shrink now that `overflow: hidden` is gone.

packages/components/src/styles/internal/_control-panel-desktop.scss
```diff
@@ -315,14 +328,27 @@ $control-panel-desktop-vertical-padding-inline-end: calc(
...
 					var(--db-logo-aspect-ratio)
 				);
+
+				// The inline-size of the container minus the height of the logo * aspect-ratio divided by 2 to center the logo in the container. This is a workaround to use animations.
+				transform: translateX(
+					calc(
+						(#{variables.$db-sizing-md} - 1lh * #{$aspect-ratio}) /
+							2
+					)
+				);
```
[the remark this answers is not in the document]
nmerget: Fixed in eefa31ea5533. The translation is now direction-aware: the offset is extracted into a `$logo-translation` variable and negated under `&:dir(rtl)`. `translateX` is physical, but the sole in-flow flex item (the logo `::before`) is pinned to the inline-start edge, which is the physical right in RTL — so it now translates left toward the center there, restoring the both-direction centering the previous `margin: auto` provided. This follows the existing `:dir(rtl)` pattern used in `tabs.scss`.

packages/components/src/styles/internal/_control-panel-desktop.scss
```diff
@@ -315,15 +328,38 @@ $control-panel-desktop-vertical-padding-inline-end: calc(
...
+				transform: translateX($logo-translation);
+				aspect-ratio: $aspect-ratio;
 				background-image: var(--db-logo-url-short, var(--db-logo-url));
+
+				// In RTL the logo sits at the physical right edge, so the
+				// translation has to move it toward the center in the opposite
+				// direction.
+				&:dir(rtl) {
+					transform: translateX(calc(-1 * #{$logo-translation}));
+				}
```
michaelmkraus: Nested inside `&::before`, this compiles to `.db-control-panel-brand:before:dir(rtl)`. A pseudo-element may only be followed by user-action pseudo-classes (`:hover`, `:active`, `:focus`), so browsers drop the rule and RTL stays uncorrected. Verified: `CSS.supports('selector(div::before:dir(rtl))')` is `false` in Chromium and WebKit (`div:dir(rtl)::before` is `true`), and on the branch preview the collapsed logo keeps `matrix(1, 0, 0, 1, 1.71429, 0)` after setting `dir="rtl"`.

Please move the pseudo-class to the host element and hoist the variables one level up:

```scss
.db-control-panel-brand {
	$aspect-ratio: var(--db-logo-aspect-ratio-short, var(--db-logo-aspect-ratio));
	$logo-translation: calc((#{variables.$db-sizing-md} - 1lh * #{$aspect-ratio}) / 2);

	&::before { transform: translateX($logo-translation); }
	&:dir(rtl)::before { transform: translateX(calc(-1 * #{$logo-translation})); }
}

File: packages/components/src/styles/internal/_control-panel-desktop.scss [hunk 4]
```diff
@@ -297,6 +297,19 @@ $control-panel-desktop-vertical-padding-inline-end: calc(
 				}
 			}
 		}
+
+		.db-control-panel-brand {
+			@media screen and (prefers-reduced-motion: reduce) {
+				&::before {
+					transition: none;
+				}
+			}
```
michaelmkraus: This `reduce` query is nested inside the surrounding `no-preference` query, so it compiles to `@media screen and (prefers-reduced-motion:no-preference) and (prefers-reduced-motion:reduce)` (confirmed in the deployed CSS) and can never match. Behaviour is still correct because the outer guard already handles reduced motion, so this block can just be removed.

nmerget: This was a false-positive auto fix from stylelint, removed it

packages/components/src/styles/internal/_control-panel-desktop.scss
nmerget: fixed

packages/components/src/styles/internal/_control-panel-desktop.scss
```diff
@@ -315,16 +322,36 @@ $control-panel-desktop-vertical-padding-inline-end: calc(
...
-					var(--db-logo-aspect-ratio)
-				);
+				transform: translateX($logo-translation);
+				aspect-ratio: $aspect-ratio;
 				background-image: var(--db-logo-url-short, var(--db-logo-url));
 			}
+
+			&:dir(rtl) {
+				transform: translateX(calc(-1 * #{$logo-translation}));
+			}
```
michaelmkraus: Compiles to `.db-control-panel-brand:dir(rtl)` — the transform hits the container, not the logo. Since an ancestor transform also moves `::before`, both translations cancel and the collapsed logo stays un-centered in RTL. Needs `&:dir(rtl)::before`.

packages/components/src/styles/internal/_control-panel-desktop.scss
```diff
@@ -315,16 +322,36 @@ $control-panel-desktop-vertical-padding-inline-end: calc(
 
 		.db-control-panel-brand {
 			inline-size: variables.$db-sizing-md;
-			overflow: hidden;
+
+			.db-control-panel-brand-text-container {
+				@extend %a11y-visually-hidden-text;
+
+				position: absolute;
+			}
```
michaelmkraus: `%a11y-visually-hidden-text` keeps the box in the layout (text pushed out via `text-indent`); adding `position: absolute` rebuilds `%a11y-visually-hidden`, which already does exactly this. Side effect: the abs-positioned box shrinks to near-zero inside the `$db-sizing-md` brand, so the text is hidden by `overflow: hidden`, not by `text-indent`. Suggest `@extend %a11y-visually-hidden;` without the extra `position`.

packages/components/src/styles/internal/_control-panel-desktop.scss
```diff
@@ -315,16 +322,36 @@ $control-panel-desktop-vertical-padding-inline-end: calc(
...
+
+			.db-control-panel-brand-end-slot {
+				display: none;
+			}
```
michaelmkraus, line 329: `display: none` also removes the slot from the a11y tree, while the text container above stays available on purpose — with `overflow: hidden` the slot was only clipped, so AT still announced it. If the `endSlot` can carry meaning (badge), the same visually-hidden treatment would be more consistent.

mfranzke:
@codex review

packages/components/src/styles/internal/_control-panel-desktop.scss
```diff
@@ -315,16 +322,36 @@ $control-panel-desktop-vertical-padding-inline-end: calc(
...
-					--db-logo-aspect-ratio-short,
-					var(--db-logo-aspect-ratio)
-				);
+				transform: translateX($logo-translation);
+				aspect-ratio: $aspect-ratio;
 				background-image: var(--db-logo-url-short, var(--db-logo-url));
 			}
+
+			&:dir(rtl) {
+				transform: translateX(calc(-1 * #{$logo-translation}));
```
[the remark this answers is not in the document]
nmerget: Fixed in f82b05e58503. Changed the RTL rule to `&:dir(rtl)::before` so the negated `translateX` is applied to the logo pseudo-element itself, not the container — the two offsets no longer cancel and the collapsed logo stays centered in RTL.

packages/components/src/styles/internal/_control-panel-desktop.scss
nmerget: Fixed in f82b05e58503. The end slot now shares the same `@extend %a11y-visually-hidden;` treatment as the text container instead of `display: none`, so a meaningful `endSlot` (e.g. a badge) stays in the a11y tree and is still announced by AT while being visually clipped in the collapsed panel.

packages/components/src/styles/internal/_control-panel-desktop.scss
nmerget: Fixed in f82b05e58503. Switched to `@extend %a11y-visually-hidden;` and dropped the extra `position: absolute` — the placeholder already handles absolute positioning and 1px clipping, so this no longer rebuilds it inconsistently.

packages/components/src/styles/internal/_control-panel-desktop.scss
nmerget: Fixed in f82b05e58503. The RTL rule is now `&:dir(rtl)::before`, so the negated translation hits the logo pseudo-element directly instead of the `.db-control-panel-brand` container. This restores the both-direction centering: in RTL the sole in-flow flex item (the logo `::before`) is pinned to the physical right edge and now translates left toward the center.

packages/components/src/styles/internal/_control-panel-desktop.scss
```diff
@@ -315,16 +322,31 @@ $control-panel-desktop-vertical-padding-inline-end: calc(
...
-					var(--db-logo-aspect-ratio)
-				);
+				transform: translateX($logo-translation);
+				aspect-ratio: $aspect-ratio;
 				background-image: var(--db-logo-url-short, var(--db-logo-url));
 			}
+
+			&:dir(rtl)::before {
+				transform: translateX(calc(-1 * #{$logo-translation}));
+			}
```
michaelmkraus: `--db-logo-aspect-ratio-short` is defined only once in `@db-ux/db-theme` 6.2.0 . All 12 `[data-logo=…]` variants set `--db-logo-aspect-ratio` only, so here the fallback resolves to the **long** ratio and `$logo-translation` goes negative — e.g. `db-systel` (544/141): logo width `1lh × 3.86` = 92.6px in a 40px rail → `translateX(-26.3px)`.

Since `overflow: hidden` is gone (line 324) nothing clips: `.db-control-panel-desktop` has no `overflow`, and `.db-control-panel-desktop-brand-container::before { overflow: hidden }` never generates (no `content`). The full wordmark now paints past both edges of the collapsed rail and over `.db-main`; before it was clipped to 40px. `data-logo="db-systel"` is what our own `shell.agent.lite.tsx` shows as the example.

CI won't catch this — no showcase sets `data-logo`, so `Desktop - Left > Navigation Left Collapse` uses the `:root` logo (10/7 = 34.3px), the only ratio that fits.

Clamp the offset and keep a clip:

```scss
overflow: clip;

&::before {
	transform: translateX(max(0px, #{$logo-translation}));
}

&:dir(rtl)::before {
	transform: translateX(min(0px, calc(-1 * #{$logo-translation})));
}

PR #8080, opened 2026-09-10
perf(ci): deploy only the current branch subtree to gh-pages

nmerget:
## Proposed changes

Task 2 of the pipeline-performance effort.

Reworks the gh-pages preview deploy so it publishes **only the current branch's subtree** instead of downloading, unpacking and re-committing the entire accumulated gh-pages site on every PR.

- `scripts/github/build-gh-page.ts` now assembles only `review/<branch>` (or `version/<name>` and `version/latest` for releases) into `./public`. No more `gh-pages` tarball fetch/unpack.
- `03-deploy-gh-pages.yml` publishes with `keep_files: true`, so other branches' previews and published versions are preserved (merged on top) rather than replaced. Stale-preview pruning stays in `cleanup.yml`.
- The deploy job no longer runs a full `pnpm install`; the script has no runtime deps and is executed with `node` via the `.nvmrc`-pinned Node 26 (a lightweight `setup-node` guarantees the correct Node without the install cost).

**Why:** on baseline run `34454854684` the deploy job took ~600s (382s "Build page" + 158s "Deploy to GH-Pages"), driven by re-processing the whole site. This cost grew with the number of accumulated previews.

Verified locally: the reworked script, run with `node build-gh-page.ts`, produces exactly `public/review/<branch>/…` and nothing else.

This branch exists to measure the deploy-job delta against the baseline side by side.

### Code Quality

### Validation

### General

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/perf-gh-pages-subtree-deploy>


.github/workflows/03-deploy-gh-pages.yml
```diff
@@ -98,13 +100,18 @@ jobs:
           REPO_NAME: ${{ inputs.repoName }}
           OWNER_NAME: ${{ inputs.repoOwner }}
         run: |
-          pnpm --silent tsx scripts/github/build-gh-page.ts
+          node scripts/github/build-gh-page.ts
```
mfranzke: ♥️

mfranzke:
@codex review

.github/workflows/03-deploy-gh-pages.yml
```diff
@@ -98,13 +100,18 @@ jobs:
...
+      # keep_files: true merges the published subtree on top of the existing
+      # gh-pages branch, so previews for other branches/versions are preserved
+      # without downloading and re-committing the whole site. Stale previews are
+      # pruned separately by cleanup.yml.
       - name: 🥅 Deploy to GH-Pages
         uses: peaceiris/actions-gh-pages@84c30a85c19949d7eee79c4ff27748b70285e453 # v4.1.0
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           publish_dir: ./public
+          keep_files: true
```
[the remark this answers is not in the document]
nmerget: Good catch, this was a real regression. Fixed in `29ce20a`.

Instead of `keep_files: true` (which never removes anything, so a file deleted from a rebuilt preview lingered), each deploy now publishes into a scoped `destination_dir`:
- PR previews -> `review/<branch>`
- pre-releases -> `version/<name>`
- releases -> `version/<name>` **and** `version/latest`

With `keep_files` back at its default (`false`), the action replaces exactly the owned subtree — removing stale files — while leaving sibling subtrees (other branches' previews, other versions) untouched. The only thing published with `keep_files: true` is the release root redirect (`index.html`), since that single file has to coexist with all subtrees.

scripts/root-redirect/index.html
mfranzke, line 1: could you please elaborate a little bit further on the reason to add this file (at best in the description of this PR)? Additionally it would be resilient to add a regular HTML5 page schema with an `<a href="https://db-ui.github.io/live-demo/version/latest">…</a>` HTML element in there as a fallback in case that the redirect doesn't work.

nmerget: Good question, and it made me reconsider whether the file is needed at all — it isn't, so I've removed it (`68ad07e`).

The root `index.html` redirect already exists on `gh-pages` and always points to `version/latest`; it never changes between releases. The earlier version of this PR was republishing it on every release for no reason. Now the deploy only updates the subtrees that actually change:

- PR previews -> `review/<branch>`
- pre-releases -> `version/<name>`
- releases -> `version/<name>` **and** `version/latest`

all via scoped `destination_dir` (so each owned subtree is replaced, stale files removed, siblings preserved). No `build-gh-page.ts`, no committed redirect file, no root republish step.

If we ever do want to harden the existing root redirect with an HTML5 fallback `<a href>`, that's a separate, one-off change to the file already on `gh-pages` rather than something this deploy needs to regenerate — happy to open a small follow-up for that if you'd like.

mfranzke: > If we ever do want to harden the existing root redirect with an HTML5 fallback `<a href>`, that's a separate, one-off change to the file already on `gh-pages` rather than something this deploy needs to regenerate — happy to open a small follow-up for that if you'd like.

oh yes, let's do that. And besides that: Do you know where I could find John Connor ?

.github/workflows/03-deploy-gh-pages.yml
```diff
@@ -38,11 +38,13 @@ jobs:
...
-        env:
-          ASSET_INIT_VECTOR: ${{ secrets.ASSET_INIT_VECTOR }}
-          ASSET_PASSWORD: ${{ secrets.ASSET_PASSWORD }}
+      # Only Node is needed here (no pnpm install): the deploy script has no
+      # runtime dependencies and just assembles this branch's subtree. .nvmrc
+      # pins Node 26, which executes the TypeScript script natively.
+      - name: 🆙 Setup node
+        uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
+        with:
+          node-version-file: ".nvmrc"
```
michaelmkraus: `build-gh-page.ts` was deleted, so nothing in this job runs Node or TypeScript anymore. The step is overhead and the comment describes a script that no longer exists. Both can go.

nmerget: Good catch, done in `abcee49`. Removed the `Setup node` step and its stale comment — the job no longer runs Node or TypeScript now that `build-gh-page.ts` is gone; it just downloads artifacts, rearranges them with `mv`, and publishes via `peaceiris/actions-gh-pages`.

PR #8083, opened 2026-09-10
perf(ci): shallow-fetch main in init instead of full history

nmerget:
## Proposed changes

Two related CI changes to the `init` job.

**1. Shallow-fetch main instead of full history.** `init` checked out with `fetch-depth: 0`, pulling the complete history of every ref (~3 min). The only step needing history is the aria-snapshot diff (`git diff origin/main` in `scripts/github/snapshot-changes/check-main-aria.ts`) — name, playwright-version, changeset and fork checks all use the Actions context, not git. Switched to `fetch-depth: 1` plus a dedicated `git fetch --depth=1 origin main`.

For this gate ("did any `__snapshots__/*-aria-snapshot.yaml` change") a tip-to-tip comparison is sufficient and errs toward running the a11y screen-reader test when unsure. On `main`/release the test is already forced via the existing `github.event.pull_request == null` condition, so gating behavior is preserved.

**2. Remove dead `test-ally` plumbing.** `02-e2e-showcases.yml` declared a `test-ally` input that no step referenced, and `default.yml` passed it into all four showcase test jobs. Removed. The only real consumer of init's `test-ally` output is the `test-screen-reader` job's `if:` condition, which is untouched. No behavior change.

**Why:** removes ~3 min from `init`, which is on the critical path before the pipeline fans out.

### Code Quality

### Validation

### General

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/perf-shallow-init-fetch>

### DIFF
--- a/.github/workflows/00-init.yml
+++ b/.github/workflows/00-init.yml
@@ -62,7 +62,11 @@ jobs:
       - name: ⏬ Checkout repo
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
-          fetch-depth: 0
+          # Shallow checkout. The only thing in this job that needs git history is
+          # the aria-snapshot diff against main (see "Should Test a11y" below), and
+          # that only needs main's tip, fetched shallowly in a dedicated step. A
+          # full fetch-depth: 0 pulled every ref's complete history (~3 min).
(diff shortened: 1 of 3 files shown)


mfranzke:
@codex review

mfranzke (approved):

PR #8095, opened 2026-09-11
perf(ci): run vite-plugin tests inside build-packages

nmerget:
## Proposed changes

Merges the vite-plugin tests into `build-packages` to remove a separate job (reduces peak runner usage under the org's 20-runner cap).

The vite-plugin `test` (generator unit tests + an integration test that runs `npm ci` and builds the `react-app`/`vue-app` fixtures) needs the `packages/{components,foundations}/build` styles and the `output/*` framework packages. All of these are produced in-tree by `pnpm run build` in `build-packages` — the fixtures reference them via `file:` deps, and `build-packages` already uploads `db-ux-output` from that same `output/`.

- Added a `Run Vite Plugin Tests` step to `01-build-packages.yml` after `pnpm run build`.
- Deleted `02-vite-plugin-test.yml`.
- Removed the `test-vite-plugin` job and its `checks-done` references from `default.yml`. A failure still blocks the pipeline via the existing `cancel-on-failure-build-packages`.

Net: one fewer job per run (its `Init Cache` + three artifact downloads gone).

### Code Quality

### Validation

### General

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/perf-merge-vite-plugin-into-build-packages>


mfranzke:
@codex review

.github/workflows/01-build-packages.yml
```diff
@@ -26,6 +26,15 @@ jobs:
           pnpm run build
           pnpm --filter scripts run docs-file-references
 
+      # Run the vite-plugin tests here rather than in a separate job. `pnpm run
+      # build` above already produced everything they need in-tree: the
+      # packages/{components,foundations}/build styles and the output/* framework
+      # packages that the integration-test fixtures reference via file: deps.
+      # Testing in-place avoids a whole extra job (Init Cache + artifact
+      # downloads) - valuable under the org's 20-runner cap.
+      - name: 🧪 Run Vite Plugin Tests
```
michaelmkraus: The step sits before all nine `⏫ Upload …` steps, so a test failure uploads no artifacts at all. Moving it after `⏫ Upload mcp-server build` (line 86–90) gates identically but keeps `packages/*/build and output/` downloadable for debugging.

.github/workflows/01-build-packages.yml
```diff
@@ -26,6 +26,15 @@ jobs:
...
+        run: pnpm --filter @db-ux/core-vite-plugin run test
```
michaelkraus: A failure here now triggers `cancel-on-failure-build-packages → gh run cancel` kills the whole run. Previously `test-vite-plugin` had no cancel job, so a failure only made `checks-done` fail and could be re-run in isolation. Is this intended?

michaelmkraus: A failure here now triggers cancel-on-failure-build-packages → gh run cancel kills the whole run. Previously test-vite-plugin had no cancel job, so a failure only made checks-done fail and could be re-run in isolation.  Is this intended?

.github/workflows/01-build-packages.yml
nmerget: Good catch. Moved the `🧪 Run Vite Plugin Tests` step to after `⏫ Upload mcp-server build`, so a test failure now still leaves `packages/*/build` and `output/` uploaded and downloadable for debugging. Gating is unchanged since it stays in the same job. Done in a60a000.

.github/workflows/01-build-packages.yml
nmerget: Yes, that's intended. We want a vite-plugin test failure to surface as fast as possible and stop the whole run rather than keep burning runners on downstream jobs — failing fast is the goal here. The trade-off (no isolated re-run like the old standalone `test-vite-plugin` job) is acceptable; the run can be re-triggered as a whole.

.github/workflows/01-build-packages.yml
```diff
@@ -80,3 +80,14 @@ jobs:
...
+      # Run the vite-plugin tests here rather than in a separate job. `pnpm run
+      # build` above already produced everything they need in-tree: the
+      # packages/{components,foundations}/build styles and the output/* framework
+      # packages that the integration-test fixtures reference via file: deps.
+      # Testing in-place avoids a whole extra job (Init Cache + artifact
+      # downloads) - valuable under the org's 20-runner cap. This step runs
+      # after all uploads so a test failure still leaves the packages/*/build
+      # and output/ artifacts downloadable for debugging.
+      - name: 🧪 Run Vite Plugin Tests
+        run: pnpm --filter @db-ux/core-vite-plugin run test
```
michaelmkraus, line 93: If a test fails, the whole pipeline run gets cancelled now. Before, only this job failed and you could re-run it. Is this intended?

nmerget: Yes it should fail as fast as possible, that's even better here

PR #8097, opened 2026-09-11
test(ci): set Playwright workers to 75% on CI

nmerget:
## Proposed changes

**Experiment / measurement PR** — follow-up to the `50%` workers run (#8093), which roughly halved the showcase e2e suites (e.g. react-showcase shard axe-core 3m42s -> 1m53s).

Sets Playwright `workers: '75%'` on CI = **3 workers** on the 4-core public `ubuntu-24.04` runners.

**Open question this measures:** 1->2 workers gained ~2x because a full core was idle. 2->3 leaves only one core for the OS, the static `webServer`, and the (multi-process, CPU-hungry) browsers, so the extra worker may give little, break even, or regress with timeouts. `retries: 1` absorbs the odd blip — the thing to watch is whether **retry counts rise** and eat the savings, and whether snapshot tests get flaky.

Compare shard timings and retries against the `50%` run before deciding which to keep. Not intended to merge as-is.

### Code Quality

### Validation

### General

🔭🐙🐈 Test this branch here: <https://design-system.deutschebahn.com/core-web/review/test-playwright-workers-75>

### DIFF
--- a/showcases/playwright.config.ts
+++ b/showcases/playwright.config.ts
@@ -67,8 +67,15 @@ const config: PlaywrightTestConfig = {
 	fullyParallel: true,
 	/* Fail the build on CI if you accidentally left test.only in the source code. */
 	forbidOnly: Boolean(process.env.CI),
-	/* Opt out of parallel tests on CI. */
-	workers: process.env.CI ? 1 : undefined,
+	/*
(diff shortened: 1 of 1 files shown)


nmerget:
We save only 20 seconds on the e2e test compared to the 50% so we won't go with this

